PR #4, opened 2020-02-28
all get_patch tests passing

orionz:
I got the tests passing but need help getting the code clean and idiomatic. Also I got `InsertSequenceElement.after` working but I think by accident.  Could you feedback on how it should behave since you wrote the other code around it.

### DIFF
--- a/automerge-backend/src/object_store.rs
+++ b/automerge-backend/src/object_store.rs
@@ -61,39 +61,46 @@ impl ListState {
     }
 
     fn generate_diffs(&self) -> Vec<Diff> {
-        let mut after = 0;
-        let ops_in_order = list_ops_in_order(&self.operations_by_elemid, &self.following)
-            .ok()
-            .unwrap_or(vec![]);
-        let mut diffs = vec![Diff {
+        let mut diffs = Vec::new();
+
+        let head = Diff {
             action: DiffAction::CreateList(self.object_id.clone(), self.sequence_type.clone()),
             conflicts: Vec::new(),
-        }];
-        diffs.extend(
-            ops_in_order
-                .iter()
-                .rev()
-                .filter_map(|(_, ops)| {
-                    ops.active_op().map(|op| {
-                        let tmp = list_op_to_assign_diff(&op.operation, &self.sequence_type, after)
-                            .map(|action| Diff {
-                                action: action,
-                                conflicts: ops.conflicts(),
-                            });
-                        after += 1;
-                        tmp
-                    })
-                })
-                .flatten(),
-        );
-        diffs.push(Diff {
+        };
+
+        let ops_in_order = list_ops_in_order(&self.operations_by_elemid, &self.following)
+            .ok()
+            .unwrap_or_default();
+
+        let inserts = ops_in_order
+            .iter()
+            .rev()
+            .enumerate()
+            .filter_map(|(after,(_, ops))|
+                ops.active_op()
+                    .and_then(|op| list_op_to_assign_diff(&op.operation, &self.sequence_type, after as u32)
+                    .map(|action|
+                        Diff {
+                            action,
+                            conflicts: ops.conflicts(),
+                        }
+                    )
+                )
+            );
+
+        let tail = Diff {
             action: DiffAction::MaxElem(
                 self.object_id.clone(),
                 self.max_elem,
                 self.sequence_type.clone(),
             ),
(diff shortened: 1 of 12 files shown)


automerge-backend/src/object_store.rs
```diff
@@ -52,6 +60,39 @@ impl ListState {
...
+            ops.active_op()
+                .map(|op| {
+                  let tmp = list_op_to_assign_diff(&op.operation, &self.sequence_type, after)
+                      .map(|action|
+                          Diff {
+                            action: action,
+                            conflicts: ops.conflicts(),
+                          }
+                      );
+                  after += 1;
```
alexjg: This feels a bit gnarly, `map` generally being assumed to be side effect free. I _think_ it would be a good candidate for `Iter::scan`.

automerge-backend/src/object_store.rs
```diff
@@ -188,6 +229,28 @@ impl MapState {
         }
     }
 
+    fn generate_diffs(&self) -> Vec<Diff> {
+        let mut diffs = vec![];
+        if self.object_id != ObjectID::Root {
+            diffs.push(
+                Diff {
+                    action: DiffAction::CreateMap(self.object_id.clone(), self.map_type.clone()),
+                    conflicts: vec![],
```
alexjg: I've been tending to use `Vec::new()` rather than `vec![]` for creating empty arrays, no idea which is better but we should probably choose one for consistency.

automerge-backend/src/object_store.rs
```diff
@@ -279,6 +342,27 @@ impl ObjectStore {
         }
     }
 
+    pub fn generate_diffs(&self) -> Vec<Diff> {
+        let mut diffs = vec![];
+        let mut seen = HashSet::new();
+        let mut next : Vec<ObjectID> = vec![ ObjectID::Root ];
+
+        while !next.is_empty() {
```
alexjg: I think this would probably be nicely expressed as `while let Some(next) = next {...}`

automerge-backend/src/patch.rs
```diff
@@ -48,6 +68,17 @@ pub struct Diff {
     pub conflicts: Vec<Conflict>,
 }
 
+impl Diff {
+    pub fn links(&self) -> Vec<ObjectID> {
+        let mut oids = vec![];
+        self.action.value().map(|v| v.object_id().map(|oid| oids.push(oid)));
```
alexjg: Probably could use `Option::and_then` rather than nested `map` calls like this. Personally I dislike using `map` for side effecting things so probably would write this as:

```
if let Some(oid) = self.action.value().and_then(|v| v.object_id()) { oids.push oid }
```

automerge-backend/src/patch.rs
```diff
@@ -48,6 +68,17 @@ pub struct Diff {
...
+        for c in self.conflicts.iter() {
+            c.value.object_id().map(|oid| oids.push(oid));
```
alexjg: Likewise re. side effecting calls I personall think `if let Some(oid) = c.value.object_id() { oids.push(oid)}` reads better.

automerge-backend/src/object_store.rs
```diff
@@ -52,6 +60,39 @@ impl ListState {
         }
     }
 
+    fn generate_diffs(&self) ->  Vec<Diff> {
+        let mut after = 0;
+        let ops_in_order = list_ops_in_order(&self.operations_by_elemid, &self.following).ok().unwrap_or(vec![]);
```
alexjg: Here it would be better to use `unwrap_or_else(Vec::new)`, this avoids allocating unless there actually is an unwrap needed.

automerge-backend/src/object_store.rs
```diff
@@ -52,6 +60,50 @@ impl ListState {
...
+        };
+
+        let ops_in_order = list_ops_in_order(&self.operations_by_elemid, &self.following)
+            .ok()
+            .unwrap_or_default();
+
+        let inserts = ops_in_order
+            .iter()
+            .rev()
+            .enumerate()
```
alexjg: Does this `enumerate` call need to be after we filter out the deleted elements (via `filter_map` on `op.active_op()`, which will be `None` if there are no operations a la delete)? I think we might end up with incorrect element indices in the case of deleted elements?

orionz: Should be fixed now.  Had to add an unwrap()

PR #8, opened 2020-03-06
Cleanup

orionz:
Did some cleanup - a lot in making the code around Clock clearer and more concise.  There are some dramatic changes here so I wanted feedback before merging.

### DIFF
--- a/automerge-backend/src/protocol.rs
+++ b/automerge-backend/src/protocol.rs
@@ -87,7 +87,6 @@ impl ActorID {
 pub struct Clock(pub HashMap<ActorID, u32>);
 
 impl Clock {
-
     pub fn empty() -> Clock {
         Clock(HashMap::new())
     }
@@ -112,9 +111,9 @@ impl Clock {
 
     pub fn set(&mut self, actor_id: &ActorID, seq: u32) {
         if seq == 0 {
-          self.0.remove(actor_id);
+            self.0.remove(actor_id);
         } else {
-          self.0.insert(actor_id.clone(), seq);
+            self.0.insert(actor_id.clone(), seq);
         }
     }
 
@@ -123,9 +122,8 @@ impl Clock {
     }
 
     fn less_or_equal(&self, other: &Clock) -> bool {
-        self.into_iter().all(|(actor_id, _)| {
-            self.get(actor_id) <= other.get(actor_id)
-        })
+        self.into_iter()
+            .all(|(actor_id, _)| self.get(actor_id) <= other.get(actor_id))
     }
 }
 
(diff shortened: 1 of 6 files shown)


automerge-backend/src/protocol.rs
```diff
@@ -87,63 +87,58 @@ impl ActorID {
...
         });
-        Clock(result)
     }
 
-    pub fn is_before_or_concurrent_with(&self, other: &Clock) -> bool {
-        other
-            .0
-            .iter()
-            .all(|(actor_id, seq)| self.0.get(actor_id).unwrap_or(&0) >= seq)
+    pub fn union(&self, other: &Clock) -> Clock {
```
alexjg: I'm not sure I like naming this `union`, the problem is that there are many possible unions for a clock as you have to decide what to do when an actor exists in both clocks with different values. Mostly what we do in Automerge is take the maximum, but you can also take the minimum. The "upper bound" name was from the idea of vector clocks as a lattice, but we could use some other name if it's not very clear, maybe `max`?

automerge-backend/src/protocol.rs
```diff
@@ -87,63 +87,58 @@ impl ActorID {
...
-    ///
-    /// TODO This feels like it should be a PartialOrd implementation but I
-    /// can't figure out quite what that should look like
-    ///
+
     pub fn less_or_equal(&self, other: &Clock) -> bool {
-        self.0.iter().chain(other.0.iter()).all(|(actor_id, _)| {
-            self.0.get(actor_id).unwrap_or(&0) <= other.0.get(actor_id).unwrap_or(&0)
-        })
+        self.into_iter()
```
alexjg: I would like to check this with Martin, the implementation I wrote, which looks up all actors in both clocks, is taken directly from the javascript implementation. I don't _think_ it's necessary according to the documentation (which was also copied from the javascript implementation), but it's probably worth checking.

Also, can we leave the documentation comment in?

automerge-backend/src/protocol.rs
orionz: In the automerge context clocks only go up - never down.  And at least for me when I think about the union of two clocks its natural to think of it as the upper bound of both.  I get that if this were a general purpose library union would make less sense... `max` also feels good to me.  I only prefer `union` b/c it seems very clear to me the result is a new clock and not a mutation of the old clock. Your call tho - I can go either way

automerge-backend/src/protocol.rs
orionz: Yeah i saw that.  Problem is all the keys in the other clock that are not in this clock are `> 0` and which means we already know they will pass.  Including those keys doesn't seem to change the result for any given two clocks.

orionz: And all the tests still passed which gave me confidence.

alexjg: Cool, let's merge and just check with Martin when we go through the code this afternoon.

automerge-backend/src/protocol.rs
alexjg: Let's go with `max`, I agree that we never need anything other than `max` but it's clearer to me when reading the code what the semantics are.

PR #9, opened 2020-03-11
fix transitive deps bug, refactor actor_histories, remove root_value

orionz:
Transitive deps was being calculated wrong.  Merged functionality of actor_histories, history and states into one struct.  Removed some dead code.


automerge-backend/src/actor_states.rs
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    pub fn is_concurrent(
+        &self,
+        actor_id1: &ActorID,
+        seq1: u32,
+        actor_id2: &ActorID,
+        seq2: u32,
+    ) -> bool {
+        let clock1 = self.get_deps(actor_id1, seq1).unwrap();
```
alexjg: Can we use default empty clocks here instead of panicking? I _think_ the semantics would still be correct. An empty clock will return 0 for all actor IDs.

automerge-backend/src/actor_states.rs
```diff
@@ -0,0 +1,95 @@
...
+    fn transitive_deps(&self, clock: &Clock) -> Clock {
+        let mut all_deps = clock.clone();
+        clock
+            .into_iter()
+            .filter_map(|(actor_id, seq)| self.get_deps(actor_id, *seq))
+            .for_each(|deps| all_deps.merge(deps));
+        all_deps
+    }
+
+    pub(crate) fn add_change(&mut self, change: Change) -> Result<bool, AutomergeError> {
```
alexjg: Feel like we should add a doc comment that the `bool` refers to whether the change has been applied already, it's not obvious from the types.

automerge-backend/src/actor_states.rs
```diff
@@ -0,0 +1,95 @@
+use crate::error::AutomergeError;
+use crate::protocol::{ActorID, Change, Clock};
+use std::collections::HashMap;
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct ActorStates {
```
alexjg: Can we add a small doc comment describing what the responsibilities of this struct are?

automerge-backend/src/backend.rs
```diff
@@ -146,7 +153,12 @@ impl Backend {
     }
 
     pub fn merge(&mut self, remote: &Backend) -> Result<Patch, AutomergeError> {
-        let missing_changes = remote.get_missing_changes(self.op_set.clock.clone());
+        let missing_changes = remote
+            .get_missing_changes(self.op_set.clock.clone())
+            .iter()
+            .cloned()
+            .cloned()
```
alexjg: Do we need to clone twice?

automerge-backend/src/op_set.rs
```diff
@@ -201,41 +187,18 @@ impl OpSet {
...
-                .actor_histories
-                .transitive_dependencies(&change.actor_id, change.seq)
-                .with(&change.actor_id, change.seq - 1),
-        });
-
         let actor_id = change.actor_id.clone();
         let seq = change.seq;
+        let operations = change.operations.clone();
+
+        if !self.states.add_change(change)? {
```
alexjg: Man this is so much cleaner :slightly_smiling_face:

automerge-backend/src/actor_states.rs
```diff
@@ -0,0 +1,95 @@
...
+        ActorStates {
+            change_by_actor: HashMap::new(),
+            deps_by_actor: HashMap::new(),
+            history: Vec::new(),
+        }
+    }
+
+    pub fn is_concurrent(
+        &self,
+        actor_id1: &ActorID,
```
alexjg: I'm concerned that by splitting out the actor ID and the sequence number from the `OperationWithMetadata` it becomes much easier to accidentally pass those arguments in the wrong order.  We only ever use this method to compare `OperationWithMetadata` structs, is there a reason not to take them directly as arguments?

automerge-backend/src/actor_states.rs
orionz: Good call

automerge-backend/src/backend.rs
orionz: Oh thats right - i had forgotten about this... Fiddled with some lifetimes and managed it with no cloning at all!

automerge-backend/src/actor_states.rs
orionz: Thats fair - the function seemed really general purpose and the OpWMetadata seemed very situational - but its a good point - i put it back

automerge-backend/src/actor_states.rs
orionz: Added

automerge-backend/src/actor_states.rs
orionz: I fussed with this for a bit that appealed to my need to use references instead of copying as much as possible.  Unfortunately it required me to add a blank clock to reference.  Let me know if this passes your sniff test or if there's a cleaner way.

PR #17, opened 2020-05-19
cli: Export state as JSON

scotttrinh:
Definitely a big work in progress at the moment, but wanted to get some early feedback on how best to structure some of this idiomatically.

Swapped out clap for structopt which gives the options a nice declarative structure and uses clap under the hood.

Probably need some test data to actually exercise this new functionality on before this is ready for final review.


automerge-cli/src/export.rs
```diff
@@ -0,0 +1,42 @@
...
+
+fn get_state(changes_file: &Path) -> Result<Value, AutomergeCliError> {
+    let input_data =
+        get_changes(changes_file).map_err(|_| AutomergeCliError::InvalidChangesFile)?;
+
+    let patch = get_patch(input_data).map_err(|_| AutomergeCliError::BackendError)?;
+    let mut frontend = automerge_frontend::Frontend::new();
+    frontend.apply_patch(patch).unwrap();
+
+    Ok(frontend.state().clone())
```
scotttrinh, line 32: Is there a way to avoid the `clone` call here?

automerge-cli/src/export.rs
```diff
@@ -0,0 +1,42 @@
...
+    backend.load_changes_binary(vec![changes_bytes]).unwrap();
+
+    backend.get_patch()
+}
+
+fn get_state(changes_file: &Path) -> Result<Value, AutomergeCliError> {
+    let input_data =
+        get_changes(changes_file).map_err(|_| AutomergeCliError::InvalidChangesFile)?;
+
+    let patch = get_patch(input_data).map_err(|_| AutomergeCliError::BackendError)?;
```
scotttrinh, line 28: Is using `map_err` idiomatic here to map different error types to application error types?

scotttrinh:
Current help output:

```
$ ./target/debug/automerge -h
automerge-cli 0.1.0
Automerge CLI

USAGE:
    automerge <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    export    Output current state of an Automerge document in a specified format
    help      Prints this message or the help of the given subcommand(s)
$ ./target/debug/automerge export -h
automerge-export 0.1.0
Output current state of an Automerge document in a specified format

USAGE:
    automerge export [OPTIONS] <changes-file>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -f, --format <format>    Format for output: json, toml [default: json]

ARGS:
    <changes-file>    File that contains automerge changes
```

automerge-cli/src/export.rs
scotttrinh: Switched this to return a `serde::Value` here since that's ultimately what we want to use in the consumer, but we'll need to update this once we need to export to other formats so it'd be good to know what the generally accepted pattern here is: we don't really need to keep the `frontend` around, so how should we return the `state` and release `frontend`? Or is that not idiomatic at all and we should return the `frontend` and use it to get the state in the consuming function?

automerge-cli/Cargo.toml
```diff
@@ -9,7 +9,8 @@ name = "automerge"
 path = "src/main.rs"
 
 [dependencies]
-clap = { git = "https://github.com/clap-rs/clap/" }
+structopt = "0.3"
```
alexjg, line 12: This is interesting, I was under the impression that structopt is now part of clap as part of clap v3, which is what is suggested by this crate https://github.com/clap-rs/clap/tree/master/clap_derive

I need to do more research but it feels like the ecosystem is consolidating around clap, thoughts?

automerge-cli/src/export.rs
alexjg: Not really, we do it in a number of places because the codebase is still very rough and ready but I think the idiomatic way to do this is to implement `From<X> for AutomergeCliError>`,  where `X` is the type you are receiving from the fallible call, that allows you to directly use `?` without `map_err`.

[anyhow](https://docs.rs/anyhow/1.0.13/anyhow/) handles a lot of the fiddly parts of this, might be worth a look.

We haven't spent much time sharpening up the error handling for this crate yet, lot's to do there.

automerge-cli/src/export.rs
```diff
@@ -0,0 +1,42 @@
+use crate::error::AutomergeCliError;
+use automerge_backend::AutomergeError;
+use automerge_protocol::Patch;
+use std::fs::File;
+use std::path::Path;
+use std::{io, io::Read};
+use serde_json::Value;
+
+fn get_changes(changes_file: &Path) -> io::Result<Vec<u8>> {
```
alexjg, line 9: I think this can be replaced by https://doc.rust-lang.org/std/fs/fn.read.html

automerge-cli/src/error.rs
```diff
@@ -0,0 +1,17 @@
+use std::error::Error;
+use std::fmt;
+
+#[derive(Debug, PartialEq)]
+pub enum AutomergeCliError {
```
alexjg, line 5: It's probably worth using `thiserror` to derive the  Error trait here, it takes away most of the pain of generating nice error types and it's already a dependency of the `automerge-protocol` trait.

automerge-cli/Cargo.toml
scotttrinh: Oh, fascinating! I've been using `structopt` for a while now and just kinda ignored what was happening over at `clap`. I'm all for using `clap`@3, I'll fix this up.

automerge-cli/src/error.rs
scotttrinh: 👀 

Great! I'll take a look there, thanks for pointing me in the right direction.

automerge-cli/src/export.rs
scotttrinh: ⚡ f176d45

automerge-cli/src/export.rs
Gozala: Idiomatic way is indeed to implement `From<X> for ...`. 

From what I've learned https://docs.rs/anyhow is great for "applciation level" errors, where you want to be able to package and pass along any kind of error.

And for "library level" errors, where there are only a handful of things that can go wrong and you want to expose those for consumers to handle and potentially restart or whatever, there is: https://docs.rs/thiserror instead.

alexjg: Yeah that's my understanding as well. In this case `anyhow` feels appropriate as the CLI is an application consuming the various other crates which are the library.

automerge-cli/src/export.rs
alexjg: :boom:

automerge-cli/src/export.rs
scotttrinh: Yeah, I've seen peripherally that error handling has been a big discussion point in Rust lately. I don't have the context to make an informed decision here, so I'm happy to explore `anyhow` here if that's helpful to the overall effort of establishing a pattern! I imagine that library code has different needs than the CLI for error handling, but perhaps it'll turn up some helpful general-purpose discussion?

alexjg: Cool, let's give it a shot and see how it works out.

automerge-cli/src/export.rs
alexjg: Hmm, there are a bunch of different approaches. The easiest way is probably to add a consuming method to the frontend which returns the state and drops the rest. I think we can probably punt on this until we have more understanding of what the usecases for the CLI look like though.

automerge-cli/Cargo.toml
scotttrinh: ⚡ https://github.com/automerge/automerge-rs/pull/17/commits/33906bd15957e2e2e60d4a23ff86619670d2e238

automerge-cli/src/export.rs
scotttrinh: ⚡ https://github.com/automerge/automerge-rs/pull/17/commits/89ca6c70b743fa62b556d62c8a0a115d57ef27c8

automerge-cli/src/export.rs
```diff
@@ -0,0 +1,42 @@
...
+    let mut input_data = Vec::new();
+    let mut input_file = File::open(changes_file)?;
+    input_file.read_to_end(&mut input_data)?;
+
+    Ok(input_data)
+}
+
+fn get_patch(changes_bytes: Vec<u8>) -> Result<Patch, AutomergeError> {
+    let mut backend = automerge_backend::Backend::init();
+    backend.load_changes_binary(vec![changes_bytes]).unwrap();
```
orionz, line 10: `load_changes_binary` followed by `get_patch` is semantically equivalent to `apply_changes_binary` and slower

automerge-cli/src/export.rs
```diff
@@ -0,0 +1,67 @@
...
+        )
+        .unwrap();
+        let value: automerge_frontend::Value =
+            automerge_frontend::Value::from_json(&initial_state_json);
+
+        let (_, initial_change) =
+            automerge_frontend::Frontend::new_with_initial_state(value).unwrap();
+        println!("{:?}", initial_change);
+        let mut backend = automerge_backend::Backend::init();
+        backend.apply_local_change(initial_change).unwrap();
```
scotttrinh, line 54: Panicking here for nested maps, but not for single-level maps.

scotttrinh: Fixed in d99b5d0 🎉

automerge-cli/src/export.rs
```diff
@@ -0,0 +1,86 @@
...
+use std::fs::read;
+use std::path::Path;
+
+fn get_state_json(input_data: Vec<u8>) -> Result<serde_json::Value> {
+    let mut backend = automerge_backend::Backend::init();
+    let changes = automerge_backend::Change::parse(&input_data)?;
+    let patch = backend.apply_changes(changes)?;
+
+    let mut frontend = automerge_frontend::Frontend::new();
+    frontend.apply_patch(patch).unwrap();
```
alexjg, line 11: Can we avoid this unwrap?

scotttrinh: Definitely!

PR #18, opened 2020-05-20
protocol: Hex-encode random ActorID

scotttrinh:
Ran into this issue when using `Frontend::new_with_initial_state` where it would panic when it was trying to decode a bare UUID as a hex-encoded string.

### DIFF
--- a/automerge-protocol/src/lib.rs
+++ b/automerge-protocol/src/lib.rs
@@ -12,7 +12,7 @@ pub struct ActorID(pub String);
 
 impl ActorID {
     pub fn random() -> ActorID {
-        ActorID(hex::encode(uuid::Uuid::new_v4().to_string()))
+        ActorID(hex::encode(uuid::Uuid::new_v4().as_bytes()))
     }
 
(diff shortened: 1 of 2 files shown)


automerge-protocol/src/lib.rs
```diff
@@ -12,7 +12,7 @@ pub struct ActorID(pub String);
 
 impl ActorID {
     pub fn random() -> ActorID {
-        ActorID(uuid::Uuid::new_v4().to_string())
+        ActorID(hex::encode(uuid::Uuid::new_v4().to_string()))
```
alexjg, line 15: It would probably be better to directly encode the `as_bytes` of the UUID I think. Something like:

```
ActorID(hex::encode(uuid::Uuid::new_v4().as_bytes()))
```

scotttrinh: Sure thing!

scotttrinh: ⚡ https://github.com/automerge/automerge-rs/pull/18/commits/a28f3990c7874d7e140b680137529f6b201ad6a5

PR #19, opened 2020-05-21
Canonical BinChange

orionz:
Ok - this is a pretty disruptive change so I really want to make sure i get it right and want some feedback.  I've introduced a new type `BinChange` which replaces change.  Basically its just a wrapper around the Vec<u8> we were using before but having validated it and stored some metadata.  All the methods that take and receive changes now use `BinChange` instead.   There is a `BinChange::extract()` function that will convert a Vec<u8> containing many changes into many BinChanges or an error.  (Not sure if extract is the right method name there - just went with something).  Change still exists but really is only useful for writing tests.  It has a `to_bin()` method and `BinChange` has a `to_change()` method.  I also implemented `From` on each other - still deciding which is best.  BinChange can be created with `BinChange::from_bytes(Vec<u8>)` or `slice.into()`.   I moved BinChange and all the encoding and columnar code into `automerge-protocol`.  I tool the time to update all the tests and make sure they work and did a pass with clippy.  So please check this out and try it with your code in progress and make sure we'll have as smooth of an update here as possible...


scotttrinh:
RE: effect on the ongoing CLI work:

The diff that this requires is pretty trivial:

```diff
@@ -4,7 +4,8 @@ use std::path::Path;
 
 fn get_state_json(input_data: Vec<u8>) -> Result<serde_json::Value> {
     let mut backend = automerge_backend::Backend::init();
-    let patch = backend.apply_changes_binary(vec![input_data])?;
+    let changes = automerge_protocol::BinChange::extract(&input_data)?;
+    let patch = backend.apply_changes(changes)?;
 
     let mut frontend = automerge_frontend::Frontend::new();
     frontend.apply_patch(patch).unwrap();
```

RE: any thoughts from me:

I don't have quite the right context and experience to provide any meaningful feedback on the changes here.

Since `BinChange` is the primary type and `Change` is mostly internal and for testing, would it make more sense to rename them the other way around? As a user or contributor coming into the codebase I might assume `Change` is the thing I'm mostly concerned with, and looking at the tests would confirm this (incorrect) assumption. No concrete suggestion here, but maybe the new `BinChange` could be `Change` and the old `Change` could be named to suggest that it's some kind of literal representation of that `Change`? I realize that's a lot more code changes 😅

alexjg:
I really like how much nicer this makes the backend tests to read. Definitely feels like a step in the right direction. Some thoughts:

Firstly I agree with Scott that the naming might be better the other way around, something like `Change` and `UnencodedChange` and an accompanying `UnencodedChange::encode` method?. I'm not a huge fan of `BinChange` because binary/non-binary is a dichotomy that makes sense in javascript - where binary data is special - but in Rust it's a bit of a strange one. If we were to stick with the current naming then maybe something like `CompressedChange` would make more sense?

Re. `extract`, would `parse` make sense?

Finally, I still think it makes sense to pull the change data structures out of the `protocol` crate entirely. They feel like they are mostly a concern of the backend crate. I've managed to write the frontend crate without mentioning changes at all so far. I am assuming that most clients will be directly passing `Vec<u8>` to whatever API we expose?

automerge-protocol/Cargo.toml
```diff
@@ -8,6 +8,8 @@ edition = "2018"
 
 [dependencies]
 hex = "^0.4.2"
+sha2 = "^0.8.1"
```
alexjg, line 11: Do these still need to be here now that we've moved the encoding back to the backend crate?

orionz: good catch

PR #21, opened 2020-05-27
cli: Add import command

scotttrinh:
Still to come:

- [ ] Integration tests between `import | export | import`
- [ ] Unit test for `import`


automerge-cli/src/main.rs
```diff
@@ -38,9 +40,49 @@ enum Command {
...
+        input_file: Option<PathBuf>,
+
+        /// Path to write Automerge changes to
+        #[clap(parse(from_os_str), long("out"), short("o"))]
+        changes_file: Option<PathBuf>,
+    },
+}
+
+fn open_file_or_stdin(maybe_path: Option<PathBuf>) -> Result<Box<dyn std::io::Read>> {
+    if atty::is(atty::Stream::Stdin) {
```
alexjg, line 63: Is this logic the wrong way around? It looks to me like it's rrequiring a path if stdin _is_ a tty, when we want to require a path if it isn't a tty right?

alexjg: Er, no wait I'm condused.

scotttrinh: I'm glad you made the exact same mistake as me! I originally wrote it backwards and was scratching my head for a whole 5 minutes.

automerge-frontend/src/change_context.rs
```diff
@@ -123,7 +123,7 @@ impl<'a> ChangeContext<'a> {
     /// change is complete we can `commit` the change context.
     ///
     pub fn apply_diff(&mut self, diff: &Diff) -> Result<(), InvalidPatch> {
-        println!("The diff: {:?}", diff);
+        // println!("The diff: {:?}", diff);
```
scotttrinh, line 126: Had to comment this line out since it ended up in the changes file when importing. 😅

alexjg: Ah, that definitely shouldn't be there, I must have accidentally left it in whilst debugging. Please feel free to remove it.

automerge-cli/Cargo.toml
```diff
@@ -10,8 +10,11 @@ path = "src/main.rs"
 
 [dependencies]
 clap = "3.0.0-beta.1"
+serde_json = "^1.0"
+anyhow = "1.0"
+atty = "0.2"
+duct = "0.13"
```
alexjg, line 16: Can we make this a dev dependency?

scotttrinh: oh, good call. didn't realize rust had the distinction!

scotttrinh: ⚡ https://github.com/automerge/automerge-rs/pull/21/commits/08d79fefe00e5f3cf954f4bce7b26dda0b50f0ae

automerge-frontend/src/change_context.rs
scotttrinh: ⚡ https://github.com/automerge/automerge-rs/pull/21/commits/5e2d74d13e7166c22950e4520da00f32bc76d01c

PR #25, opened 2020-06-09
cli: Add mutation CLI subcommands to IDEAS documentation

scotttrinh:
As discussed with @alexjg , we want to tackle mutation next, so added some examples for `set` and `get`.

### DIFF
--- a/automerge-cli/IDEAS.md
+++ b/automerge-cli/IDEAS.md
@@ -17,6 +17,9 @@
   ]
   $ automerge import foo.json --out bar.mpl
   $ automerge export foo.mpl | automerge import - > bar.mpl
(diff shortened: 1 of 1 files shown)


scotttrinh:
Should we consider something more standard like [JSON Path](https://goessner.net/articles/JsonPath/)?

alexjg:
Hmm, possibly. JSON Path is actually a pretty expansive standard, I'm uncertain whether we would need most of the things it offers but it's definitely worth looking at if this particular feature becomes something we use a lot. Right now I think we actually just need a very small sub set of it, more like that which is offered by JSON pointer.

I'm not too concerned either way though, what I'm really interested in right now is getting something using the mutation stuff so we can see how that API should work.

Issue #33, opened 2020-11-06
Browser/Node.js packaging story


nornagon:
I've been thinking about the best way to go about packaging automerge-rs such that it can become the One True Automerge Backend, used by default in mainline automerge, and installable with no more fuss than an `npm install`. Here's what I've learned so far, and what I think the path forward could be.

The majority of the groundwork for this is already laid by the `automerge-backend-wasm` crate, using `wasm-pack`. The tricky part is getting the wasm bundle that `wasm-pack` generates to function seamlessly in the various JS environments that automerge currently supports (i.e. Node.js and web browsers via a bundler such as webpack). The tooling in this area seems to be relatively immature (see e.g. https://github.com/rustwasm/wasm-bindgen/issues/2265), so sadly this is cutting-edge rust+wasm+js technology. I expect this story to improve significantly in the coming years.

wasm-pack is able to generate output for a variety of different targets, most importantly, `"bundler"`, `"nodejs"` and `"web"`. Notably, none of these output formats work for both Node.js and the browser. See https://github.com/rustwasm/wasm-pack/pull/705 for a long discussion and a variety of workarounds. In particular, I think this workaround is probably the best way forward for automerge-rs: https://github.com/rustwasm/wasm-pack/pull/705#issuecomment-529705607

e.g.

> package.json:
> 
> ```json
> {
>   "name": "automerge-backend-wasm",
>   "files": [
>     "pkg-bundler/index_bg.wasm",
>     "pkg-bundler/index.js",
>     "pkg-bundler/index.d.ts",
>     "pkg-node/index_bg.wasm",
>     "pkg-node/index.js",
>     "pkg-node/index_bg.js",
>     "pkg-node/index.d.ts"
>   ],
>   "main": "pkg-node/index.js",
>   "module": "pkg-bundler/index.js",
>   "types": "pkg-bundler/index.d.ts",
>   "sideEffects": "false"
> }
> ```

and then building with:

> ```
> wasm-pack build --target bundler --out-dir pkg-bundler
> wasm-pack build --target nodejs --out-dir pkg-node
> ```

The downside of this approach is that it builds two distinct wasm bundles. The bundles are nearly identical but differ in a few minor details. This is a fairly minor downside I think, as in a production app, in Node.js install time is not a big concern, and in a web app, the bundler will only pull the web version of the wasm bundle to the client.

With this in place, the following should work:

```js
import wasmBackend from 'automerge-backend-wasm'
import Automerge from 'automerge'

[...]

PR #34, opened 2020-11-06
package for node and bundler

nornagon:
Based on investigations detailed in #33.

With this PR, and the package published to npm under `automerge-backend-wasm`,
the following would work:

```js
import wasmBackend from 'automerge-backend-wasm'
import Automerge from 'automerge'

wasmBackend.initCodecFunctions(Automerge)
Automerge.setDefaultBackend(wasmBackend)

// ... use Automerge, but it's fast now
```

I think we can follow up to remove the need to call `initCodecFunctions`, but
this works for now!


nornagon:
hm, actually it looks like this doesn't work with webpack as-is, due to the async nature of webassembly loading. more investigation required!

nornagon:
I got this to work with the old-style synchronous webpack wasm support, but the new-style async module version has issues with recursive imports.

sync version: https://gist.github.com/nornagon/5d0ba28f9509a85604628e70406f3f3c

with `--experiments-async-web-assembly`, an exception is thrown:

```
test_mjs.js:156 Uncaught (in promise) ReferenceError: Cannot access '__wbindgen_json_parse' before initialization
    at Module.__wbindgen_json_parse (test_mjs.js:156)
    at test_mjs.js:673
    at :5000/async http:/localhost:5000/test_mjs.js:187
    at :5000/async http:/localhost:5000/test_mjs.js:105
    at :5000/async http:/localhost:5000/test_mjs.js:22
    at :5000/async http:/localhost:5000/test_mjs.js:829
```

it looks like this is due to the recursive dependency between `index_bg.js` and `index_bg.wasm` generated by wasm-pack:

index_bg.js:
```js
import * as wasm from './index_bg.wasm';
// ...
```

index_bg.wasm (via wasm2wat):
```wat
// ...
  (import "./index_bg.js" "__wbindgen_json_parse" (func (;0;) (type 1)))
// ...
```

I think this might be a webpack bug. async modules and the experimental async webassembly support are both still very new (less than a month old) so it wouldn't be too surprising if that were the case.

nornagon:
Hm, the async version seems to work properly if i `yarn pack`, untar the file and then `import wasmBackend from './package/index-bundler.js'`. So perhaps the webpack issue is a problem with `yarn link` and wouldn't be an issue if these packages were properly published to npm.

echarles:
> Hm, the async version seems to work properly if i yarn pack and untar the module into node_modules (vs. yarn link). So perhaps the webpack issue is a problem with yarn link and wouldn't be an issue if these packages were properly published to npm.

I did not have good experience with yarn link, and moved in general to lerna. 

With this PR, will consumers need to expcitely import `automerge-backend-wasm/index-node.js` for node applications and  `automerge-backend-wasm/index-bundler.js` for e.g. webpack applications?

orionz:
Hey guys - I did a bunch of experiments with different packaging but hit a dead end.  One, Im inexperienced with webpack - most of my experience with javascript is node/commonjs only.  Second I have some concerns about how people would want to consume a package like this - I posted my notes in a readme here - https://github.com/orionz/automerge-wasm-testapp

echarles:
For sync wasm in webpack, I have used

```
module.exports = {
  experiments: {
      syncWebAssembly: true,
  }
};
```

I now read on https://webpack.js.org/configuration/experiments/#experiments

syncWebAssembly: Support the old WebAssembly like in webpack 4
asyncWebAssembly: Support the new WebAssembly according to the updated specification, it makes a WebAssembly module an async module

tylersamples:
Maybe a naive question, but in the solutions presented it seems like there is a focus on working against the default behavior of browsers (i.e. trying to get the import to be sync or at the very least seem sync,) is there a reason why the `Automerge.setDefaultBackend` shouldn't be responsible for resolving the import (if it's necessary) instead of trying to introduce multi-layered packaging or shims?  If it has to be sync, has anyone explored proxies?

matrushka:
I might be dropping in without being accurate about my comments but recently tried to run this with [Vite](https://vitejs.dev/), it was only possible with lots of modifications and would like to share some ideas.

I previously I tried [sql.js](https://sql.js.org/#/?id=usage) and running it with wasm was a breeze. Maybe something like that can work here as well to make the whole experience easier.

I tried using `web` target for for [wasm-pack](https://rustwasm.github.io/docs/wasm-bindgen/reference/deployment.html#without-a-bundler) and was able to get to a working solution (although I had to manually change some overlapping function names in the generated JS file). In my case my code looked like:

```js
import generateBackend from "automerge-backend-wasm/web";

const run = async () => {
  const backend = await generateBackend("/node_modules/automerge-backend-wasm/web/index_bg.wasm");
  Automerge.setDefaultBackend(backend);
  const doc = Automerge.from({});
  const newDoc = Automerge.change(doc, (docToChange: any) => {
    docToChange.list = [];
  })
};

run();
```

Although it's working now (at least creting an empty document), I somehow ended up with an `Automerge error: Decode failed` when a change triggers `applyLocalChange` method of the backend called (so I can't make any changes to a document). Not really sure it this is due to the way I made it work though.

Based on this little journey, adding an additional web target to the package looks like a lot easier way than finding & documenting the right configs for different bundlers or bundler versions over time (at least until bundlers align on the implementation of the standard).

Issue #50, opened 2021-03-14
Getting system_time on wasm32 target


jeffa5:
Currently the system_time is obtained through this function.

https://github.com/automerge/automerge-rs/blob/c103b0638e24076e6cd755f102f3c75926c25d51/automerge-frontend/src/lib.rs#L417-L426

To be able to use this in a web project (using seed) I'm working on I've had to change this to use `chrono`:

https://github.com/jeffa5/automerge-rs/blob/b290f9117aced3e3ed1535ed0ccde3a819d88232/automerge-frontend/src/lib.rs#L420-L427

As there was already a comment I wonder whether you'd had ideas about how best to solve the issue and if chrono is a possible option for this project?

alexjg:
Interesting, what leads you to use the Rust frontend in a web environment? I would have assumed the JS frontend would be a better fit?

That aside I'm not sure about bringing in chrono. I love the library but I'm hesitant to add a dependency on it to Automerge for the sake of a single timestamp. I would be more inclined to make the way we obtain the timestamp pluggable so that application developers in `no_std` environments can use something other than `time::SystemTime`.

jeffa5:
I'm building the entire frontend in Rust so trying not to use js. Basically for the type safety and other nice things in Rust. Also nice to be able to use similar shared logic for frontend and backend.

Yep, adding a dependency on it may be a bit much but would be good to have a way to do custom time in `no_std`.

If you had any draft ideas I wouldn't mind having a go at a PR :)

alexjg:
These are the things I think such a PR should achieve

- Just Work™ if compiling on a platform with `time::SystemTime` support. i.e the current situation
- If compiling for `no_std` then require an additional argument when creating the frontend which is a function that can supply timestamps
- Provide a way to customize the timestamp supplier even when not compiling for `no_std` 

I think that once WASI is a bit more mature we shouldn't need this for browser targets as `time::SystemTime` has a WASI implementation.

The work then is this:
-  Add a constructor to the frontend (in `automerge-frontend/src/lib.rs`) which accepts a timestamp supplier, add this supplier as a field on the frontend and feed it through to the places where `time::SystemTime::now` is currently called. 
- Modify the current constructor so that it is only available when compiling via no_std and it passes the default `time::SystemTime::now` through to the new constructor

PR #51, opened 2021-03-16
Nostd time

jeffa5:
Fixes https://github.com/automerge/automerge-rs/issues/50

I've gone with using a `Box` rather than a type variable on the `Frontend` struct to maintain compatibility with existing code.

Also had to manually implement `Debug` for `Frontend` since the `timestamper` doesn't implement it.


alexjg:
This is exactly what I was imagining, thank you!

PR #56, opened 2021-03-21
backend: Check predecessor exists

jeffa5:
Ensure that the predecessor exists before trying to do operations on it.
If it doesn't exist then we can't do things with it so return an error
for the group of operations.


jeffa5:
I think my issue which caused this should be fixed by https://github.com/automerge/automerge-rs/pull/63 so I'll close this for now

PR #63, opened 2021-03-26
Fix issue in RleDecoder with null runs

### LINKED ISSUE #57: Flaky behaviour on save then load backend
I've come across a situation where the backend loading returns a `MissingObjectError` only sometimes. This seems rather strange behaviour to observe.

I've got it being flaky in a test case [here](https://github.com/jeffa5/automergeable/blob/fb2dfb96bb76255b6c931823732ca06e8bae8ac9/automergeable/tests/quickcheck.rs#L381-L519). `cargo test broken_save_load` a few times should give the error.

I found this via quickcheck so that might explain the randomness of the values.

The changes observed seem to be these:
```sh
# for the initial creation of  a value (from Null to this) 
changes: [LocalChange { path: Path([]), operation: Set(Map({"\u{2}": Sequence([Primitive(Null), Primitive(Uint(0)), Primitive(Str("")), Primitive(Counter(0)), Primitive(Str(""))]), "\u{0}": Sequence([Primitive(Counter(0)), Primitive(Str("")), Primitive(Uint(0)), Primitive(Timestamp(0)), Primitive(Int(0)), Primitive(Uint(0)), Primitive(Null), Primitive(Null), Primitive(F32(0.0)), Primitive(Null), Primitive(Counter(0)), Primitive(Uint(0)), Primitive(Null), Primitive(Uint(0)), Primitive(Str("")), Primitive(Null), Primitive(Timestamp(0)), Primitive(Timestamp(0)), Primitive(Uint(0)), Primitive(Counter(0)), Primitive(Uint(0)), Primitive(F32(0.0)), Primitive(Str(""))]), "\u{0}\u{0}": Sequence([Primitive(Str("")), Primitive(Counter(0)), Primitive(Str("")), Primitive(Boolean(false)), Primitive(Timestamp(0)), Primitive(Int(0)), Primitive(F64(0.0)), Primitive(Timestamp(0)), Primitive(Null), Primitive(Uint(0)), Primitive(F64(0.0)), Primitive(Boolean(false)), Primitive(F64(0.0)), Primitive(F64(0.0)), Primitive(Null), Primitive(F64(0.0)), Primitive(F64(0.0))]), "": Sequence([Primitive(Null), Primitive(Uint(0)), Primitive(Int(0)), Primitive(Null), Primitive(F32(0.0)), Primitive(F64(0.0)), Primitive(Uint(0)), Primitive(F64(0.0)), Primitive(Timestamp(0)), Primitive(Str("")), Primitive(Boolean(false)), Primitive(Counter(0)), Primitive(Int(0)), Primitive(Null), Primitive(F64(0.0)), Primitive(Null), Primitive(F64(0.0)), Primitive(Counter(0)), Primitive(Boolean(false))]), "\u{1}": Map({"": Primitive(F64(0.0))}, Table)}, Map)) }]

# for the change from previous to this value
changes: [LocalChange { path: Path([Key("\u{2}")]), operation: Delete }, LocalChange { path: Path([Key("\u{0}")]), operation: Delete }, LocalChange { path: Path([Key("\u{0}\u{0}")]), operation: Delete }, LocalChange { path: Path([Key("")]), operation: Delete }, LocalChange { path: Path([Key("\u{1}")]), operation: Delete }]
```

jeffa5:
I've managed to hunt down this bug with another more reliable test case and found that the RLE decoder was reading the length of the null run as `isize` when it is encoded as `usize`. This becomes an issue when the length is above a certain value (64 here). Then the length gets read as negative and the decoder enters an infinite loop of repeating the null value, even when there is more input.

Maybe in future we shouldn't use `*size` either but `*64` or `*32` explicitly so different platforms don't have issues saving/loading shared documents. I'll leave that for another time.

Fixes #57

### DIFF
--- a/automerge-backend/src/columnar.rs
+++ b/automerge-backend/src/columnar.rs
@@ -1190,3 +1190,97 @@ const DOCUMENT_COLUMNS = {
   extraRaw:  5 << 3 | COLUMN_TYPE.VALUE_RAW
 }
 */
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_rle_encoder_for_strings_from_key() {
+        // this seems like a strange case but checks that we write nulls into the encoder as usize and read them out the same.
+        // if we don't then the 64 nulls gets interpreted as -64 and causes the rle decoder to never read the next values.
+        let ops = vec![
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            None,
+            Some("a".to_owned()),
+        ];
+        let mut encoder = RleEncoder::new();
+        for op in &ops {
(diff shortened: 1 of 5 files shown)


alexjg:
Nice catch, and yes I agree, we should switch to using `u64` over `usize`.

PR #64, opened 2021-03-26
Simplify if statement to remove unreachable case

jeffa5:
If `op_by_id` contains `succ` then we can get the index straight away,
else we can do the original work. This means we don't need an `else`
case for a third situation as in reality there isn't one.


orionz:
Looks great!  Thanks jeff!

PR #68, opened 2021-03-27
Simplify while stack not empty with while let

jeffa5:


alexjg:
This looks like it's based off of the compression PR which I don't want to merge until the corresponding PR in the JS `performance` branch is merged. It all looks good otherwise so happy to merge once that's done.

jeffa5:
Ah, I don't think it relies on that branch, rebased it to just be off main. My fault.

PR #72, opened 2021-03-28
Use grapheme clusters in Text instead of char

### LINKED ISSUE #44: Rebuilding a frontend from a backend fails when using UTF-8 strings
Hi, 

For the project I work on, we often rebuild an automerge frontend from an automerge backend.

I noticed a crash when using UTF-8 characters inside a string. Here are the steps to reproduce : 

#### Reproduction 
```rust 

// This is an example of test string that fails. More examples below.
let test_str = String::from("🌍🌎🌏");

// Create a backend and frontend
let mut backend = automerge_backend::Backend::init();
let mut frontend = automerge_frontend::Frontend::new();

// Main data structure
let mut hashmap: HashMap<String, automerge_frontend::Value> = HashMap::new();
hashmap.insert(
    "key1".to_string(),
    automerge_frontend::Value::Text(test_str.chars().collect()),
);

// Create a "change" action, that sets the hashmap as root of my automerge document.
let change = automerge_frontend::LocalChange::set(
    automerge_frontend::Path::root(),
    automerge_frontend::Value::Map(hashmap, automerge_protocol::MapType::Map),
);

// Apply this change on the frontend
let change_request = frontend
    .change::<_, automerge_frontend::InvalidChangeRequest>(
        Some("set root object".into()),
        |frontend| {
            frontend.add_change(change)?;
            Ok(())
        },
    )
    .unwrap();

// We can notice that the state of the frontend is OK
println!("frontend state : {:?}", frontend.state());

// Trying to retrieve the root dict - this works with the original frontend
let root_dict: automerge_frontend::Value = frontend
    .get_value(&automerge_frontend::Path::root())
    .unwrap();

println!("root_dict {:?}", root_dict);

// Apply it on the backend
backend
    .apply_local_change(change_request.unwrap())
    .unwrap()
    .0;

// Create a new frontend
frontend = automerge_frontend::Frontend::new();

// Rebuild the frontend from the backend
frontend.apply_patch(backend.get_patch().unwrap());

// We can notice that the state of the frontend is NOT OK. 
println!("frontend state : {:?}", frontend.state());

// Trying to retrieve the root dict - this crashes, as the frontend has not been rebuilt properly.
let root_dict: automerge_frontend::Value = frontend
    .get_value(&automerge_frontend::Path::root())
    .unwrap();

```
#### Test cases 
Here is the list of all the strings I have tested and their status running the code above : 

| Test | Status
| --- | --- 
| `"Hello"` | Works
| `"السلام عليكم"` | Fails
| `"Dobrý den"` | Fails
| `"שָׁלוֹם"` | Fails
| `"नमस्ते"` | Fails
| `"こんにちは"` | Fails
| `"안녕하세요"` | Fails
| `"你好"` | Fails
| `"Olá"` | Fails
| `"Здравствуйте"` | Fails
| `"🌍🌎🌏"` | Fails

#### Stacktrace
I'm not sure it is very relevant, but I add it anyway. Notice that we use PyO3 to bind to python.

```
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', src/hashmap2.rs:283:14
stack backtrace:
   0:        0x1103905e4 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hcfc48256a5ab8835
   1:        0x1103ab9b0 - core::fmt::write::haf3903118f694c48
   2:        0x11038e0b6 - std::io::Write::write_fmt::h7385463ac87804ed
   3:        0x110391cef - std::panicking::default_hook::{{closure}}::h91bd4c58cf71392b
   4:        0x1103919bd - std::panicking::default_hook::h7bd29c87df967048
   5:        0x1103922bb - std::panicking::rust_panic_with_hook::hae2b05f08a320721
   6:        0x110391e3b - std::panicking::begin_panic_handler::{{closure}}::h72d68d3a77e0b718
   7:        0x110390a58 - std::sys_common::backtrace::__rust_end_short_backtrace::h7c5e286792f94edb
   8:        0x110391dfa - _rust_begin_unwind
   9:        0x1103c012f - core::panicking::panic_fmt::h1b194bb80d76fb10
  10:        0x1103c0087 - core::panicking::panic::hdb9dddaff64fd68b
  11:        0x11029f133 - jupyter_rtc_automerge::hashmap2::HashmapDocument::test_utf8::h193fd5e9fe887717
  12:        0x110296977 - jupyter_rtc_automerge::hashmap2::__init8698952135223600566::__wrap::{{closure}}::h79f502163f7d59bc
  13:        0x1102a20fc - jupyter_rtc_automerge::hashmap2::__init8698952135223600566::__wrap::h287fa5e7c76fc49c
  14:        0x10fe6e4e8 - __PyMethodDef_RawFastCallKeywords
  15:        0x10fe7ae24 - __PyMethodDescr_FastCallKeywords
  16:        0x10ffaad65 - _call_function
  17:        0x10ffa79ed - __PyEval_EvalFrameDefault
  18:        0x10ff9c34a - __PyEval_EvalCodeWithName
  19:        0x10ffffb80 - _PyRun_FileExFlags
  20:        0x10fffeff7 - _PyRun_SimpleFileExFlags
  21:        0x11002cccf - _pymain_main
  22:        0x10fe408dd - _main
```

If I can provide any other information to help you fix this, please let me know.

Best regards.

jeffa5:
`char` is not what users typically expect as a 'letter' so we should use
grapheme clusters instead. These can't always be represented as single
rust `char` types so the `Text` now stores a `Vec` of single grapheme `String`s.

Fixes #44 

While this makes sense from the rust side I'm not sure if it has an effect on the javascript implementation compatability. Also, not sure of its affect on the binary encoding from the javascript side again.

### DIFF
--- a/automerge-frontend/Cargo.toml
+++ b/automerge-frontend/Cargo.toml
@@ -15,6 +15,7 @@ uuid = { version = "^0.8.2", features=["v4"] }
 maplit = "1.0.2"
 thiserror = "1.0.16"
 im-rc = "15.0.0"
+unicode-segmentation = "1.7.1"
 
 [dev-dependencies]
 automerge-backend = { path = "../automerge-backend" }
--- a/automerge-frontend/src/mutation.rs
+++ b/automerge-frontend/src/mutation.rs
@@ -3,6 +3,7 @@ use crate::state_tree::{LocalOperationResult, SetOrInsertPayload, StateTree, Tar
 use crate::value::{Cursor, Primitive, Value};
 use crate::{Path, PathElement};
 use automerge_protocol as amp;
+use unicode_segmentation::UnicodeSegmentation;
 
 pub trait MutableDocument {
     fn value_at_path(&self, path: &Path) -> Option<Value>;
@@ -189,11 +190,11 @@ impl MutableDocument for MutationTracker {
                             }
                             (PathElement::Index(i), Target::Text(ref text)) => match value {
                                 Value::Primitive(Primitive::Str(s)) => {
-                                    if s.len() == 1 {
+                                    if s.graphemes(true).count() == 1 {
                                         let payload = SetOrInsertPayload {
                                             start_op: self.max_op + 1,
                                             actor: &self.actor_id.clone(),
-                                            value: s.chars().next().unwrap(),
+                                            value: s.clone(),
                                         };
                                         self.apply_state_change(text.set(*i, payload)?);
                                         Ok(())
@@ -339,11 +340,11 @@ impl MutableDocument for MutationTracker {
                             }
                             (Target::Text(text_target), val) => match val {
                                 Value::Primitive(Primitive::Str(s)) => {
-                                    if s.len() == 1 {
+                                    if s.graphemes(true).count() == 1 {
                                         let payload = SetOrInsertPayload {
                                             start_op: self.max_op + 1,
                                             actor: &self.actor_id.clone(),
-                                            value: s.chars().next().unwrap(),
+                                            value: s.clone(),
                                         };
                                         self.apply_state_change(
                                             text_target.insert(*index, payload)?,
--- a/automerge-frontend/src/state_tree/diffable_sequence.rs
+++ b/automerge-frontend/src/state_tree/diffable_sequence.rs
@@ -1,4 +1,4 @@
-use super::{DiffApplicationResult, DiffToApply, MultiChar, MultiValue, StateTreeChange};
+use super::{DiffApplicationResult, DiffToApply, MultiGrapheme, MultiValue, StateTreeChange};
 use crate::error::InvalidPatch;
 use automerge_protocol as amp;
 use std::collections::HashMap;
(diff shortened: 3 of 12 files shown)


alexjg:
This is great stuff, I'm going to go ahead and merge. I am a little concerned about performance as the code paths relating to inserting and removing characters are some of the hottest parts of the frontend. However, it's going to change reasonably substantially when the new sequence patch format is merged (I have a branch with the implementation which I keep up to date with `main`, I'm just waiting for the JS branch to merge before putting in a PR). Once that's done then I think we should revisit performance and perhaps consider adding some benchmarks to CI.

ept:
Nice! I agree that text should best be segmented by grapheme cluster. I also looked into adding something like this on the JS side, but I was a bit hesitant about adding a dependency that would increase the code size. Am I right in understanding that the dependency on the unicode-segmentation crate is only in the Rust frontend, so it won't affect the size of the compiled wasm module?

In terms of compatibility between JS and Rust, I think both backends should allow each text "character" to be any sequence of unicode code points. The segmentation into grapheme clusters should happen once when the text is first inserted on the frontend, and thereafter everything else just preserves the segmentation that was initially performed. This gives us robustness in case the rules for grapheme cluster segmentation change (which I believe could happen from one version of the Unicode standard to the next).

alexjg:
Yeah this dependency will only be pulled in by applications using the Rust frontend, the backend is completely agnostic to the segmentations as you describe.

PR #78, opened 2021-03-30
Remove Rc from the backend

### LINKED ISSUE #75: Should the backend be Send + Sync?
Given the ideas proposing to have a single backend with multiple frontends I think it would make sense for the backend to be `Send` + `Sync`. This means we could send a backend across threads as well as use it in shared memory settings with, for example, a `Mutex`. I can see this being particularly useful for server implementations in Rust where clients for a document may not always land on the same thread.

Basically this comes down to replacing instances of `Rc` with either `Arc` or removing them if they aren't necessary. Also we'd need to switch from `im_rc` to `im` which has a minor performance penalty I believe.

I'm not sure of the impact of this on wasm but think it should still work the same. I've got an example branch of this [here](https://github.com/jeffa5/automerge-rs/tree/remove-rc)

jeffa5:
This mechanically removes `Rc` from the backend code. There weren't any major issues so it seems as though it wasn't needed.

If someone has more knowledge on why it was needed I'd be happy to hear.

Fixes #75

### DIFF
--- a/automerge-backend/src/backend.rs
+++ b/automerge-backend/src/backend.rs
@@ -9,21 +9,20 @@ use crate::Change;
 use automerge_protocol as amp;
 use core::cmp::max;
 use std::collections::{HashMap, HashSet};
-use std::rc::Rc;
 
 #[derive(Debug, PartialEq, Clone)]
 pub struct Backend {
-    queue: Vec<Rc<Change>>,
-    op_set: Rc<OpSet>,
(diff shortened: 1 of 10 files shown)


orionz:
These RC's were there because I wanted to make fork operations extremely fast and memory efficient.  Likely a premature optimization and something we can revisit later.  I think getting Sync and Send are more important right now.  Thanks for the patch!

PR #80, opened 2021-04-01
Backend: remove unnecessary draining of changes to be applied

gterzian:
A few nits, caught my attention when looking at https://github.com/automerge/automerge-rs/pull/78

### DIFF
--- a/automerge-backend/src/backend.rs
+++ b/automerge-backend/src/backend.rs
@@ -64,17 +64,15 @@ impl Backend {
         })
     }
 
-    pub fn load_changes(&mut self, mut changes: Vec<Change>) -> Result<(), AutomergeError> {
-        let changes = changes.drain(0..).collect();
+    pub fn load_changes(&mut self, changes: Vec<Change>) -> Result<(), AutomergeError> {
         self.apply(changes, None)?;
         Ok(())
     }
 
     pub fn apply_changes(
         &mut self,
-        mut changes: Vec<Change>,
+        changes: Vec<Change>,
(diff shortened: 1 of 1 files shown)


automerge-backend/src/backend.rs
```diff
@@ -84,12 +82,12 @@ impl Backend {
...
     fn apply(
         &mut self,
-        mut changes: Vec<Change>,
+        changes: Vec<Change>,
         actor: Option<(amp::ActorId, u64)>,
     ) -> Result<amp::Patch, AutomergeError> {
         let mut pending_diffs = HashMap::new();
 
-        for change in changes.drain(..) {
+        for change in changes.into_iter() {
```
gterzian, line 90: Here using `drain` would have made a lot of sense if you had a mutable reference to the vector, and you expected the caller to still use the empty one after the changes had been applied.

automerge-backend/src/lib.rs
```diff
@@ -32,17 +32,12 @@ pub use error::AutomergeError;
...
 mod tests {
-    use std::{
-        sync::{Arc, Mutex},
-        thread,
-    };
+    use std::thread;
 
     #[test]
-    fn sync_and_send_backend() {
+    fn test_backend_is_send() {
```
gterzian, line 38: It appears to me that it's sufficient to test that the backend is `Send`, since by definition anything that is can be made `Sync` by putting a lock around it. So the previous test doesn't actually test that the backend is `Sync`, it makes it so by putting the mutex around it. 

If one wanted to test for a `Sync` back-end, one could do it by putting only an `Arc` around it.

Issue #70, opened 2021-03-27
Update encoding/decoding logic to use *64 rather than *size


jeffa5:
This ensures we have consistent behaviour across 32bit and 64bit environments.

jeffa5:
Actually, I'm not sure whether we should actually use *32 so that things like `Vec` can be ensured to work on 32 bit targets.

orionz:
Im pretty sure - everywhere we use usize its as an index to a vector and thus usize makes sense.   I would expect a 32bit machine running automerge to break on more than MAX_INT ops or changes etc (but also run out of memory before then).

PR #82, opened 2021-04-01
Use 32 bits on binary encoding

jeffa5:
This changes the binary encoding to not use `*size` types but `*32` types. `*size` shouldn't be used since it is different sizes on different platforms so could lead to different issues. Also on 32 bit platforms we can't represent things like `Vec` to the same length as on 64 bit platforms so I think we should prefer to limit things where possible to 32 bits in length.

Relates to #70 

What are your thoughts?

### DIFF
--- a/automerge-backend/src/encoding.rs
+++ b/automerge-backend/src/encoding.rs
@@ -59,10 +59,16 @@ impl<'a> Decoder<'a> {
     }
 }
 
+/// Encodes booleans by storing the count of the same value.
+///
+/// The sequence of numbers describes the count of false values on even indices (0-indexed) and the
+/// count of true values on odd indices (0-indexed).
+///
+/// Counts are encoded as u32s.
 pub(crate) struct BooleanEncoder {
     buf: Vec<u8>,
     last: bool,
-    count: usize,
+    count: u32,
 }
 
 impl BooleanEncoder {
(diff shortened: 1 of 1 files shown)


ept:
Agree that it would be good to have consistent behaviour across both 32-bit and 64-bit platforms. I will note that this limits documents to a maximum of 2^32 operations in their history. At the moment this is not a problem since Automerge would probably get unusably slow long before this limit is reached, but at some point in the future we may want to support longer histories. For that reason I think it's okay if the current implementation uses 32-bit variables, but I think the specification of the binary format should allow values greater than 2^32.

orionz:
Is this desirable?  Under the current code a 64bit system could continue to run if it encountered a > 2^32 bit run on ops, but a 32bit system would fail.   Isn't that better than just making the 64bit system also fail?  (hobbling the 64bit system so it fails in the same way as a 32 bit one).  Is there another scenario I'm not thinking about?  I suppose its moot since we're likely never going to see a document of that size...

jeffa5:
I think using usize should be fine then as it should allow 64 bit systems to run with larger docs, and it seems the lower level uses the variable encoding (leb128) for all number types anyway.

I'll repost this as a documentation PR.

orionz:
Maybe we should put in an informative panic if usize max < the number that comes out of the leb128

PR #84, opened 2021-04-01
Allow change to return value from users closure

### LINKED ISSUE #83: Should the frontend return a users value?
The frontend `change` function takes a closure: https://github.com/automerge/automerge-rs/blob/d3d1b48c4889e374b58e723c16b3c13a5a5d30bb/automerge-frontend/src/lib.rs#L382-L389

This type forces the user to return a unit value in the case of success but they may want to do something with a value from inside this closure. This leads to them being forced to some more ugly style with 

```rust
let mut my_val = None; // or default
frontend.change(None, |doc| {
  my_val = doc.get(some_path);
  Ok(())
});
```

Which isn't particularly idiomatic.

I propose changing the signature of the change function to something like:

```rust
pub fn change<F, O, E>( 
     &mut self, 
     message: Option<String>, 
     change_closure: F, 
 ) -> Result<(O, Option<UncompressedChange>), E> 
 where 
     E: Error, 
     F: FnOnce(&mut dyn MutableDocument) -> Result<O, E>,
```

Such that we can see more idiomatic code such as:

```rust
let (my_val, change) = frontend.change(None, |doc| {
  Ok(doc.get(some_path))
}).unwrap();
```

What do you think?

jeffa5:
Fixes #83

### DIFF
--- a/automerge-frontend/src/lib.rs
+++ b/automerge-frontend/src/lib.rs
@@ -151,15 +151,15 @@ impl FrontendState {
     /// which it can use to query the document state and make changes. It
     /// can also throw an error of type `E`. If an error is thrown in the
     /// closure no chnages are made and the error is returned.
-    pub fn optimistically_apply_change<F, E>(
+    pub fn optimistically_apply_change<F, O, E>(
         self,
         actor: &ActorId,
         change_closure: F,
         seq: u64,
-    ) -> Result<OptimisticChangeResult, E>
+    ) -> Result<(O, OptimisticChangeResult), E>
     where
         E: Error,
-        F: FnOnce(&mut dyn MutableDocument) -> Result<(), E>,
+        F: FnOnce(&mut dyn MutableDocument) -> Result<O, E>,
     {
         match self {
             FrontendState::WaitingForInFlightRequests {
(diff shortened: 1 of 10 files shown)


automerge-frontend/src/lib.rs
```diff
@@ -379,18 +385,18 @@ impl Frontend {
         }
     }
 
-    pub fn change<F, E>(
+    pub fn change<F, O, E>(
         &mut self,
         message: Option<String>,
         change_closure: F,
-    ) -> Result<Option<UncompressedChange>, E>
+    ) -> Result<(O, Option<UncompressedChange>), E>
```
alexjg, line 392: We could add a type here instead of the tuple, something like `ChangeResult`. I'm not sure if it would really clarify anything though, the existing types probably communicate intent clearly enough as `O` is obviously the return type of `F`.

I also think the number of type parameters on `change` is becoming really unpleasant to work with, but I think solving this is probably a matter of building richer APIs on top of `change` (e.g the `automergeable` work).

automerge-frontend/src/lib.rs
```diff
@@ -151,15 +151,15 @@ impl FrontendState {
...
     /// can also throw an error of type `E`. If an error is thrown in the
     /// closure no chnages are made and the error is returned.
-    pub fn optimistically_apply_change<F, E>(
+    pub fn optimistically_apply_change<F, O, E>(
         self,
         actor: &ActorId,
         change_closure: F,
         seq: u64,
-    ) -> Result<OptimisticChangeResult, E>
+    ) -> Result<(O, OptimisticChangeResult), E>
```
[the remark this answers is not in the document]
jeffa5: I agree, updated.

automerge-frontend/src/lib.rs
jeffa5: I think it still makes sense to keep these values separate here, mainly since they serve two purposes, one for the user to do as they wish with and the other to be passed to a backend.

alexjg (approved):

PR #85, opened 2021-04-01
encoding: add documentation and handle RleDecoder errors more explicitly

jeffa5:


orionz:
Will this blow up on 32bit systems?  Is there a clean place to drop in an error message should that event ever arise?

jeffa5:
Since the numbers are encoded with leb I think the unwrap would generate a panic only if the leb number was larger than 32 bits and the cast to i64 should be fine on 32 bit platforms.

Basically, I'd say it is fine for now until we find someone experiencing an issue or build in some testing with specific architectures.

jeffa5:
I've reduced this to more of a documentation PR with a fixup for handling RleDecoder errors a bit better (rather than just treating them as null runs).

PR #88, opened 2021-04-05
Implement new sequence patch format

alexjg:
This is mostly complete, the only thing remaining to do is correct the diff serialization for the new edits format. I'll be fixing that this week. Once that's done we're waiting on merging the JS implementation.


orionz:
@alexjg - I suspect we'll be merging this is very soon - can you look at the conflicts so we can have a clean merge when needed?

alexjg:
I've rebased but the interop tests are failing hard, I'm working on fixing that. I'm also anticipating some changes based on Martin's work on the JS side of this, when that's stable I'll implement the changes here.

orionz:
Awesome man!  Thanks for the work

alexjg:
The JS implementation has landed now. I'll get to this at the weekend unless someone picks it up before then.

PR #91, opened 2021-04-05
add js interop testing

vedantroy:


automerge-backend-wasm/scripts/js_tests.sh
```diff
@@ -0,0 +1,14 @@
+#! /usr/bin/env bash
+
+if [[ -e "../automerge" ]]
+then
+    echo "Automerge repo exists. Skipping cloning..."
+else
+    echo "Cloning automerge repo to ../../automerge"
+    git clone https://github.com/automerge/automerge.git ../../automerge
```
alexjg, line 8: Can we checkout a particular commit to make these tests reproducible?

vedantroy: Done

automerge-backend-wasm/scripts/js_tests.sh
```diff
@@ -0,0 +1,15 @@
+#! /usr/bin/env bash
+
+if [[ -e "../../automerge" ]]
+then
+    echo "Automerge repo exists. Skipping cloning..."
```
alexjg, line 5: Is this likely to lead to issues where one has checked out the code but then don't update it when we change the `automerge` commit we depend on leading to tests that fail on a fresh repo but not on the out of date checkout? If so we should check the specific commit version in this script

automerge-backend-wasm/scripts/js_tests.sh
```diff
@@ -0,0 +1,15 @@
+#! /usr/bin/env bash
+
+if [[ -e "../../automerge" ]]
```
alexjg, line 3: This seems to hardcode the location of the `automerge` checkout. Can we instead clone it into the working directory and then add to `.gitignore`?

alexjg:
Can we add this checkout to `.travis.yml`?

automerge-backend-wasm/scripts/js_tests.sh
vedantroy: I imagine the following usage:
- Locally, you probably have an `automerge` folder in the same directory as `automerge-rs`. When developing, you don't want to blow away your local copy of `automerge` with the commit version specified in this script.
- In CI, you probably want to test with a specific commit version and there won't be a `automerge` folder in the same directory.

automerge-backend-wasm/scripts/js_tests.sh
vedantroy: This is set-up so that given a `automerge-rs` folder, this will checkout `automerge` as a folder underneath the same parent directory so there's no absolute paths (if that's what you mean by hardcode).

I'm not sure how we can avoid hard coding. Won't we always have to clone the `automerge` repository to some specified location?

vedantroy:
> Can we add this checkout to `.travis.yml`?

Is this equivalent to adding another line under the `script` section in `.travis.yml`?

automerge-backend-wasm/scripts/js_tests.sh
alexjg: Yeah what I'm suggesting is that we should clone it to a subdirectory of the `automerge-rs` repo, then add whatever directory we clone it to the `.gitignore`. I _think_ that should do the trick.

IMO We definitely shouldn't clone to somewhere outside the project directory. I think that's quite surprising behaviour that will annoy a lot of people and break build systems.

automerge-backend-wasm/scripts/js_tests.sh
alexjg: Ah interesting, I was assuming this would be a checkout specifically for testing. I can see how that might be difficult if you're developing both a JS and Rust change at once (which is not uncommon). 

I wonder if an approach like providing an environment variable to override the JS location would work. If the variable is set we skip the commit version check and just use whatever we find there.

That means the default case will be the same as CI so hopefully avoiding some of the more annoying reproducibility accidents that can happen because you have to explicitly set an env bar, which hopefully is visually obvious

automerge-backend-wasm/scripts/js_tests.sh
jeffa5, line 3:
```suggestion
if [[ -d "../../automerge" ]]
```
I'd expect `automerge` to be a directory

jeffa5: I agree that we should also just have it inside the automerge-rs directory somewhere. I wouldn't want it interfering externally.

automerge-backend-wasm/scripts/js_tests.sh
```diff
@@ -0,0 +1,15 @@
...
+then
+    echo "Automerge repo exists. Skipping cloning..."
+else
+    echo "Cloning automerge repo to ../../automerge"
+    git clone https://github.com/automerge/automerge.git ../../automerge
+fi
+
+ORIGINAL_PWD=$PWD
+cd ../../automerge
+git checkout 81079ff75d2234b47cb912bad728158f2e71c527
```
jeffa5, line 13: Should there be some sort of check to see if this commit even exists? (could have an old repo so could ask user to update)

automerge-backend-wasm/scripts/js_tests.sh
```diff
@@ -0,0 +1,15 @@
+#! /usr/bin/env bash
+
```
jeffa5, line 2: Could add a `set -e` here.

automerge-backend-wasm/scripts/js_tests.sh
```diff
@@ -0,0 +1,15 @@
...
+else
+    echo "Cloning automerge repo to ../../automerge"
+    git clone https://github.com/automerge/automerge.git ../../automerge
+fi
+
+ORIGINAL_PWD=$PWD
+cd ../../automerge
+git checkout 81079ff75d2234b47cb912bad728158f2e71c527
+yarn install
+env WASM_BACKEND_PATH="$ORIGINAL_PWD/build" yarn testwasm
```
jeffa5, line 15: This doesn't seem to build the backend at all so with just running `yarn test:js` I get an `Error: Cannot find module '/home/andrew/projects/automerge-rs/automerge-backend-wasm/build'`. I'd expect this to handle the dependencies of the task (like cloning automerge) including building the wasm backend.

jeffa5 (changes requested):

jeffa5:
> Is this equivalent to adding another line under the `script` section in `.travis.yml`?

I think it would go something like adding the following to the bottom of `script`:

```yaml
- travis_fold start "js-interop-test"
- cd automerge-backend-wasm && yarn test:js
- travis_fold end "js-interop-test"
```

automerge-backend-wasm/scripts/js_tests.sh
vedantroy: The reason I don't build the wasm backend is b/c sometimes we might want to build dev & sometimes prod (depending on how much time you have, etc.) so I left that up to the user. (It's 1 command: `yarn run build` or `yarn run release`).

Does this seem reasonable?

automerge-backend-wasm/scripts/js_tests.sh
vedantroy: Done

automerge-backend-wasm/scripts/js_tests.sh
vedantroy: Done

automerge-backend-wasm/scripts/js_tests.sh
vedantroy: Done

automerge-backend-wasm/scripts/js_tests.sh
vedantroy: Going to temporary resolve to trigger CI. (Not sure if unresolved issue is preventing CI from running?)

vedantroy:
@alexjg @jeffa5 I refactored the shell script so it does the following:
- If the user sets `AUTOMERGE_JS_DIR` then the script tries to use that directory to run the js-interop tests
- If the user doesn't set the above variable, the script will clone to a default location and check if the given commit hash exists

I also added this test to travis, but the CI isn't running so I don't know if it it's working or not.

automerge-backend-wasm/scripts/js_tests.sh
alexjg: I think the best thing we can do to ensure interoperability is have interop testing as part of CI. In that context I think we do need to automate the build step. 

I think the environment variable approach o alluded to earlier should help here for cases where the user wants a different build?

vedantroy: In CI (inside .travis.yml), I do:

```
cd automerge-backend-wasm && yarn build && yarn test:js
```

To build & then test js interop with the assumption that locally users can run `yarn build` or `yarn release` themselves since they want flexibility while developing. Should I also handle building while developing.

jeffa5: I'm happy to have the users do a `yarn build` or equivalent. A friendlier error message would be nice, e.g. if the `build` directory is missing then ask them to build, giving them an example of build or release.

automerge-backend-wasm/scripts/js_tests.sh
```diff
@@ -0,0 +1,34 @@
...
+        git clone https://github.com/automerge/automerge.git $AUTOMERGE_JS_DIR
+    fi
+    cd $ORIGINAL_PWD
+    cd $AUTOMERGE_JS_DIR
+    git checkout $COMMIT_HASH
+else
+    # if the env var is set, assume the user is using an existing checkout of automerge
+    echo "Using $AUTOMERGE_JS_DIR"
+    if [[ -d $AUTOMERGE_JS_DIR ]]; then
+        echo "$AUTOMERGE_JS_DIR dir not found."
```
jeffa5, line 26: This is the case when the directory does exist.

Also need to have an `exit 1` when the directory doesn't exist here.

jeffa5 (changes requested):
Looking good, thanks for adding the checks.

Just an inverted if statement.

automerge-backend-wasm/scripts/js_tests.sh
vedantroy: done

automerge-backend-wasm/scripts/js_tests.sh
vedantroy: fixed

vedantroy:
@jeffa5  fixed the bug + added friendlier error message. I also just realized that the script currently always runs tests with the release backend, so theoretically it could just build the release backend every time, but it might still be better to build the release backend outside the test script because I'm not sure if you always want to run interop tests locally whenever you build the release backend and if you already built the release backend it seems unnecessary to build it again for interop tests.

jeffa5 (approved):
Thanks Ved. All looks good now and I agree that the building can be done outside of this script for easier local usage. We can always rework things if needs be.

Thanks!

jeffa5:
Fixes https://github.com/automerge/automerge-rs/issues/89

PR #93, opened 2021-04-14
Add newer version of nodejs

jeffa5:
This makes the final stage of CI work.


jeffa5:
Using node version 10 as this is the minimum required and the minimum automerge tests with https://github.com/automerge/automerge/blob/main/.travis.yml

PR #94, opened 2021-04-14
Add rust cache for travis

jeffa5:
There are more stages to CI now so this hopes to reduce some of the time spent rebuilding dependencies and unchanged code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ install:
 - rustup self update
 - rustup component add clippy
 - rustup component add rustfmt
-- curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+- cargo install --force --version 0.9.1 wasm-pack
 - nvm install 10
 before_script:
 - export -f travis_fold
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 os: linux
 dist: xenial
 language: rust
+cache: cargo
 rust:
 - stable
 - beta


jeffa5:
This should pass after #93

alexjg:
I deliberately removed the rust cache because it caches a lockfile, this causes failures in CI which are hard to debug because you don't know what version of the various dependencies it's locked to and can't easily reproduce it locally. That said, caching is generally useful, so I think if we're going to turn on the cache then we need to also add a script which deletes the cached lockfile.

jeffa5:
Interesting, do you recall which lockfile it was or the path to it? I wouldn't expect the `Cargo.lock` file to be cached as per https://docs.travis-ci.com/user/languages/rust/#dependency-management it isn't in one of those directories.

orionz:
@alexjg - What are your thoughts?  Should we give this a shot?  We can always revert if it causes problems I suppose.

alexjg:
The issue I had was actually not the lockfile being cached. The problem was that we don't have a lockfile for this project (as we shouldn't) but an updated version of a dependency was breaking the build. Because of the cache this breakage was not appearing in CI. This made things difficult to debug.

I guess there's not a great solution to this, possibly run `cargo update` before building, or have a test specific lockfile?

jeffa5:
Yeah, I think we may be able to remove the record of the registry cache so it re-checks versions. Another issue may be how to clean out the cache when we do update versions as I think they'll just stay around. May be able to use something like [cargo-cache](https://github.com/matthiaskrgr/cargo-cache) periodically?

jeffa5:
I'll close this as it doesn't seem to be a huge pain point and caching may just add more complexity.

PR #97, opened 2021-04-18
Sync protocol

jeffa5:
This is an implementation of the work in https://github.com/automerge/automerge/tree/sync-protocol-3


jeffa5:
Currently this passes most of the interop tests (after rebasing the sync-protocol-orion branch onto performance) with just 3 failing tests.

jeffa5:
Now down to just 2 failing (both where `null` is expected but the serde conversion gives us `undefined`)

jeffa5:
And adding a custom serializer (kinda hacky but not sure of another way at the moment) brings us to all tests passing!

orionz:
All tests are passing - code looks good!

ept:
@jeffa5 If you want to make the SyncState completely opaque, I think we could change the JS tests to avoid referring to fields within the SyncState. Would that help?

Another thing we were discussing with @orionz is freeing the SyncState memory when it is no longer needed. The wasm side doesn't know when an object goes out of scope in JS and is freed; does this mean we need an explicit function to free wasm memory?

jeffa5:
> @jeffa5 If you want to make the SyncState completely opaque, I think we could change the JS tests to avoid referring to fields within the SyncState. Would that help?

That would be great @ept! I've only got the implementations for getters and setters there for the tests and otherwise I consider them unstable so they can be removed whenever (something we should document if they remain).

> Another thing we were discussing with @orionz is freeing the SyncState memory when it is no longer needed. The wasm side doesn't know when an object goes out of scope in JS and is freed; does this mean we need an explicit function to free wasm memory?

Yes, I suppose we will need a free function for that then, seems a bit painful to have to do manual memory management for something that exists in wasm memory. Do we know of a nicer way? This may be more of a problem with the immutable api since we'll be creating lots of sync states, each needing freeing by the dev.

ept:
Actually, it looks like [most modern browsers](https://caniuse.com/mdn-javascript_builtins_finalizationregistry) now support [FinalizationRegistry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry), which could probably be used to free wasm memory when JS is no longer referencing it. Maybe this is becoming a viable alternative to having to call `free` manually?

jeffa5:
And support seems to be ready in wasm-bindgen https://github.com/rustwasm/wasm-bindgen/blob/44d577f6b89dc7cc572ea0747833d38ba680e93b/guide/src/reference/weak-references.md

PR #101, opened 2021-04-22
add fastpath to get_changes

vedantroy:


automerge-backend/src/backend.rs
```diff
@@ -18,6 +18,7 @@ pub struct Backend {
     actors: ActorMap,
     hashes: HashMap<amp::ChangeHash, Change>,
     history: Vec<amp::ChangeHash>,
+    history_idx: HashMap<amp::ChangeHash, usize>,
```
jeffa5, line 21: I feel like there is some redundancy here, could we instead move to something like:
```rust
history: Vec<Change>,
```
and drop the `hashes` field?

And nitpicking, I prefer using whole words, e.g. index instead of idx but that is very minor.

automerge-backend/src/backend.rs
```diff
@@ -230,6 +233,47 @@ impl Backend {
...
+        let mut missing_changes = vec![];
+        let mut has_seen: HashSet<_> = have_deps.iter().collect();
+        for hash in &self.history[*lowest_idx..] {
+            let change = self.hashes.get(hash).unwrap();
+            let mut deps_seen = 0;
+            for dep in &change.deps {
+                if has_seen.contains(dep) {
+                    deps_seen += 1;
+                }
+            }
```
jeffa5, line 264: Could be an iter -> map -> sum

automerge-backend/src/backend.rs
```diff
@@ -230,6 +233,47 @@ impl Backend {
...
+            if deps_seen > 0 && deps_seen == change.deps.len() {
+                missing_changes.push(change);
+                has_seen.insert(hash);
+            } else if deps_seen > 0 {
+                break;
+            }
```
jeffa5, line 270: Could reduce duplication of the `deps_seen > 0` check here by moving the else if to just an else inside the first block.

jeffa5 (changes requested):
Looking good :)

automerge-backend/src/backend.rs
orionz: I agree!  hashes has been removed

automerge-backend/src/backend.rs
orionz: fixed

automerge-backend/src/backend.rs
orionz: goot catch

jeffa5:
Think this just needs a rebase and quick fixup of some errors there and this should be good to go

orionz:
I had some series conficts with this and had to hand port the code over to another branch

PR #102, opened 2021-04-23
Reduce cloning of JsSyncState in wasm backend

jeffa5:
Sync may be called frequently, cloning the sync state a lot thrashes the
garbage collector in js and is a bit pointless in our wasm version.

The only reason for keeping the clone is that the js side may want to
use the old value. This means they can have the previous and the new
one, however it isn't free and I don't think it likely they will be
interested in a previous sync state anyway.

Thus we now take the sync state by value to prevent having to clone it
and then pass it back again in the return list.

### DIFF
--- a/automerge-backend-wasm/src/lib.rs
+++ b/automerge-backend-wasm/src/lib.rs
@@ -232,9 +232,11 @@ fn export_changes(changes: Vec<&Change>) -> Array {
 }
 
 #[wasm_bindgen(js_name = generateSyncMessage)]
-pub fn generate_sync_message(input: Object, sync_state: &JsSyncState) -> Result<JsValue, JsValue> {
+pub fn generate_sync_message(
+    input: Object,
+    mut sync_state: JsSyncState,
+) -> Result<JsValue, JsValue> {
     get_input(input, |state| {
-        let mut sync_state = sync_state.clone();
         let message = state.0.generate_sync_message(&mut sync_state.0);
         let result = Array::new();
         result.push(&JsValue::from(sync_state));
@@ -251,7 +253,7 @@ pub fn generate_sync_message(input: Object, sync_state: &JsSyncState) -> Result<
 #[wasm_bindgen(js_name = receiveSyncMessage)]
 pub fn receive_sync_message(
     input: Object,
-    sync_state: &JsSyncState,
+    mut sync_state: JsSyncState,
     message: JsValue,
 ) -> Result<JsValue, JsValue> {
     let mut state: State = get_state(&input)?;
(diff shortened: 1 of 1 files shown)


jeffa5:
This has the downside of mutating the input which makes the js api no longer immutable/functional so may not want to introduce this.

orionz:
I like this patch but I breaks a test - we may need to adjust the test.  

@ept - Rust can either mutate the peer state coming in (which would be bad since js is expecting the input to not change - or to burn it so it cannot be resued - this patch (which I like burns it).  This however breaks a test - specifically this line 

```
      ;[r, rSyncState] = [Automerge.clone(n2), s2]
```

Where n2 is cloned by s2 is just renamed and reused a few lines down.  The first use is fine and the second use throws an error.  My idea would be to change this line to

```
      ;[r, rSyncState] = [Automerge.clone(n2), Automerge.clone(s2) ]
```
And overload Automerge.clone to work on backends or  sync state.  What do you think

ept:
Happy to consider this, although the overloaded `clone` function should probably be `Backend.clone`, not `Automerge.clone`.

However, I'm wondering if this is optimising the wrong thing. Maybe we can make it cheap to copy the SyncState by not storing the whole list of hashes we have sent, but rather storing a set of heads indicating the changes that we think the other node already has. I think @jeffa5 suggested something like this. This would make the SyncState so small that there is no need to make it mutable. I think this might be a better solution.

orionz:
> This would make the SyncState so small that there is no need to make it mutable. I think this might be a better solution

That's an interesting idea - I thought that we still had an unbounded number unapplied changes in the peer state that would need to be serialized on each call but looking at the data structure that's gone too!  😄 If this data structure is small enough serializing it on each call could be pretty cheap - what do you think @jeffa5 - if we get a more compact SyncState would you be ok keeping it in the Js memory manager so we don't need to change the api?

jeffa5:
Yep, we no longer keep an unbounded number of `ChangeHash`s in the `sent_hashes` field since this PR https://github.com/automerge/automerge-rs/pull/103/files. @ept we may want to mirror this in the js side but I feel I lack the sufficient js skill.

At the time of making this PR we did have a lot of changes and this fixed the issue, the above solution brings performance closer (though still not to the same level).

I'm happy to keep it the immutable way (what is on main). This helps make it similar to the js API. One query though is how the js side needs to handle the memory, calling free etc. With the immutable method they may have to call it quite a lot for old copies they don't need which can end up thrashing the malloc/free system rather than doing useful work.

jeffa5:
I'll close this as we currently want to keep the immutable aspect of the JS API.

PR #104, opened 2021-04-25
get changes fast

orionz:
Had to hand copy code over - created some tests - broke the fast path into its own function to test

### DIFF
--- a/automerge-backend/src/backend.rs
+++ b/automerge-backend/src/backend.rs
@@ -276,8 +276,8 @@ impl Backend {
             if has_seen.contains(&hash) {
                 continue;
             }
-            if let Some(idx) = self.history_index.get(&hash) {
-                stack.extend(self.history[*idx].deps.iter());
+            if let Some(change) = self.history_index.get(&hash).and_then(|i| self.history.get(*i)) {
+                stack.extend(change.deps.iter());
             }
             has_seen.insert(hash);
         }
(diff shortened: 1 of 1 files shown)


automerge-backend/src/backend.rs
```diff
@@ -232,32 +236,67 @@ impl Backend {
...
         let mut stack: Vec<_> = have_deps.iter().collect();
         let mut has_seen = HashSet::new();
         while let Some(hash) = stack.pop() {
             if has_seen.contains(&hash) {
                 continue;
             }
-            if let Some(change) = self.hashes.get(&hash) {
-                stack.extend(change.deps.iter());
+            if let Some(idx) = self.history_index.get(&hash) {
+                stack.extend(self.history[*idx].deps.iter());
```
jeffa5, line 280: To avoid the direct indexing here the if let could be 
```rust
if let Some(change) = self.history_index.get(&hash).and_then(|i| self.history.get(*i)) {
...
}
```

automerge-backend/src/backend.rs
```diff
@@ -322,9 +363,9 @@ impl Backend {
             }
 
             for dep in self
-                .hashes
+                .history_index
                 .get(&hash)
-                .map(|c| c.deps.as_slice())
+                .map(|i| self.history[*i].deps.as_slice())
```
jeffa5, line 368: Could use `and_then`

jeffa5 (changes requested):
Looking good, nice to have the split be clear.

Just some style changes which are up to you but the test name should be changed.

orionz:
> Just some style changes which are up to you but the test name should be changed.

Good catch

PR #106, opened 2021-04-26
Remove automerge bench harness

jeffa5:
This removes the default wrapper which throws errors with some cli
arguments, e.g. --save-baseline which is nice for being able to compare
versions!


alexjg:
LGTM

PR #112, opened 2021-05-01
Ensure that ChangeBytes::compressed contains the original compressed bytes

### LINKED ISSUE #95: getAllChanges returns incorrect data when DEFLATE is used
I have created a failing test here: https://github.com/livingspec/automerge-upstream/blob/apply-changes-bug/test/test.js#L1315

Using the latest [automerge](https://github.com/automerge/automerge/commit/81079ff75d2234b47cb912bad728158f2e71c527) and [automerge-backend-wasm](https://github.com/automerge/automerge-rs/commit/e607d2e22baa67ab42b0aec15015c0de4d67c53d).

Based on my observations, the first 11 bytes are clipped from the change.

```
AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected ... Lines skipped

  [
+   Uint8Array(323) [
-   Uint8Array(334) [
-     133,
-     111,
-     74,
-     131,
-     252,
-     38,
-     106,
-     255,
-     2,
-     195,
-     2,
      117,
...
      0
    ]
  ]
```

The gist of the change is that i have created a long text object containing:

```
Donec mollis hendrerit risus. Praesent venenatis metus at tortor pulvinar varius. Donec vitae orci sed dolor rutrum auctor. Nullam cursus lacinia erat. Cras ultricies mi eu turpis hendrerit fringilla.
```

I'm happy to have a stab at fixing the issue, but hoping someone quicker than me would :)

alexjg:
fixes #95 

The issue here is that we represent uncompressed and compressed changes using this data type:

```rust
#[derive(PartialEq, Debug, Clone)]
enum ChangeBytes {
    Compressed {
        compressed: Vec<u8>,
        uncompressed: Vec<u8>,
    },
    Uncompressed(Vec<u8>),
}
```

This is so that we can operate on the uncompressed changes but avoid recompressing when we already have the compressed bytes. We were not correctly setting the value of `ChangeBytes::Compressed::compressed` to the original compressed bytes - specifically we were removing the header - this didn't affect any logic operating over the uncompressed changes but meant that the incorrect `compressed` bytes were returned by `Backend::get_changes` et al.

I've fixed the issue and added a regression test. We should probably consider adding a type to represent the original bytes and accessing elements of those original bytes through it rather than directly slicing the compressed buffer. This might avoid problems like this in future.


automerge-backend/src/change.rs
```diff
@@ -508,12 +508,16 @@ fn decode_change(bytes: Vec<u8>) -> Result<Change, AutomergeError> {
     })
 }
 
-fn decompress_chunk(preamble: &[u8], buf: &[u8]) -> Result<ChangeBytes, AutomergeError> {
-    let mut decoder = DeflateDecoder::new(buf);
+fn decompress_chunk(
+    preamble: Range<usize>,
+    body: Range<usize>,
+    buf: &[u8],
```
jeffa5, line 514: Could `buf` instead be a `Vec<u8>` since we convert it with `.into()` at the end of this function?

jeffa5 (commented):
Looks good to me, unfortunately the diff cuts of the final usage of `buf` but extending that makes it clearer where the problem was.

PR #113, opened 2021-05-04
Backend tidy

jeffa5:
Move to storing changes only in the history to reduce duplication and some other style cleanups. Also introduced some more clippy linting and I've disabled some annoying lints but find that it does improve the code further and we can always re-enable them later.

Happy to discuss things / update clippy rules. May be best to look commit-by-commit.

### DIFF
--- a/automerge-backend/src/actor_map.rs
+++ b/automerge-backend/src/actor_map.rs
@@ -15,7 +15,7 @@ impl ActorMap {
     pub fn import_key(&mut self, key: &amp::Key) -> Key {
         match key {
             amp::Key::Map(string) => Key::Map(string.to_string()),
-            amp::Key::Seq(eid) => Key::Seq(self.import_element_id(&eid)),
+            amp::Key::Seq(eid) => Key::Seq(self.import_element_id(eid)),
         }
     }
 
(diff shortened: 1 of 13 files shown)


alexjg:
Lot's of excellent cleanups here :broom: .

I notice that we have a lot of places in the backend where we store an index to something as a kind of reference (e.g actor IDs, element IDs, etc.). I wonder if it would be worth adding a type alias for `usize` in each case  (e.g `type ActorIDIndex = usize`). This wouldn't really give us any additional type safety but might be worth it as a kind of documentation.

jeffa5:
> I notice that we have a lot of places in the backend where we store an index to something as a kind of reference (e.g actor IDs, element IDs, etc.). I wonder if it would be worth adding a type alias for `usize` in each case (e.g `type ActorIDIndex = usize`). This wouldn't really give us any additional type safety but might be worth it as a kind of documentation. This isn't something we should address in this PR I think, just putting the idea out in the world so it doesn't get trapped in my head.

Going further with this idea with lots of indices going around it may make sense to use newtype structs and create wrappers for the collections that only allow those as indices. Thus we'd have type safety for where we use `usize`s in ambiguous places.

jeffa5:
I think the use of aliases or newtype structs for more type safety in the backend is good but belongs in its own PR so I'll merge this as is.

PR #116, opened 2021-05-07
Add event handler logic

jeffa5:
Remove partialeq from Backend since I'm not sure of its relevance or
correctness anyway and complicates things with the event handlers.

### DIFF
--- a/automerge-backend/src/backend.rs
+++ b/automerge-backend/src/backend.rs
@@ -8,9 +8,15 @@ use amp::ChangeHash;
 use automerge_protocol as amp;
 
 use crate::{
-    actor_map::ActorMap, change::encode_document, error::AutomergeError,
-    event_handlers::EventHandlers, internal::ObjectId, op_handle::OpHandle, op_set::OpSet,
-    pending_diff::PendingDiff, Change,
+    actor_map::ActorMap,
+    change::encode_document,
+    error::AutomergeError,
+    event_handlers::{EventHandlerId, EventHandlers},
+    internal::ObjectId,
+    op_handle::OpHandle,
+    op_set::OpSet,
+    pending_diff::PendingDiff,
+    Change, EventHandler,
 };
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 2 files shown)


automerge-backend/src/event_handlers.rs
```diff
@@ -0,0 +1,67 @@
...
+        for handler in &mut self.0 {
+            if let EventHandler::AfterApplyChange(f) = handler {
+                f.0(change)
+            }
+        }
+    }
+
+    /// Remove the handler with the given id.
+    pub fn remove_handler(&mut self, id: EventHandlerId) {
+        self.0.remove(id.0);
```
alexjg, line 39: This can panic if the index is out of bounds. The `EventHandlerId` wrapper makes this less likely to happen but you could somehow get an ID from a different document.

automerge-backend/src/backend.rs
```diff
@@ -1,22 +1,27 @@
...
-#[derive(Debug, PartialEq, Clone)]
+#[derive(Debug, Clone)]
 pub struct Backend {
     queue: Vec<Change>,
     op_set: OpSet,
     states: HashMap<amp::ActorId, Vec<usize>>,
     actors: ActorMap,
     history: Vec<Change>,
     history_index: HashMap<amp::ChangeHash, usize>,
+    pub event_handlers: EventHandlers,
```
alexjg, line 24: Do we want to make this public rather than proxying the add and remove handler methods? Feels like it's coupling us to this particular implementation strategy unnecessarily.

automerge-backend/src/event_handlers.rs
jeffa5: Ah thanks, good catch. I'll fix

automerge-backend/src/backend.rs
jeffa5: Yeah, I think I'll change this to proxy the functions to give us the adaptability.

alexjg (approved):

PR #117, opened 2021-05-07
Allow travis to fail on beta

jeffa5:


jeffa5:
Closing as actually I think these lints should just be fixed.

PR #123, opened 2021-05-12
Add C bindings for sync functions

vedantroy:
For testing, I copied the "test_not_reply_if_we_have_no_data_as_well" test from the JS frontend.

### DIFF
--- a/automerge-c/src/lib.rs
+++ b/automerge-c/src/lib.rs
@@ -511,9 +511,7 @@ pub unsafe extern "C" fn automerge_generate_sync_message(
 #[no_mangle]
 pub extern "C" fn automerge_sync_state_init() -> *mut SyncState {
     let state = SyncState {
-        handle: automerge_backend::SyncState {
-            ..Default::default()
-        },
+        handle: automerge_backend::SyncState::default(),
     };
     state.into()
 }
--- a/automerge-c/src/lib.rs
+++ b/automerge-c/src/lib.rs
@@ -461,6 +461,10 @@ impl From<SyncState> for *mut SyncState {
     }
 }
 
+/// # Safety
+/// Must be called with a valid backend pointer
+/// sync_state must be a valid pointer to a SyncState
+/// `encoded_msg_[ptr|len]` must be the address & length of a byte array
 // Returns an `isize` indicating the length of the patch as a JSON string
 // (-1 if there was an error, 0 if there is no patch)
 #[no_mangle]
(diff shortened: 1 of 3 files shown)


automerge-c/src/lib.rs
```diff
@@ -450,3 +450,78 @@ pub unsafe extern "C" fn automerge_read_binary(backend: *mut Backend, buffer: *m
...
+    } else {
+        0
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn automerge_sync_state_init() -> *mut SyncState {
+    let state = SyncState {
+        handle: automerge_backend::SyncState {
+            ..Default::default()
```
jeffa5, line 515: I think this should use `SyncState::default()` as that encodes some required defaults.

automerge-c/src/lib.rs
```diff
@@ -450,3 +450,78 @@ pub unsafe extern "C" fn automerge_read_binary(backend: *mut Backend, buffer: *m
...
+/// Returns an `isize` indicating the length of the binary message
+/// (-1 if there was an error, 0 if there is no message)
+#[no_mangle]
+pub unsafe extern "C" fn automerge_generate_sync_message(
+    backend: *mut Backend,
+    sync_state: &mut SyncState,
+) -> isize {
+    let msg = (*backend).generate_sync_message(&mut sync_state.handle);
+    if let Some(msg) = msg {
+        (*backend).handle_binary(msg.encode())
```
jeffa5, line 505: `encode` and `decode` on the `SyncState` don't keep all of the data. This was intentional as it is intended for storage rather than this kind of conversion. This will likely need a separate method of passing this to the C code.

jeffa5 (changes requested):
Thanks for adding these! Just some things that need working out with the encode/decode logic. We may also want to add the encode and decode operations for a `SyncState` directly for users wanting to persist it.

automerge-c/src/lib.rs
orionz: Wouldn't the correct thing here be to just pass the opaque pointer back and forth rather than encoding and decoding?

jeffa5: Yes, that sounds better since it should be treated as opaque anyway.

vedantroy: Hmm.. not sure I understand. I am calling `encode` on the `SyncMessage`, not the `SyncState`?

jeffa5: Ah yes, sorry!

orionz:
Im going to merge this since we need it on swift development - we'll do the syncstate functions on a different PR

PR #131, opened 2021-05-16
Set console_error_panic_hook when initing a backend

jeffa5:
This means we can get nicer panic messages in the JS side when using the wasm backend.

This was already in the `Cargo.toml` so I expect it was there at some point in history in the code too but I don't know what happened to it.

Ideally we wouldn't ever panic but especially as this is still under development with things breaking it will at least make debugging wasm tests a bit easier.

We could also add it in more places, e.g. where we create sync states but I think the backend is the main place.


alexjg:
LGTM but I haven't done much on the WASM side

jeffa5:
@orionz If you've got a minute to weigh in on this as I think you've done more on the wasm stuff.

PR #133, opened 2021-05-18
Implement getAddedChanges

vedantroy:
Implemented getAddedChanges which was in the JS backend, but not the Rust backend.

Not sure how to test this b/c many of the JS interop tests are failing right now due to  this (https://github.com/automerge/automerge-rs/commit/87410e4c8fe2a295c0c66e413f0451649193f541) I assume.

### DIFF
--- a/automerge-backend/src/backend.rs
+++ b/automerge-backend/src/backend.rs
@@ -357,22 +357,17 @@ impl Backend {
     pub fn get_changes_added<'a>(&self, other: &'a Self) -> Vec<&'a Change> {
         let mut stack: Vec<_> = other.op_set.deps.iter().collect();
         let mut seen_hashes = HashSet::new();
-        let mut added_changes = Vec::new();
+        let mut added_change_hashes = Vec::new();
         while let Some(hash) = stack.pop() {
-            if !seen_hashes.contains(&hash) && self.get_change_by_hash(&hash).is_some() {
+            if !seen_hashes.contains(&hash) && self.get_change_by_hash(&hash).is_none() {
                 seen_hashes.insert(hash);
-                added_changes.push(hash);
-                let idx = match other.history_index.get(&hash) {
-                    Some(i) => i,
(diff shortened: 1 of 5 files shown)


jeffa5:
On that branch there is only currently one failing test.

vedantroy:
Only 1 test failing now!

automerge-backend/src/backend.rs
```diff
@@ -354,6 +354,25 @@ impl Backend {
             .and_then(|index| self.history.get(*index))
     }
 
+    pub fn get_changes_added<'a>(&self, other: &'a Self) -> Vec<&'a Change> {
+        let mut stack: Vec<_> = other.op_set.deps.iter().collect();
+        let mut seen_hashes = HashSet::new();
+        let mut added_change_hashes = Vec::new();
```
alexjg, line 360: Do we need both `seen_hashes` and  `added_change_hashes`? They look like they are the same so we could just use the `seen_hashes`.

automerge-backend/src/backend.rs
```diff
@@ -354,6 +354,25 @@ impl Backend {
             .and_then(|index| self.history.get(*index))
     }
 
+    pub fn get_changes_added<'a>(&self, other: &'a Self) -> Vec<&'a Change> {
```
alexjg, line 357: Could we add some documentation on what this does? It's not immediately obvious to me from the name.

vedantroy: Copied in the comments from the JS version.

automerge-backend/src/backend.rs
vedantroy: From looking at the comments in the JS version, it seems like we want to preserve the order of the DFS.

vedantroy:
Also added a C binding, although it is untested because there are no tests for this method in the JS version of Automerge.

Will see if I can brainstorm a test.

PR #134, opened 2021-05-18
Use elem id from insertion in frontend

jeffa5:
Element ids in sequences are tied to the operation that created the element. Performing a set shouldn't update the element id as it was previously doing. This was from a test case found by quickcheck in automergeable.


alexjg (approved):

jeffa5:
Whats the preferred practice with these stacked PRs, merge them into the in progress branch now or wait for that to be merged and then merge this directly into main?

alexjg:
I think we should merge them in to the in progress branch immediately.

PR #136, opened 2021-05-18
Add bytes type

vedantroy:
Implements the bytes datatype for Automerge + 3 tests for it (2 frontend, 1 backend). (Do we need more?)

### DIFF
--- a/automerge-backend/src/columnar.rs
+++ b/automerge-backend/src/columnar.rs
@@ -430,10 +430,7 @@ impl<'a> Iterator for ValueIterator<'a> {
             v if v % 16 == VALUE_TYPE_BYTES => {
                 let len = v >> 4;
                 let data = self.val_raw.read_bytes(len).ok()?;
-                Some(amp::ScalarValue::Bytes(
-                    // This unwrap is fine b/c we encode as valid UTF-8 when serializing
-                    String::from_utf8(data.into()).unwrap(),
-                ))
+                Some(amp::ScalarValue::Bytes(data.to_vec()))
             }
             v if v % 16 >= VALUE_TYPE_MIN_UNKNOWN && v % 16 <= VALUE_TYPE_MAX_UNKNOWN => {
                 let len = v >> 4;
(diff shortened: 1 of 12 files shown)


automerge-protocol/src/serde_impls/op.rs
```diff
@@ -155,7 +155,7 @@ impl<'de> Deserialize<'de> for Op {
                         Some(ScalarValue::F32(n)) => Ok(OpType::Inc(n as i64)),
                         Some(ScalarValue::Counter(n)) => Ok(OpType::Inc(n)),
                         Some(ScalarValue::Timestamp(n)) => Ok(OpType::Inc(n)),
-                        Some(ScalarValue::Str(s)) => {
+                        Some(ScalarValue::Str(s) | ScalarValue::Bytes(s)) => {
```
jeffa5, line 158: This is giving a build error.

automerge-backend/src/columnar.rs
```diff
@@ -429,9 +429,11 @@ impl<'a> Iterator for ValueIterator<'a> {
             }
             v if v % 16 == VALUE_TYPE_BYTES => {
                 let len = v >> 4;
-                let _data = self.val_raw.read_bytes(len).ok()?;
-                unimplemented!()
-                //Some((amp::Value::Bytes(data))
+                let data = self.val_raw.read_bytes(len).ok()?;
+                Some(amp::ScalarValue::Bytes(
+                    // This unwrap is fine b/c we encode as valid UTF-8 when serializing
+                    String::from_utf8(data.into()).unwrap(),
```
jeffa5, line 435: This is only safe if we know we encoded it but someone else could have so I think this should return the error.

jeffa5 (changes requested):
I'm not totally aware of the context for adding this as it seems like users could always just base64 encode stuff themselves and use `Str` to store it but I'm still fine with this being added.

automerge-protocol/src/serde_impls/op.rs
vedantroy: Fixed

automerge-backend/src/columnar.rs
vedantroy: Closing b/c new version stores raw bytes (no string).

automerge-protocol/src/lib.rs
```diff
@@ -214,6 +217,13 @@ impl ScalarValue {
                     datatype,
                 }),
             },
+            (DataType::Bytes, ScalarValue::Bytes(bytes)) => Ok(ScalarValue::Bytes(bytes.clone())),
+            (DataType::Bytes, v) => Err(error::InvalidScalarValue {
+                raw_value: self.clone(),
+                expected: "a base64 encoded string of bytes".to_string(),
```
jeffa5, line 223: I think this needs updating since we aren't using b64 anymore

jeffa5 (changes requested):
Looking good, CI is passing now. I think the previous failure was due to match patterns.

Just one bit of b64 left to remove.

automerge-protocol/src/lib.rs
vedantroy: Fixed

PR #141, opened 2021-05-20
Fix backend panics

jeffa5:
This adds some fuzzing to check loading arbitrary bytes. While these shouldn't always generate actual values it shouldn't panic.

The fuzzing found some issues that I've encapsulated in tests and fixed.


alexjg:
Very nice!

Issue #143, opened 2021-05-21
New encoding/decoding breaks downstream libraries.


johnfaucette:
The encoding::Error and decoding::Error are not added to AutomergeError so they cannot be caught by consuming libraries in general.

Here was my quick hack to fix this issue locally in (`automerge-backend/src/lib.rs`)
```rust
pub use encoding::Error as EncodingError;
pub use decoding::Error as DecodingError;
```

I don't know much about the structure of automerge-rs or rust in general, but it might make sense to take these up in the `errors.rs` file. Right now there is only the encoding::Error and decoding::Error from the external crates being considered, we need to be able to respond to the errors being thrown from the the encoding and decoding modules.

jeffa5:
Thanks for the report @johnfaucette. We've applied the fix and should be on main now :)

PR #144, opened 2021-05-21
Make encoding and decoding mods public for errors

jeffa5:
Fixes https://github.com/automerge/automerge-rs/issues/143

This makes the errors public and usable by consumers of the API.

Up to discussing whether we want to keep a flat API or nested?


alexjg:
My preference would be to directly re-expose `encode::Error as EncodingError` and `decoding::Error as DecodingError`. I think we should avoid exposing the internal crate structure in the API as we are very likely to want to change it.

jeffa5:
Thanks @alexjg, updated it to be flat and added some quick comments on them.

alexjg (approved):

PR #147, opened 2021-05-25
Sort predecessors during encoding

### LINKED ISSUE #145: `pred` should be sorted in encoded change
This issue was found by @endeavour42. The following two changes differ only in the order of the `preds` on the operation:

```js
{"actor": "f4d13a8a", "seq": 67, "startOp": 140, "time": 6207000, "message": "",
  "deps": ["3a1ebe64fdb4b442317e518e0a03238cbde72291394217e0954061588ca8c2e5"],
  "hash": "1fe5702c0d516dba1bf04847d4bcf64faedc22d6b00a389cb08ddf1da8d42fca",
  "ops": [
    {"obj": "_root", "key": "commonVar", "action": "set", "insert": false, "value": "4091494582",
     "pred": ["139@f4d13a8a", "139@37ea42f2"]}
  ]
}

{"actor": "f4d13a8a", "seq": 67, "startOp": 140, "time": 6207000, "message": "",
  "deps": ["3a1ebe64fdb4b442317e518e0a03238cbde72291394217e0954061588ca8c2e5"],
  "hash": "2c42aea2581b92b157a83c250e7f7a1cd3d528f651774555a996bb7eb6ed3339",
  "ops": [
    {"obj": "_root", "key": "commonVar", "action": "set", "insert": false, "value": "4091494582",
     "pred": ["139@37ea42f2", "139@f4d13a8a"]}
  ]
}
```

The binary format requires that `preds` appear in sorted order (in ascending Lamport timestamp order). The JS implementation does this sorting [here](https://github.com/automerge/automerge/blob/main/backend/columnar.js#L441), but it seems like the Rust implementation is not doing this sorting step. The correct encoded hash for both changes above should be `2c42aea2581b92b157a83c250e7f7a1cd3d528f651774555a996bb7eb6ed3339`.

jeffa5:
Fixes #145

### DIFF
--- a/automerge-backend/src/columnar.rs
+++ b/automerge-backend/src/columnar.rs
@@ -129,6 +129,12 @@ impl<'a> Iterator for OperationIterator<'a> {
         let obj = self.objs.next()?;
         let key = self.keys.next()?;
         let pred = self.pred.next()?;
+        #[cfg(debug_assertions)]
+        {
+            let mut pred_sorted = pred.clone();
+            pred_sorted.sort();
+            debug_assert_eq!(pred, pred_sorted, "pred should be sorted");
+        }
         let value = self.value.next()?;
         let action = match action {
             Action::Set => InternalOpType::Set(value),
@@ -997,7 +1003,7 @@ struct ColumnOp<'a> {
     action: InternalOpType,
     obj: Cow<'a, amp::ObjectId>,
     key: Cow<'a, amp::Key>,
-    pred: Cow<'a, [amp::OpId]>,
+    pred: Vec<amp::OpId>,
     insert: bool,
 }
 
@@ -1023,7 +1029,7 @@ impl ColumnEncoder {
             obj: o.obj,
             key: o.key,
             action: o.action,
-            pred: o.pred,
+            pred: o.pred.into_owned(),
             insert: o.insert,
         });
         e.encode(colops, actors);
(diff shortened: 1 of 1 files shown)


alexjg (approved):
I'm increasingly wondering whether we should actually be checking all these things up front when deconding the document rather than just when iterating over it. The current architecture leaves us 
no way of signaling to the user that the document is invalid.

That's a concern for the future though, this patch fixes the problem.

jeffa5:
Yeah, this fixes the problem for now.

As for checking all the things in the document before iterating over it, I'm not sure. I've put it as a debug assert as technically we don't rely on things like the predecessors being sorted internally (as far as I know) so it isn't actually an issue. Kind of goes by the be liberal in what you accept thing.

ept:
Thanks @jeffa5!

My personal opinion is that with the Automerge change format, the principle "be liberal in what you accept and strict in what you produce" does not apply. We should be both strict in what we accept and strict in what we produce. The reason is that every change should have exactly one canonical representation, and thus one unambiguous hash. This is important because when we load a compressed document and want to reconstruct the original change history, we need to produce the byte-for-byte identical changes to what they originally were — otherwise we can't recompute the hash graph. If we make the change decoder so strict that it refuses to load any change that does not fit the canonical format, that means any ambiguities in the encoder will be found and fixed much faster.

PR #152, opened 2021-05-27
strict numbers

orionz:
This PR is to match https://github.com/automerge/automerge/pull/371.  The json for scalar values now contains datatype 'int', 'uint', 'float32', 'float64'.


automerge-protocol/src/lib.rs
```diff
@@ -286,6 +294,38 @@ impl ScalarValue {
...
+                    raw_value: self.clone(),
+                    expected: "a uint".to_string(),
+                    unexpected: v.to_string(),
+                    datatype,
+                },
+            )?)),
+            (DataType::F32, v) => Ok(ScalarValue::F32(v.to_f32().ok_or(
+                error::InvalidScalarValue {
+                    raw_value: self.clone(),
+                    expected: "a uint".to_string(),
```
jeffa5, line 316: Should be f32

automerge-protocol/src/lib.rs
```diff
@@ -286,6 +294,38 @@ impl ScalarValue {
...
+                    unexpected: v.to_string(),
+                    datatype,
+                },
+            )?)),
+            (DataType::F64, v) => Ok(ScalarValue::F64(v.to_f64().ok_or(
+                error::InvalidScalarValue {
+                    raw_value: self.clone(),
+                    expected: "a uint".to_string(),
```
jeffa5, line 324: Should be f64

automerge-protocol/src/lib.rs
```diff
@@ -303,10 +343,51 @@ impl ScalarValue {
         }
     }
 
+    pub fn to_u64(&self) -> Option<u64> {
+        match self {
+            ScalarValue::Int(n) => Some(*n as u64),
```
jeffa5, line 348: Shouldn't this be stricter? I would it expect to only return a `Some` for a `Uint`, similarly for the other `to_*` functions.

automerge-protocol/tests/serde_round_trip_proptest.rs
```diff
@@ -117,11 +117,13 @@ prop_compose! {
     }
 }
 
+/*
 enum Mode {
     Json,
     MessagePack,
 }
 
+>>>>>>> strict numbers
```
jeffa5, line 126: Needs a cleanup

jeffa5 (changes requested):

vedantroy:
Fixed the issue with the missing "preds" field.

automerge-protocol/src/serde_impls/op.rs
```diff
@@ -53,6 +59,22 @@ impl Serialize for Op {
...
+            OpType::Set(ScalarValue::F32(value)) => {
+                op.serialize_field("value", &value)?;
+                op.serialize_field("datatype", &DataType::F32)?;
+            }
+            OpType::Set(ScalarValue::F64(value)) => {
+                op.serialize_field("value", &value)?;
+                op.serialize_field("datatype", &DataType::F64)?;
+            }
             OpType::Set(value) => op.serialize_field("value", &value)?,
             OpType::MultiSet(values) => op.serialize_field("values", &values)?,
```
vedantroy, line 79: Will this result in the `datatype` not being specified?

vedantroy:
Looks like we are using an old version of the Rust compiler that thinks the [this syntax](https://github.com/rust-lang/rust/issues/54883) is experimental.

automerge-protocol/src/serde_impls/op.rs
vedantroy: Here's an example:

```json
{
  "ops": [
    {
      "action": "set",
      "obj": "_root",
      "elemId": "_head",
      "values": [
        "hello",
        3
      ],
      "pred": []
    }
  ],
  "actor": "",
  "seq": 0,
  "startOp": 0,
  "time": 0,
  "message": null,
  "deps": []
}
```

vedantroy: I think the correct way to solve this is to add a serializer for `ScalarValue`s.

vedantroy:
@jeffa5 Is this ok to merge? (once I fix CI?)

jeffa5:
@vedantroy There still seems to be a few discussions unresolved. As for CI, I'm not sure about the MSRV for us, we should probably decide on one and document it in the README.

I also think we should probably wait until the corresponding PR is merged in the js repo, then we won't have to rely on orion's fork for the interop tests.

automerge-protocol/src/serde_impls/op.rs
alexjg: How is this intended to work in general? Is the idea that all the `values` would be the same datatype?

automerge-protocol/src/lib.rs
vedantroy: It seems like all the other `to_` functions are pretty lax?

Example:

```rust
    /// If this value can be coerced to an i64, return the i64 value
    pub fn to_i64(&self) -> Option<i64> {
        match self {
            ScalarValue::Int(n) => Some(*n),
            ScalarValue::Uint(n) => Some(*n as i64),
            ScalarValue::F32(n) => Some(*n as i64),
            ScalarValue::F64(n) => Some(*n as i64),
            ScalarValue::Counter(n) => Some(*n),
            ScalarValue::Timestamp(n) => Some(*n),
            _ => None,
        }
    }
```

jeffa5: Ok, we can leave it for now then.

automerge-protocol/src/serde_impls/op.rs
vedantroy: @alexjg I believe so

vedantroy: @alexjg Is there any reason in particular why values in a `MultiSet` should have the same type? Otherwise I can remove some code I added here.

alexjg: Well if they are different types then we'll need to change the serialization logic to add a type tag to each value right? Otherwise how do you indicate that (for example) the first element in a list of numbers is a counter but the second is just an integer?

alexjg: Note that I am using the word `datatype` quite specifically to mean the `datatype` tag that is used for single element inserts

ept: I think MultiSet should focus on the case where all elements have the same datatype, since including a separate datatype tag for every value would add bloat, and the whole point of MultiSet is to be more compact. If the data contains values of different datatypes, they should be split into runs of MultiSet where each run only contains items of the same type (or use a plain Set if the run is of length 1).

jeffa5 (approved):
I think this looks ok.

Up to you whether we wait for the js side to be merged first or not for the interop tests.

vedantroy:
@jeffa5 Can you run CI on this again (not sure why it isn't running?)

jeffa5:
@vedantroy I've had a quick look but seems to be something above my access level, perhaps @orionz or @alexjg can see?

From looking at the requests section I can see

>  Could not authorize build request for automerge/automerge-rs. 

So it may be billing related or due to the move from travis-ci.org to .com which happened recently

automerge-frontend/src/state_tree/resolved_path.rs
```diff
@@ -930,36 +935,116 @@ pub struct ResolvedPrimitiveMut<'a> {
...
     if let Some(v) = op_iter.next() {
-        if let Some(prim) = prim_from_op_action(&v.action) {
-            prim_vals.push(prim);
-            preds.extend(v.pred.clone());
-        }
-        for o in op_iter {
-            if let Some(prim) = prim_from_op_action(&o.action) {
-                prim_vals.push(prim);
-                preds.extend(o.pred.clone());
+        let mut new_ops = vec![];
```
alexjg, line 944: Could use `Vec::with_capacity(ops.len())` here.

automerge-protocol/src/lib.rs
```diff
@@ -227,15 +244,134 @@ impl DataType {
...
+            ScalarValueKind::Timestamp => Some(DataType::Timestamp),
+            ScalarValueKind::Int => Some(DataType::Int),
+            ScalarValueKind::Uint => Some(DataType::Uint),
+            ScalarValueKind::F64 => Some(DataType::F64),
+            _ => None,
+        }
+    }
+}
+
+impl TryFrom<Vec<ScalarValue>> for ScalarValues {
```
alexjg, line 333: Can we move this and the `From` and `fmt::Display` impls below to `utility_impls`

alexjg:
I think this will need to be re-added to the Automerge travis because of the .com -> .org migration, @orionz I think you are the only person with permission to do that.

vedantroy:
@jeffa5 Getting some clippy errors about SmolStr & Arbitrary? Are these expected? Do I need to fix them?

vedantroy:
There are 2 wasm interop tests failing. However, they are also failing on the main branch. So not related to this PR.

vedantroy:
In the wasm interop tests, when checking out the JS repository, I'm getting a weird error "fatal: reference is not a tree: 14e01eebdcc7677a2316df5326ca104823c23183".

vedantroy:
Ok, so there's an error in the wasm interop tests about a `TextDecoder` not being defined, I suspect we need to update our version of node.

ept:
TextDecoder should be available in Node >= 12, which is the oldest currently supported version. I suggest updating to Node 16, which is the [currently active long-term-support version](https://nodejs.org/en/about/releases/).

PR #153, opened 2021-05-27
added binary search to doc decode for big speedup

orionz:
found a place with in document loads that was bottlenecking - sped it up with a binary search.  Also did some code cleanup for easier reading.

### DIFF
--- a/automerge-backend/src/change.rs
+++ b/automerge-backend/src/change.rs
@@ -580,14 +580,14 @@ fn decompress_chunk(
 
 fn group_doc_change_and_doc_ops(
     changes: &mut [DocChange],
-    ops: &mut Vec<DocOp>,
+    mut ops: Vec<DocOp>,
     actors: &[amp::ActorId],
 ) -> Result<(), decoding::Error> {
-    let mut change_actors = HashMap::new();
-    let mut actor_max = HashMap::new();
+    let mut changes_by_actor: HashMap<usize, Vec<usize>> = HashMap::new();
 
     for (i, change) in changes.iter().enumerate() {
-        if change.seq != *actor_max.get(&change.actor).unwrap_or(&1) {
+        let actor_change_index = changes_by_actor.entry(change.actor).or_default();
+        if change.seq != (actor_change_index.len() + 1) as u64 {
             return Err(decoding::Error::ChangeDecompressFailed(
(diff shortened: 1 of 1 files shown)


automerge-backend/src/change.rs
```diff
@@ -634,23 +634,25 @@ fn group_doc_change_and_doc_ops(
...
-            // this is safe - invalid seq would have thrown an error earlier
-            let idx: usize = *change_actors.get(&(op.actor, seq)).unwrap();
-            // this is safe since I build the array above ^^
-            let change = &mut changes[idx];
-            if op.ctr <= change.max_op {
-                change.ops.push(op.clone());
-                continue 'outer;
+    for op in ops {
+        // binary search for our change
+        let actor_change_index = changes_by_actor.entry(op.actor).or_default();
```
jeffa5, line 639: This `actor_change_index` should probably be `actor_change_indices` here and above.

jeffa5 (commented):
Logic looks good, one thing I'm thinking about is if this now relies on changes being encoded in causal order? I'm not sure if that is a current requirement of the binary format or not - haven't really seen docs for it.

automerge-backend/src/change.rs
jeffa5: I know that is true in our implementation but is it always required? I remember seeing another comment, maybe in columnar, that they may not always be in causal order.

automerge-backend/src/change.rs
```diff
@@ -634,23 +634,25 @@ fn group_doc_change_and_doc_ops(
...
+            if changes[actor_change_index[seq]].max_op < op.ctr {
+                left = seq + 1
+            } else {
+                right = seq
             }
         }
-        return Err(decoding::Error::ChangeDecompressFailed(
-            "Doc MaxOp Invalid".into(),
-        ));
+        if left >= actor_change_index.len() {
```
jeffa5, line 650: Instead of Len should this check the last value in the vec?

automerge-backend/src/change.rs
orionz: Ill double check with martin but both implementations follow the same logic on doc save and use this same binary search.

jeffa5: Ok, just something to make sure gets documented then.

PR #154, opened 2021-05-27
Serialize/Deserialize OpType so it works with msgpack

vedantroy:
There an issue serializing a `Change` into msgpack b/c of some oddities in how rmp-serde implements enum serialization/deserialization.

This PR fixes that issue.

Note: There is a TODO for error handling b/c I couldn't get the types to work out, I'll continue working on it.


vedantroy:
Was running into this issue with a different enum. It is now fixed.

alexjg:
I just added the code to report the missing variant error to serde. Happy to merge

vedantroy:
Thanks! I'll add the missing code to the other enum, and then feel free to merge 😄 .

automerge-protocol/src/serde_impls/diff.rs
```diff
@@ -69,8 +70,7 @@ impl<'de> Deserialize<'de> for RawDiffType {
             "text" => RawDiffType::Text,
             "list" => RawDiffType::List,
             "table" => RawDiffType::Table,
-            // TODO: Not sure how to idiomatically return an error
-            _ => panic!("unexpected variant"),
+            other => return Err(Error::unknown_variant(other, VARIANTS)),
```
alexjg, line 73: Rather than using return it's probably neater to just return the result of the `match`, like

```rust
match raw_type.as_str() {
    "text" => Ok(RawDiffType::Text),
    ...
    other => Err(Error::unknown_variant(other, VARIANTS))
}
```

vedantroy: Can change -- I used the return b/c I didn't want to deal with Adding `Ok` to each arm

vedantroy:
@alexjg  What about this? I'm still using `return`, but it's a bit shorter.

alexjg:
My personal preference is to avoid `return` as much as possible as I find it's easier to read code if I imagine everything as an expression with a result. That's my functional programming scars showing though, if you find this version easier to read then go with that.

vedantroy:
Changed & merged!

PR #155, opened 2021-05-27
C bindings msgpack

vedantroy:
This updates the C bindings to use msgpack.


vedantroy:
2 of the C tests that used for loops to iterate over return values have been commented out b/c I didn't feel like writing a for loop over message pack arrays in C, but I can re-add them if necessary (maybe through the use of some Rust helper functions?)

PR #157, opened 2021-06-04
fix deflate column decode order

orionz:
Simple fix to remove deflate bit from column decode

### DIFF
--- a/automerge-backend-wasm/Cargo.toml
+++ b/automerge-backend-wasm/Cargo.toml
@@ -35,11 +35,15 @@ serde_bytes = "0.11.5"
 version = "^0.2"
 features = ["serde-serialize"]
 
+[package.metadata.wasm-pack.profile.release]
+wasm-opt = false
+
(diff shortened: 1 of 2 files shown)


automerge-backend-wasm/Cargo.toml
```diff
@@ -35,12 +35,16 @@ serde_bytes = "0.11.5"
 version = "^0.2"
 features = ["serde-serialize"]
 
+[package.metadata.wasm-pack.profile.release]
+wasm-opt = false
+
```
jeffa5, line 40: Is this related?

jeffa5 (commented):
Looks good, just the one seemingly unrelated bit.

automerge-backend-wasm/Cargo.toml
orionz: Good catch...  I needed that to get the tests to pass on my new M1 MacBook but it should not be in the PR!

PR #161, opened 2021-06-08
Make path.elements() method public

jeffa5:
This makes it so that users can manipulate values based on paths.

### DIFF
--- a/automerge-frontend/src/path.rs
+++ b/automerge-frontend/src/path.rs
@@ -1,7 +1,7 @@
 use std::fmt;
 
 #[derive(Debug, Clone, PartialEq)]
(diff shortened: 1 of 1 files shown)


alexjg:
What's the use case for this? I'm very unsure about the Path API at the moment so it would  be interesting to see what you need to do with it.

jeffa5:
I was trying out something like being able to traverse a rust struct given the path, it got a bit difficult without being able to break down the path. It seems like we should have an open description of the path since it is used in quite a few public places and we may want to be able to use it in other libraries. I'm not saying that the current representation will always be there but at least for now being able to get the internal representation is useful.

PR #165, opened 2021-06-10
Mutation tracker rollback

jeffa5:
This implements rollback on a `MutableDocument`. This is useful to handle when the user applies a change to the document that causes an error (e.g. missing index in insert). This mainly avoids us having to clone the current state each time and lets us make the hot path quicker and cheaper.

Thankfully when we overwrite a value in the document (map or seq), including deletes, we get the old value back so we can get away with not having to clone things and this ends up just delaying the drop call until we have finished with all of our changes.

tldr: cloning a large document for each change is expensive, tracking the individual changes for rollback is much cheaper

### DIFF
--- a/automerge-frontend/src/mutation.rs
+++ b/automerge-frontend/src/mutation.rs
@@ -4,7 +4,8 @@ use unicode_segmentation::UnicodeSegmentation;
 use crate::{
     error::InvalidChangeRequest,
     state_tree::{
-        LocalOperationResult, ResolvedPath, ResolvedPathMut, SetOrInsertPayload, StateTree,
+        LocalOperationResult, MultiGrapheme, MultiValue, ResolvedPath, ResolvedPathMut,
+        SetOrInsertPayload, StateTree,
     },
     value::{Cursor, Primitive, Value},
     Path, PathElement,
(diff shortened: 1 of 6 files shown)


jeffa5:
This should be pointed at https://github.com/automerge/automerge-rs/pull/156 but that branch doesn't live on this repo :/

automerge-frontend/src/mutation.rs
```diff
@@ -82,6 +83,36 @@ impl LocalChange {
...
+
+    fn into_grapheme(self) -> MultiGrapheme {
+        match self {
+            Self::Value(_) => unreachable!(),
+            Self::Grapheme(grapheme) => grapheme,
+        }
+    }
+}
+
+enum LocalOperationForRollback {
```
alexjg, line 108: We could avoid the `MultiThing` wrapper by making the operations specific to the data type right? i.e `Set` for everything which needs a `MultiValue` and `SetText` for anything which needs a `MultiGrapheme`?

alexjg (commented):
This generally looks great. Do we have any tests covering it?

automerge-frontend/src/mutation.rs
```diff
@@ -82,6 +83,36 @@ impl LocalChange {
...
+enum MultiThing {
+    Value(MultiValue),
+    Grapheme(MultiGrapheme),
+}
+
+impl MultiThing {
+    fn into_value(self) -> MultiValue {
+        match self {
+            Self::Value(value) => value,
+            Self::Grapheme(_) => unreachable!(),
```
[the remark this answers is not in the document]
jeffa5: I've now removed `MultiThing`

automerge-frontend/src/mutation.rs
jeffa5: Ah yes, that's a much better idea! (I've also added a `SetList` as it expects the old value to always exist unlike in a map so avoids an unwrap).

alexjg (approved):

PR #168, opened 2021-06-14
Remove MapType from Map and add Table variant

### LINKED ISSUE #158: Change Value::Map to just be a map and add Value::Table
Currently the frontend's `Value` enum has one `Map` variant which represents both maps and tables. While these are both maptypes we don't do this for both sequence types (sequence and text). I also find it sometimes awkward to always have to add the `MapType::Map` when I'm only working with those.

Adding a new variant and updating the existing map one shouldn't be too invasive while making things a bit clearer too.

Thoughts?

jeffa5:
This simplifies the `Value` enum and hides the protocol internals from the
variants.

Fixes #158

### DIFF
--- a/automerge-frontend/src/lib.rs
+++ b/automerge-frontend/src/lib.rs
@@ -1,6 +1,4 @@
-use automerge_protocol::{
-    ActorId, ChangeHash, MapType, ObjectId, Op, OpId, Patch, UncompressedChange,
-};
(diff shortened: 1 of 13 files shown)


alexjg (approved):

alexjg:
We should probably look at doing this in the protocol types as well. The two level map and sequence types are really an artifact of serialization and don't IMO make a good model of the domain.

PR #170, opened 2021-06-15
New value zero copy multivalue

jeffa5:
This helps to squeeze out some more performance on map heavy documents.

### DIFF
--- a/automerge-frontend/src/state_tree/mod.rs
+++ b/automerge-frontend/src/state_tree/mod.rs
@@ -1032,7 +1032,7 @@ impl Cursors {
         })
     }
 
-    fn union(&mut self, other: Cursors) {
+    fn extend(&mut self, other: Cursors) {
         for (k, v) in other.0 {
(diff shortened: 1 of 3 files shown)


automerge-frontend/src/state_tree/mod.rs
```diff
@@ -1019,24 +1019,27 @@ struct CursorState {
...
             cursor.referred_object_id.clone() => vec![cursor],
         })
     }
 
-    fn union(&self, other: Cursors) -> Cursors {
-        Cursors(self.0.clone().union_with(other.0, |mut c1, c2| {
-            c1.extend(c2);
-            c1
-        }))
+    fn union(&mut self, other: Cursors) {
```
alexjg, line 1035: Given that this is no longer an immutable operation should we rename it to something like `extend`?

jeffa5: Yeah that makes sense, done.

alexjg (approved):

PR #175, opened 2021-06-17
Flatten objtype

jeffa5:
Fixes https://github.com/automerge/automerge-rs/pull/168#issuecomment-860846059

### DIFF
--- a/automerge-protocol/src/lib.rs
+++ b/automerge-protocol/src/lib.rs
@@ -46,35 +46,43 @@ impl ActorId {
 #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Copy, Hash)]
 #[serde(rename_all = "camelCase", untagged)]
 pub enum ObjType {
-    Map(MapType),
-    Sequence(SequenceType),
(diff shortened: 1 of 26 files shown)


automerge-protocol/src/lib.rs
```diff
@@ -400,7 +408,7 @@ pub enum Diff {
 pub struct MapDiff {
     pub object_id: ObjectId,
     #[serde(rename = "type")]
-    pub obj_type: MapType,
+    pub map_type: MapType,
```
alexjg, line 411: Rather than having these `map_type` and `seq_type` tags, would it be worth also splitting the diffs up. E.g `MapDiff`, `TableDiff`, `ListDiff`, `TextDiff`?

jeffa5: Ah, I was thinking about doing this but stopped short. I'll follow through with this now.

jeffa5: Pushed

alexjg (approved):
Veery nice

PR #176, opened 2021-06-17
Add SortedVec struct to ensure preds are sorted

jeffa5:


automerge-protocol/src/lib.rs
```diff
@@ -330,12 +330,60 @@ pub enum OpType {
...
+    fn from_iter<I>(iter: I) -> Self
+    where
+        I: std::iter::IntoIterator<Item = T>,
+    {
+        let mut inner: Vec<T> = iter.into_iter().collect();
+        inner.sort_unstable();
+        Self(inner)
+    }
+}
+
```
alexjg, line 380: I think we should probably add an `IntoIterator` implementation

jeffa5: Done

alexjg (approved):

PR #178, opened 2021-06-17
Change ActorId to be exact size array

jeffa5:


jeffa5:
This has a surprising performance improvement but I'm not sure on whether it is specifically conforming to the automerge spec, I know that ActorIds should be hex-like but I can't see a defined length. They do document it as a UUID so it should be this specified length. I'll probably need to update the js side to be stricter here too.

ept:
I would like to avoid hard-coding the size of actorIds because different applications may want to define them differently. For example, some may want to define them to be elliptic-curve public keys in order to be able to have actors sign their changes, and those keys are typically 32 bytes in size.

jeffa5:
Hmm, interesting. Do you think there will be some standardisation of accepted sizes e.g. 16, 32 and 64 or will it just remain an arbitrary sequence of bytes?

ept:
I think it would be safest to keep it as an arbitrary-length byte sequence, since it's difficult to foresee what people might want to do in the future. As a better approach, how about having a lookup table that you use to translate between actor IDs (arbitrary-length byte arrays) and actor numbers (small, fixed-size integers)? The mapping between IDs and numbers would be specific to a particular instance of the backend, so numbers need to be translated into IDs for any communication with the outside world, but internally within the backend a small integer is sufficient to identify a particular actor.

jeffa5:
Yeah, we do have that translation in the actor map, this just simplified bits around the protocol but the reasoning for keeping it variable makes sense.

PR #179, opened 2021-06-18
Use tinyvec for actor id

jeffa5:
Most of the time users will not be using their own identifiers and so
use random uuids. These can fit nicely on the stack for a speedup.
TinyVec allows us to capture this mostly stack, sometimes heap
behaviour.

### DIFF
--- a/automerge-protocol/Cargo.toml
+++ b/automerge-protocol/Cargo.toml
@@ -16,6 +16,7 @@ serde = { version = "^1.0", features=["derive"] }
 strum = { version = "0.21.0", features=["derive"]}
 arbitrary = { version = "1", features = ["derive"], optional = true }
 smol_str = { version = "0.1.18", features = ["serde"] }
+tinyvec = { version = "1.2.0", features = ["alloc"] }
 
 [dev-dependencies]
 maplit = "^1.0.2"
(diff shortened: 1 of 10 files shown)


automerge-protocol/Cargo.toml
```diff
@@ -14,6 +14,7 @@ uuid = { version = "^0.8.2", features=["v4"] }
 thiserror = "1.0.16"
 serde = { version = "^1.0", features=["derive"] }
 arbitrary = { version = "1", features = ["derive"], optional = true }
+tinyvec = { version = "1.2.0", features = ["alloc"] }
```
alexjg, line 17: Have you checked that this builds okay in a `no_std` environment? It looks like it should but I'm not certain.

jeffa5: I don't think automerge builds in a `no_std` environment for a while (at least running `cargo nono check --package automerge` gives lots of issues). Is there a concrete use case for use in `no_std` at the moment? I can tell it might be nice but at this point I think it will be quite a lot of work to get there.

orionz:
Have you measured the impact on performance?

jeffa5:
@orionz I'm seeing about a 7-11% improvement on the B1 benchmarks compared to main

automerge-protocol/Cargo.toml
alexjg: Fair. Let's not worry about that now then.

alexjg:
I'm happy to merge this once conflicts are fixed. I'll wait until the conflicts are fixed to mark it as approved though or my approval will get dismissed by the updates.

jeffa5:
I'll wait on https://github.com/Lokathor/tinyvec/pull/146 to work on this as we should still retain the arbitrary functionality.

jeffa5:
My TinyVec PR got merged so I've updated this PR to use the git version.

alexjg (approved):

PR #185, opened 2021-06-21
Remove debug info for release builds

jeffa5:
This should help fix the size of compiled library.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,3 @@ members = [
     "fuzz",
     "perf",
 ]
-
-[profile.release]
-debug = true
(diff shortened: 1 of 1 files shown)


endeavour42:
@jeffa5 it helps although the release library size (checked with mac and ios libs) is still about 4x (40MB vs 10MB) compared to versions prior to a6e955abfdd02a96ab9feb127c66b6192b46d38b of 13 June. cc @orionz

jeffa5:
@endeavour42 I've re-enabled lto which may be having the rest of the impact. Otherwise I'm not sure what is impacting it so much.

endeavour42:
@jeffa5 yes that helped a lot, the old and new sizes: ios old: 15.6 MB new: 18.7 MB, macOS old: 10.1MB new: 11.3 MB
(old is pre 13 june version, new is the current version)

PR #186, opened 2021-06-24
Reduce deflate compression on the hot path

jeffa5:
Currently all changes get compressed when converting from a protocol change to a backend change.

This compression is actually quite costly time-wise and some use cases might not want to always compress changes, rather they can tolerate the slightly larger size. (For instance I am happy to save the changes in larger format and just save an entire document when space gets too much).

This PR makes the change to have compression as an opt-in. Users just call `compress` on a change they have and then the `raw_bytes` call will return the compressed version.

I've added compression to changes to be synced as this is a use case I can imagine it may be more acceptable to run compression on. Also note that this is likely to be a lot fewer changes getting compressed.

Some other question points:
- [ ] Should the wasm interface call compress by default to maintain the previous behaviour?
- [ ] Should the C library do the same? 
- [ ] Rather than mutating the `ChangeBytes` would it be better to have it do the compression in the function call and return that data directly?

### DIFF
--- a/automerge-backend-wasm/src/lib.rs
+++ b/automerge-backend-wasm/src/lib.rs
@@ -129,7 +129,8 @@ pub fn apply_local_change(input: Object, change: JsValue) -> Result<JsValue, JsV
         let change: amp::Change = change
             .into_serde()
             .map_err(|_| AutomergeError::DecodeFailed)?;
-        let (patch, change) = state.0.apply_local_change(change)?;
+        let (patch, mut change) = state.0.apply_local_change(change)?;
+        change.compress();
         let result = Array::new();
         let change_bytes = types::BinaryChange(change.raw_bytes().to_vec());
         // FIXME unwrap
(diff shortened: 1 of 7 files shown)


alexjg:
I think we should expose an interface to the C and WASM APIs to compress changes, much like the `Change::compress` function you introduce here. I'm happy for that to be in another PR but @orionz and @vedantroy might have different opinions.

alexjg (approved):

alexjg (approved):

jeffa5:
I've added calls to compress automatically in the wasm and c libs. This mimics the old behaviour (hopefully) so that tests still pass and there aren't too many surprises until we expose an actual `compress` api.

jeffa5:
I'm not totally convinced about compressing each change individually in the sync message, I think it may be better to let users compress the entire message however they want depending on network characteristics but I can leave it there for now and see if it gets to be a large proportion.

alexjg:
Are we good to merge this?

jeffa5:
I think so.

PR #189, opened 2021-06-26
C API  MessagePack improvements

vedantroy:


orionz:
Some in progress improvements to the MessagePack C API

orionz:
Known issues:  Should be able to skip datatype for int/uint/float.  Hash should be binary, not array of uints.  Does binary work properly.  .h header files are seriously out of date.

orionz:
Looks like we're going a different way on the c-api.  It's going to get a major overhaul.  Good work otherwise tho.

PR #198, opened 2021-06-28
Add cards example.

dvc94ch:


alexjg:
I like this example and I think that at some point we should probably put something like the `Automerge` struct in the library proper. 

It's probably worth removing some or all of the comments which (I think) are left over from the JS version of this example. They're confusing as the API is a bit different.

automerge/Cargo.toml
```diff
@@ -34,6 +34,7 @@ tracing-subscriber = {version = "0.2", features = ["chrono", "env-filter", "fmt"
 unicode-segmentation = "1.7.1"
 maplit = "^1.0.2"
 smol_str = "0.1.17"
+anyhow = "1.0.41"
```
alexjg, line 37: This should be an example only dependency.

jeffa5: This is already in the `dev-dependencies` section, I'm not aware of an examples only one?

alexjg: Ah, right you are, I assumed you could have dependencies scoped by target. `dev-dependencies` seems fine.

alexjg:
Last nit, can you add a `Default` implementation for the `Automerge` struct to satisfy clippy?

PR #201, opened 2021-06-30
Add send feature flag

jeffa5:
Some situations require use of a non-`Send` event handler, this makes the entire backend non-`Send` so it is behind a feature flag.


jeffa5:
Cargo features should be additive so this doesn't really work. I'll look into another way of doing this, likely having both a send and non-send backend available.

PR #202, opened 2021-07-01
Generic actor

dvc94ch:


jeffa5:
Thanks! CI isn't working at the moment, would you be able to run `make` locally and fix up the errors?

jeffa5:
Thanks again!

PR #204, opened 2021-07-01
Return an error to the user when trying to delete a non-existent key

jeffa5:


alexjg:
Do we have no tests covering this code path?

jeffa5:
I don't think so, I can look into adding some.

alexjg (approved):

PR #205, opened 2021-07-03
Add sendable backend

jeffa5:
This introduces both a `Send`able and non-`Send`able backend by sharing a
common `GenericBackend`. This currently just changes the event handler
to require `Send` or not. This is preferred to a mutually exclusive
feature as features should be additive and so both can exist at the same
time (for instance two crates in a workspace with one needing `Send` and
the other not).

### DIFF
--- a/automerge-backend/src/backend.rs
+++ b/automerge-backend/src/backend.rs
@@ -2,6 +2,7 @@ use core::cmp::max;
 use std::{
     collections::{HashMap, HashSet, VecDeque},
     fmt::Debug,
+    ops::{Deref, DerefMut},
 };
 
 use amp::ChangeHash;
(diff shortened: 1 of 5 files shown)


jeffa5:
Hmm, this still seems far from ideal. The complexity is just coming from the event handlers which in turn come from the fact that we can't intercept changes applied from the sync process but I think we can.

To capture the changes applied from the sync we should be able to get the current heads before the `receive_sync_message` call and then ask for the changes since those heads.

This currently is flawed as the backend keeps a queue of changes to be applied rather than applying them or returning an error and letting the user manage the queuing. Due to this the functionality for detecting when a change has been applied means we'd have to get the heads of the document before any function call and then get changes since, just in case it made a queued change applicable.

As per https://github.com/automerge/automerge/issues/400 I think we're moving in the direction of at least having queuing be optional so I might take a look into that first, and then the event handlers should be able to be removed.

automerge-backend/src/backend.rs
```diff
@@ -11,26 +12,118 @@ use crate::{
...
+            states: HashMap::default(),
+            actors: ActorMap::default(),
+            history: Vec::default(),
+            history_index: HashMap::default(),
+            event_handlers: EventHandlers::default(),
+        }
+    }
+}
+
+impl<H: EventHandler> Clone for GenericBackend<H> {
```
alexjg, line 111: Why do we need a hand rolled `Clone`?

jeffa5: It was because otherwise the `EventHandlers` clone doesn't clone but instead made an empty vec, I don't think this PR is needed if we remove queueing in the backend (or at least have the option to).

PR #206, opened 2021-07-04
Remove queue from backend

jeffa5:
This also adds a check for any missing dependencies when applying the
changes.

Currently the JS tests are failing so we could either implement the queue for them in the WASM lib or wait for the tests to be updated in the JS repo, thoughts?


jeffa5:
Closing due to API changes

PR #208, opened 2021-07-04
load: Reduce intermediate collects and use iterators more

jeffa5:
This results in a performance boost due to doing less allocation but also should be more memory efficient.


alexjg (approved):

jeffa5:
Just rebased due to a conflict.

PR #209, opened 2021-07-04
load: Don't generate a patch that will just be dropped

jeffa5:


automerge-backend/src/backend.rs
```diff
@@ -96,6 +96,16 @@ impl Backend {
         self.make_patch(diffs, actor)
     }
 
+    fn apply_without_patch(&mut self, changes: Vec<Change>) -> Result<(), AutomergeError> {
```
alexjg, line 99: I think it might be worth adding a little docstring here describing why you might do this.

jeffa5: Yep, that makes sense. Added!

alexjg (approved):

PR #212, opened 2021-07-07
Value proxy

jeffa5:
This represents a read-only proxy for values in the frontend's state. This avoids having to rebuild collections unnecessarily and avoids cloning primitives, letting the user decide.

The proxies are split up by composite type (and primitives). I have tried to implement some baseline useful functions for each, roughly taken from the standard library counterparts. These all use the default (winning) value from the MultiValue (or MultiGrapheme) so don't handle conflicts. I think conflicts are rather specialised and might be too tricky to work into this API.

In the future we may be able to adapt this to also have mutable functionality and record changes but I'm not too sure yet and think a read-only version is still useful!


alexjg:
This is great, should make a lot of things more efficient. My only quibble is the naming. `Proxy` doesn't seem quite right as it's not really a proxy for anything, would `*Ref` work better? E.g `ValueRef` or `TableRef`?

jeffa5:
I imagined it being a proxy for the state tree but am happy to change it to `*Ref` if that's still preferred?

alexjg:
Let's go for `*Ref` then

alexjg (approved):

PR #215, opened 2021-07-08
Binary docs

jeffa5:
This starts an mdbook with the format of the binary representation (the document one). There will need to be other chapters explaining things in automerge but the binary format is something I want to get documented due to the stabilization.


jeffa5:
@alexjg That sounds good to me too.

okdistribute:
Hey thanks for this! We are starting a docs site, thinking to use Docasaurus because we can expand it to include step-by-step tutorials with rich, long-form table of contents on the right-hand sidebar, as well as scale to a visually-designed front page someday. It will all be markdown so if we don't like Docasaurus we can also switch to a different static site generator sometime.

If you're ok with it, I'd like to move this markdown specification of the binary docs to that repository.

heckj:
:wave: I'm completely external to this, but I'd like to reference the details included in this PR for some documentation work I'm doing within automerge-swift, if possible (WIP PR for that work:https://github.com/automerge/automerge-swift/pull/43). Not really time critical or anything, but if there's any general documentation effort happening, I'd like to sync up with whomever is doing it and leverage it while I'm writing things for the swift-specific API interfaces.

okdistribute:
Awesome @heckj! Your PR looks great. We should be able to embed the swift repos docs into the automerge docs

jeffa5:
Closing due to the other main docs effort

PR #218, opened 2021-07-13
Rename Sequence to List

### LINKED ISSUE #214: Rename Value::Sequence to Value::List
It tends to be called list internally and in the js automerge side more I think so may be better to expose it as such.

jeffa5:
Fixes #214


alexjg (approved):
Whoop! It's all naming.

PR #224, opened 2021-07-19
Sync improvements

jeffa5:
This replaces the `get_changes_slow` function with a vector clock using method.

`get_changes_slow` had to traverse the entire graph which was not pretty, also it is rather hard to cache evaluations due to needing to store all the hashes that formed the closure.

The vector clock mechanism allows us to have a cache as well as shortcut some of the traversal.

The hash graph is good for some things but I think keeping some vector clocks internally still helps things performance wise, perhaps we could compactly store them as we go along, removing the need to calculate them on demand?


alexjg:
I like this but I worry that we're putting quite a lot of incidental code into `Backend`. Could we introduce a `VectorClock` type and move some of the work of creating and merging clocks into `impl VectorClock`. I would especially like to move the BFS traversal somewhere else if possible.

jeffa5:
Thanks for taking a look @alexjg. I agree that it should be split up a bit, I'll see how much I can factor the BFS out.

automerge-backend/src/backend.rs
```diff
@@ -26,6 +30,8 @@ pub struct Backend {
     actors: ActorMap,
     history: Vec<Change>,
     history_index: HashMap<amp::ChangeHash, usize>,
+    /// A cache of vector clocks for speeding up sync operations.
+    clocks_cache: Arc<Mutex<HashMap<amp::ChangeHash, VectorClock>>>,
```
jeffa5, line 34: @alexjg Currently this map doesn't shrink, I'm not sure of a good way to decide when to do that besides a fixed size cache, but then what size do we use? Any thoughts?

PR #227, opened 2021-07-22
Optimistic state tree

### LINKED ISSUE #226: Frontend: avoiding cloning the optimistic state when applying remote patches
Currently we avoid cloning the frontend's state when only local operations occur which is a big win, however we still have the issue of cloning the reconciled state when we encounter a non-local patch:

https://github.com/automerge/automerge-rs/blob/4ce27ac18b52a33690b090f459937629374cda8c/automerge-frontend/src/lib.rs#L127-L129

This is due to the optimistic state not having updates from remote peers (fine) but not having a way to get back to the newly reconciled state. I can think of two ways to avoid this:

1. Track the updates that get applied to the remote state and have a mechanism to undo them (similar to the mutation tracker's rollback). After undoing these changes we can then apply the `checked_diff` to the state to get us to the same reconciled state.

2. Alternatively, probably a lot more work, we could alter the way that the frontend works by keeping one reconciled state and not an optimistic state. Instead, to operate optimistically we keep the changes that have been applied optimistically and when querying the document these can be used to figure out the correct operation to perform, e.g. an optimistic delete of a key can be checked for on the map object before trying to obtain it from the reconciled state. There would be quite a bit of complexity here but it would lead to a reduced memory footprint (one state instead of two) and avoids the cloning problem.

I'm currently leaning towards 1 for now as it is quicker to implement and get working while 2 represents a much larger change which should probably be done as a longer term thing.

Anyway, happy to discuss alternatives and comments :)

jeffa5:
This change allows us to store only one copy of the `StateTree` while avoiding having to `clone` it for safety. This is doable by adding rollbacks and an optimistic state wrapper.

Fixes #226

### DIFF
--- a/automerge-frontend/src/mutation.rs
+++ b/automerge-frontend/src/mutation.rs
@@ -8,6 +8,7 @@ use crate::{
         SetOrInsertPayload, StateTree,
     },
     value::{Cursor, Primitive, Value},
+    value_ref::RootRef,
     Path, PathElement,
 };
 
@@ -83,6 +84,7 @@ impl LocalChange {
     }
 }
 
+#[derive(Clone, Debug)]
 enum LocalOperationForRollback {
     Set { old: Option<MultiValue> },
     SetList { old: MultiValue },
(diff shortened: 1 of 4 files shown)


automerge-frontend/src/state_tree/optimistic.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    /// Rollback all applied operations.
+    pub fn rollback_all(&mut self) {
+        let rollback_ops = std::mem::take(&mut self.copies_for_rollback);
+        self.rollback(rollback_ops.into_iter())
+    }
+
+    /// Undo the operations applied to this document.
+    ///
+    /// This is used in the case of an error to undo the already applied changes.
```
alexjg, line 56: This docstring is slightly inaccurate now right? As we also use this in the case of non-local patches?

automerge-frontend/src/mutation.rs
```diff
@@ -83,17 +84,6 @@ impl LocalChange {
...
-    Delete { old: MultiValue },
-    DeleteText { old: MultiGrapheme },
-    Insert,
-    InsertMany { count: usize },
-    Increment { by: i64 },
-}
-
 /// `MutationTracker` is used as the context in which a mutation closure is
 /// applied. The mutation tracker implements `MutableDocument`, which is how it
 /// captures the changes that the mutation closure is making.
```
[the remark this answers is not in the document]
jeffa5: I've added a bit about the use of `OptimisticStateTree`

automerge-frontend/src/state_tree/optimistic.rs
jeffa5: Yep, I've removed this line now.

alexjg (approved):

PR #230, opened 2021-07-28
Remove default from ActorId

jeffa5:


alexjg:
I find this a little worrying, mainly with regard to this sentence in the docs for `Default`:

> Default values are often some kind of initial value, identity value, or anything else that may make sense as a default.

This implies (to me) that `default()` should satisfy `default() == default()`, which `random()` does not. 

However, I'm aware that this is more of an objection to the fuzziness of the definition of `Default` than to this implementation (I'm sure there are people who do not have the intuition that `default() == default()`), is there some ergonomic reason to add this?

jeffa5:
I can see why it would make sense to have `default() == default()` in most cases but I just noticed this as quite a footgun for users. Most users may not be wanting to use anything other than the standard ActorId  (uuids at the moment) so would reach for `ActorId::default()`, rather than giving them a sane default this gives them an empty actor id which will cause problems when sharing documents and with reuse of actor ids.

alexjg:
My argument would be that we can't implement `Default` for `ActorId` at all. I agree that we definitely shouldn't have some kind of "empty" `ActorId` or anything for the reasons you give.

As I said though, this is more of a personal opinion than anything else, are there good reasons we might want a `Default` implementation (rather than just using `ActorId::random` directly)?

jeffa5:
The empty value was just needed for the `mem::take` in the `diffable_sequence` so I've removed the `Default` implementation on `ActorId` and `OpId` and just implemented `Default` for `MultiValue` and `MultiGrapheme` directly (needed as the diffable sequence is generic over them so we can't use a `mem::replace`)

alexjg:
This is better, it still upsets me to have a `Default` implementation for `MultiValue` as it's a way incorrect values could sneak in. Could we instead have a constant `UNSAFE_DUMMY: MultiValue` and then do `mem::replace(&target, UNSAFE_DUMMY.clone())` or similar?

alexjg:
Or, it might make more sense to use an `Option<MultiValue>` and some `unwrap`s where we expect it not to be `None`, that way at least the logic error is obvious.

jeffa5:
As the value is generic I've added a `take` function to the `DiffableValue` trait, this avoids having the default implementation while still keeping it quite simple.

alexjg:
Nice, thanks for indulging me 🙂

alexjg (approved):

jeffa5:
Just missed the `mem::replace` but that should work now.

alexjg (approved):

PR #231, opened 2021-07-29
Make apply_local_change return ref to change

jeffa5:
The `_mut` variant is used so that people can `compress` after.

### DIFF
--- a/automerge-backend/src/backend.rs
+++ b/automerge-backend/src/backend.rs
@@ -122,7 +122,7 @@ impl Backend {
     pub fn apply_local_change(
         &mut self,
         mut change: amp::Change,
-    ) -> Result<(amp::Patch, Change), AutomergeError> {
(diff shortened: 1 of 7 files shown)


automerge-backend/src/backend.rs
```diff
@@ -135,9 +135,25 @@ impl Backend {
...
 
-        Ok((patch, bin_change))
+        let patch: amp::Patch = self.apply(vec![bin_change], Some(actor_seq))?;
+
+        let change = self.get_change_by_hash(&hash).unwrap();
+
+        Ok((patch, change))
+    }
+
+    pub fn apply_local_change_mut(
```
alexjg, line 147: Can we document that you would typically use this if you want to compress the change?

jeffa5: Yep, done.

alexjg (approved):

PR #232, opened 2021-07-29
Backend patch generation

jeffa5:
Various updates to patch generation, trying to make it more performant and also more type safe.


jeffa5:
Closing as this isn't really needed for the new API

PR #233, opened 2021-07-29
Nonzero seq

jeffa5:
Sequence numbers start at 1 so should be nonzero at least when in changes. This also translates to opids, object ids, element ids etc.

Similarly, it doesn't seem like it should be legal to have no operations in a change (though I'm not entirely sure on that so if someone thinks otherwise do let me know).


jeffa5:
Seems like operations can be empty...

jeffa5:
Closing due to new API rewrite.

PR #234, opened 2021-07-29
Frontend sorted map

jeffa5:
An attempt to be able to generically specify that some paths should be sorted maps.

This alters the internal representation. When (if) we get sets in the future this would also extend to those.

This implementation tries to introduce quite a generic schema functionality that we may be able to leverage for things like cambria integration or a more strict frontend.


alexjg:
@jeffa5 I'm inclined to close this now given that it's a large change and we're moving to quite a different API for the Rust crate. Do you think that makes sense?

jeffa5:
Yeah, I think this can be closed for now. I might have a go at trying something similar in the new API at some point.

PR #236, opened 2021-09-03
Sequence tree

jeffa5:
Taking some inspiration from [joseph's blog article](https://josephg.com/blog/crdts-go-brrr/) as well as seeing very poor performance in the flamegraphs from `im_rc::Vector` I decided to work on a simple btree-like implementation for sequences.

This doesn't do anything fancy like RLE encoding sequences of opids but does give us a very nice speed up and changes the bottleneck.

On my machine the main branch (`3c786b26`) runs the automerge trace (as implemented in the `perf` binary, roughly) in about 45 seconds, while this branch (at time of writing) runs in about 2 seconds.

### DIFF
--- a/automerge-frontend/Cargo.toml
+++ b/automerge-frontend/Cargo.toml
@@ -10,13 +10,11 @@ bench = false
 
 [dependencies]
 automerge-protocol = { path = "../automerge-protocol" }
-futures = "0.3.4"
 serde = { version = "^1.0", features=["derive"] }
 serde_json = "^1.0"
 uuid = { version = "^0.8.2", features=["v4"] }
 maplit = "1.0.2"
 thiserror = "1.0.16"
-im-rc = "15.0.0"
 unicode-segmentation = "1.7.1"
 arbitrary = { version = "1", features = ["derive"], optional = true }
 smol_str = "0.1.18"
--- a/automerge-frontend/Cargo.toml
+++ b/automerge-frontend/Cargo.toml
@@ -33,6 +33,7 @@ env_logger = "0.8.3"
 log = "0.4.14"
 wasm-bindgen-test = "0.3.22"
 pretty_assertions = "0.7.1"
+proptest = "0.10.1"
 
 [[bench]]
 name = "statetree_apply_diff"
--- a/automerge-frontend/Cargo.toml
+++ b/automerge-frontend/Cargo.toml
@@ -31,7 +31,7 @@ env_logger = "0.8.3"
 log = "0.4.14"
 wasm-bindgen-test = "0.3.22"
 pretty_assertions = "0.7.1"
-proptest = "0.10.1"
+proptest = { version = "0.10.1", default-features = false, features = ["std"] }
 
 [[bench]]
 name = "statetree_apply_diff"
--- a/automerge-frontend/src/state_tree/sequence_tree.rs
+++ b/automerge-frontend/src/state_tree/sequence_tree.rs
@@ -2,7 +2,7 @@ use std::fmt::Debug;
 
 use automerge_protocol::OpId;
 
-const T: usize = 5;
+const T: usize = 3;
 
 #[derive(Clone, Debug, PartialEq)]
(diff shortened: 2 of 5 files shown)


josephg:
Very cool. And sequencetree is a great name for this structure.

It might be a terrible time to mention this but I'd be happy to pull my tree out into its own crate for reuse. (I've been thinking about doing that for awhile anyway). Its fully tested now and I'm quite happy with its performance, though it still uses a bit more unsafe than I'd like.

automerge-frontend/Cargo.toml
```diff
@@ -10,13 +10,11 @@ bench = false
 
 [dependencies]
 automerge-protocol = { path = "../automerge-protocol" }
-futures = "0.3.4"
 serde = { version = "^1.0", features=["derive"] }
 serde_json = "^1.0"
 uuid = { version = "^0.8.2", features=["v4"] }
 maplit = "1.0.2"
 thiserror = "1.0.16"
-im-rc = "15.0.0"
```
alexjg, line 19: :tada:

automerge-frontend/src/state_tree/sequence_tree.rs
```diff
@@ -0,0 +1,672 @@
...
+                } else {
+                    self.root_node = Some(root.children.remove(0));
+                }
+            }
+
+            #[cfg(debug_assertions)]
+            debug_assert_eq!(len, self.root_node.as_ref().map_or(0, |r| r.check()) + 1);
+            old.1
+        } else {
+            panic!("remove from empty tree")
```
alexjg, line 117: Could we return a `Result` of some kind rather than panicking? If that's not possible we should at least document the cases when this can panic I think.

automerge-frontend/src/state_tree/sequence_tree.rs
```diff
@@ -0,0 +1,672 @@
...
+                            if cumulative_len + child.len() >= index {
+                                child.insert_into_non_full_node(
+                                    index - cumulative_len,
+                                    opid,
+                                    element,
+                                );
+                                self.length += 1;
+                                break;
+                            } else {
+                                cumulative_len += child.len() + 1;
```
alexjg, line 173: This `child.len() + 1` fragment pops up a lot. I feel like the code would be a little easier to read and less error prone if instead of `SequenceTreeNode::children` being a `Vec<SequenceTreeNode<..>>` it was instead a newtype wrapper with some convenience methods for this commonly used logic. Thoughts?

alexjg (commented):
This is generally very nice, I've left a few small comments.

alexjg:
@josephg I would be interested in a separate crate implementing this stuff, did you end up spinning your stuff out?

jeffa5:
> Very cool. And sequencetree is a great name for this structure.
> 
> It might be a terrible time to mention this but I'd be happy to pull my tree out into its own crate for reuse. (I've been thinking about doing that for awhile anyway). Its fully tested now and I'm quite happy with its performance, though it still uses a bit more unsafe than I'd like.

I think that would be great! More reusable data structures like this make it all easier in the long run and can always be updated given some level of compatibility :)

automerge-frontend/src/state_tree/sequence_tree.rs
jeffa5: I've tried to follow a similar API to `Vec`, which doesn't return an option but panics if you access an element outside of the range. Since this is an internal datastructure (for now) I think it is ok to panic, will document it though.

automerge-frontend/src/state_tree/sequence_tree.rs
jeffa5: I've had a little refactor and tried having a `cumulative_iterator` function but the borrow-checker gets in the way as it borrows the entire struct rather than just the field. Something I'll leave for another time I think.

josephg:
> @josephg I would be interested in a separate crate implementing this stuff, did you end up spinning your stuff out?

Uploaded it just now [here](https://crates.io/crates/content-tree)

PR #237, opened 2021-09-14
Add initial github actions setup

jeffa5:


automerge-cli/tests/integration.rs
```diff
@@ -2,45 +2,45 @@ use std::env;
...
-fn import_stdin() {
-    let bin = env!("CARGO_BIN_EXE_automerge");
-    let initial_state_json = serde_json::json!({
-        "birds": {
-            "wrens": 3.0,
-            "sparrows": 15.0
-        }
-    });
-    let json_bytes = serde_json::to_string_pretty(&initial_state_json).unwrap();
+// #[test]
```
jeffa5, line 5: These were causing the new CI to take a very long time. I haven't diagnosed the issue but personally I'm not sure it should block moving away from Travis at the moment.

alexjg: Hmm, it's a shame to lose these tests. Were they taking a long time but succeeding?

jeffa5: I never saw them complete, despite being about 20mins. Also gives some weird output, e.g. here: https://github.com/automerge/automerge-rs/runs/3603850245?check_suite_focus=true

PR #238, opened 2021-10-08
Combined frontend and backend interface

jeffa5:
This provides a simpler interface to automerge when one doesn't
particularly need to separate things across multiple threads.

Also useful for newcomers to automerge to get going quickly and to look inside for more detail.


jeffa5:
I'm aware that due to some ongoing work this may become irrelevant but would like to use this myself so thought it might be useful at least for a bit.

automerge/Cargo.toml
```diff
@@ -18,6 +18,7 @@ bench = false
 serde = { version = "^1.0", features=["derive"] }
 serde_json = "^1.0"
 uuid = { version = "^0.8.2", features=["v4"] }
+thiserror = "1.0.16"
```
alexjg, line 21: Can we just use `"^1.0"` here?

automerge/src/combined.rs
```diff
@@ -0,0 +1,247 @@
...
+        change_closure: F,
+    ) -> Result<(O, Option<&Change>), AutomergeError>
+    where
+        E: std::error::Error + 'static,
+        F: FnOnce(&mut dyn MutableDocument) -> Result<O, E>,
+    {
+        let (out, change) = self
+            .frontend
+            .change(message, change_closure)
+            .map_err(|e| AutomergeError::UserError(Box::new(e)))?;
```
alexjg, line 71: Do we need to return errors here and in the following `apply_patch`. The change has been generated by the frontend we control and likewise the patch is generated by the backend we control and delivered in order. I think there is no way this can fail which isn't a programming error on our part, in which case we should probably just panic?

jeffa5: Agreed, thanks for the simplification. I've removed the variant on the `AutomergeError` enum too now.

PR #239, opened 2021-10-13
Add a `compact` command to the CLI

alexjg:
`compact` loads documents either from a list of path, or from standard
input, compacts the document, and emits it to standard output or a given
path.


pvh:
When I run merge, even if it won't merge it creates a zero length file.

Also, my files are now corrupted (this is probably an automerge-rs bug that I haven't found yet), though might be a bug in the new JS backend.

PR #241, opened 2021-10-14
Correctly sort actor IDs when encoding changes

### LINKED ISSUE #240: automerge doc created by JS won't read in RS
Have not looked into why yet, but this file works in automerge preview.4 (old backend) but automerge-rs won't read it with this error message:

```
$ automerge export .\other.mrg
Error: Decoding error Calculated heads differed from actual heads

Caused by:
    Calculated heads differed from actual heads
```

alexjg:
The javascript implementation of automerge sorts actor IDs
lexicographically when encoding changes. We were sorting actor IDs in
the order the appear in the change we're encoding. This meant that the
index that we assigned to operations in the encoded change was different
to that which the javascript implementation assigns, resulting in
mismatched head errors as the hashes we created did not match the
javascript implementation.

This change fixes the issue by introducing a `HexActoId` type which we
use to sort actor IDs lexicographically. We make a pass over the
operations in the change before encoding to collect the actor IDs and
sort them. This means we no longer need to pass a mutable `Vec<ActorId>`
to the various encode functions, which cleans things up a little.

Fixes #240

### DIFF
--- a/automerge-backend/src/change.rs
+++ b/automerge-backend/src/change.rs
@@ -129,8 +129,9 @@ fn encode_chunk(change: &amp::Change, deps: &[amp::ChangeHash]) -> ChunkIntermed
     }
 
     // encode first actor
-    let mut actors = vec![change.actor_id.clone()];
     change.actor_id.to_bytes().encode(&mut bytes).unwrap();
+
+    let actors = actor_ids_in_change(change);
 
     // encode seq, start_op, time, message
     change.seq.encode(&mut bytes).unwrap();
@@ -144,7 +145,7 @@ fn encode_chunk(change: &amp::Change, deps: &[amp::ChangeHash]) -> ChunkIntermed
         ExpandedOpIterator::new(&change.operations, change.start_op, change.actor_id.clone());
 
     // encode ops into a side buffer - collect all other actors
-    let (ops_buf, mut ops) = ColumnEncoder::encode_ops(expanded_ops, &mut actors);
+    let (ops_buf, mut ops) = ColumnEncoder::encode_ops(expanded_ops, &actors);
 
     // encode all other actors
     actors[1..].encode(&mut bytes).unwrap();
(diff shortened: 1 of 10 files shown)


automerge-backend/src/change.rs
```diff
@@ -171,6 +173,54 @@ fn encode_chunk(change: &amp::Change, deps: &[amp::ChangeHash]) -> ChunkIntermed
...
+        .operations
+        .iter()
+        .flat_map(opids_in_operation)
+        .filter(|a| *a != &change.actor_id)
+        .map(|a| a.into())
+        .unique()
+        .collect();
+    other_ids.sort();
+    // Now prepend the change actor and convert all the HexActorIds back into ActorIds
+    Some(&change.actor_id)
```
jeffa5, line 196: Could this use https://doc.rust-lang.org/std/iter/fn.once.html instead of the option?

automerge-backend/src/change.rs
```diff
@@ -171,6 +173,54 @@ fn encode_chunk(change: &amp::Change, deps: &[amp::ChangeHash]) -> ChunkIntermed
...
+fn actor_ids_in_change(change: &amp::Change) -> Vec<amp::ActorId> {
+    // We construct a vector of HexActorIds to make use of it's Ord implementation which compares
+    // actor IDs lexicographically
+    let mut other_ids: Vec<HexActorId<'_>> = change
+        .operations
+        .iter()
+        .flat_map(opids_in_operation)
+        .filter(|a| *a != &change.actor_id)
+        .map(|a| a.into())
+        .unique()
```
jeffa5, line 192: Would it be good to do the unique call before the map so that we don't unnecessarily do some conversions?

automerge-protocol/src/lib.rs
```diff
@@ -54,6 +55,20 @@ impl ActorId {
...
+        self.0 == other.0
+    }
+}
+
+impl<'a> PartialOrd for HexActorId<'a> {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        self.1.partial_cmp(&other.1)
+    }
+}
+
```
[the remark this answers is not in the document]
alexjg: Good shout, removed.

alexjg:
Ah nice, I confess I built the `HexActorId` thing mostly to not have to try and figure out whether the two representations gave equivalent orderings :sweat_smile: . I've removed the `HexActorId` and left a comment on the `ActorId` struct to warn future maintainers to be careful if changing the way `ActorId` is implemented.

PR #242, opened 2021-11-12
Fix logic for condensing insert ops

alexjg:
The logic for condensing multiple inserts was faulty in that it did not
take into account the `insert` flag on the ops that were being
condensed. This led to an issue where multiple set operations on an
object were converted into a multi-insert operation in the patch. This
commit adds a fix and a few tests.


automerge-frontend/tests/test_frontend.rs
```diff
@@ -907,3 +907,68 @@ fn test_inserts_at_end_of_lists() {
...
     assert_eq!(value, expected_value);
 }
+
+// The logic for condensing multiple inserts was faulty in that it did not take
+// into account the `insert` flag on the ops that were being condensed. This led
+// to an issue where multiple set operations on an object were converted into a
+// multi-insert operation in the patch. This test reproduces that issue for the
+// avoidance of regressions.
+#[test]
+fn reproduce_condense_op_isse() {
```
jeffa5, line 917:
```suggestion
fn reproduce_condense_op_issue() {
```

automerge-frontend/src/state_tree/resolved_path.rs
```diff
@@ -981,10 +981,11 @@ fn condense_insert_ops(ops: Vec<amp::Op>) -> Vec<amp::Op> {
                     cur_multiset_start = None;
                     new_ops.push(v);
                 }
-                (Some(typ), Some(scalar)) => match typ == discriminant(&scalar) {
-                    // there is a multiset in progress & the current op could be part of it
-                    true => cur_prim_vals.push(scalar),
-                    false => {
+                (Some(typ), Some(scalar)) => {
+                    if typ == discriminant(&scalar) && v.insert {
+                        // there is a multiset in progress & the current op could be part of it
```
jeffa5, line 986: Could this comment do with updating?

automerge-frontend/src/state_tree/resolved_path.rs
```diff
@@ -981,10 +981,11 @@ fn condense_insert_ops(ops: Vec<amp::Op>) -> Vec<amp::Op> {
...
+                        cur_prim_vals.push(scalar)
+                    } else {
```
jeffa5, line 988: Should there be a case for the types being the same and it actually being a set, not an insert?

jeffa5 (changes requested):

automerge-frontend/src/state_tree/resolved_path.rs
alexjg: I _think_ the logic is the same right? If the current discriminant is scalarvalue but the op is not `insert` then we can't do a multi insert, otherwise we can't. In the latter case it doesn't actually matter whether we're in a multi set or not. Unless you mean we could avoid doing `finish_multiset` in the case where we're not in a multiset?

automerge-frontend/tests/test_frontend.rs
alexjg: Fixed.

automerge-frontend/src/state_tree/resolved_path.rs
jeffa5: I was thinking that it might be different for a run of sets but I think this is fine.

jeffa5 (approved):

PR #246, opened 2021-12-18
automerge::set - don't generate ops for noops

alexjg:
Repeatedly setting the same value for a particular (obj, key)
combination now no longer generates an operation. To allow this we
modify the return value of `automerge::set` so that it may return an
`Option<OpId>` instead of an `OpId`.

### DIFF
--- a/automerge/src/lib.rs
+++ b/automerge/src/lib.rs
@@ -286,12 +286,25 @@ impl Automerge {
     // inc(obj, prop, value)
     // insert(obj, index, value)
 
+    /// Set the value of property `P` to value `V` in object `obj`.
+    ///
+    /// # Returns
+    ///
+    /// The opid of the operation which was created, or `None` if this operation doesn't change the
+    /// document.
+    ///
+    /// # Errors
+    ///
+    /// This will return an error if
+    /// - The object does not exist
+    /// - The key is the wrong type for the object
+    /// - The key does not exist in the object
     pub fn set<P: Into<Prop>, V: Into<Value>>(
         &mut self,
         obj: OpId,
         prop: P,
         value: V,
-    ) -> Result<OpId, AutomergeError> {
+    ) -> Result<Option<OpId>, AutomergeError> {
         let value = value.into();
         self.local_op(obj.into(), prop.into(), value.into())
     }
@@ -333,11 +346,22 @@ impl Automerge {
         prop: P,
         value: i64,
     ) -> Result<OpId, AutomergeError> {
-        self.local_op(obj.into(), prop.into(), OpType::Inc(value))
+        match self.local_op(obj.into(), prop.into(), OpType::Inc(value))? {
+            Some(opid) => Ok(opid),
+            None => {
+                panic!("increment should always create a new op")
+            }
+        }
     }
 
     pub fn del<P: Into<Prop>>(&mut self, obj: OpId, prop: P) -> Result<OpId, AutomergeError> {
-        self.local_op(obj.into(), prop.into(), OpType::Del)
+        // TODO: Should we also no-op multiple delete operations?
+        match self.local_op(obj.into(), prop.into(), OpType::Del)? {
+            Some(opid) => Ok(opid),
+            None => {
+                panic!("delete should always create a new op")
+            }
+        }
     }
 
     pub fn splice(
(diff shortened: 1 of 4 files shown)


automerge/src/lib.rs
```diff
@@ -281,12 +281,25 @@ impl Automerge {
     // inc(obj, prop, value)
     // insert(obj, index, value)
 
+    /// Set the value of property `P` to value `V` in object `obj`. 
+    ///
+    /// # Returns
+    ///
+    /// The opid of the operation which was created, or None if this operation doesn't change the
```
jeffa5, line 288:
```suggestion
    /// The opid of the operation which was created, or `None` if this operation doesn't change the
```

automerge/src/lib.rs
```diff
@@ -281,12 +281,25 @@ impl Automerge {
...
+    /// document
```
jeffa5, line 289:
```suggestion
    /// document.
```

automerge/src/lib.rs
```diff
@@ -327,11 +340,18 @@ impl Automerge {
...
+            None => { panic!("increment should always create a new op") }
+        }
     }
 
     pub fn del<P: Into<Prop>>(&mut self, obj: OpId, prop: P) -> Result<OpId, AutomergeError> {
-        self.local_op(obj.into(), prop.into(), OpType::Del)
+        // TODO: Should we also no-op multiple delete operations?
+        match self.local_op(obj.into(), prop.into(), OpType::Del)? {
+            Some(opid) => Ok(opid),
+            None => { panic!("increment should always create a new op") }
```
jeffa5, line 353: I think this error message is wrong.

automerge/src/lib.rs
```diff
@@ -327,11 +340,18 @@ impl Automerge {
...
-        self.local_op(obj.into(), prop.into(), OpType::Inc(value))
+        match self.local_op(obj.into(), prop.into(), OpType::Inc(value))? {
+            Some(opid) => Ok(opid),
+            None => { panic!("increment should always create a new op") }
+        }
     }
 
     pub fn del<P: Into<Prop>>(&mut self, obj: OpId, prop: P) -> Result<OpId, AutomergeError> {
-        self.local_op(obj.into(), prop.into(), OpType::Del)
+        // TODO: Should we also no-op multiple delete operations?
```
jeffa5, line 350: I think it makes sense to noop a delete if there isn't anything there to delete.

automerge/Cargo.toml
```diff
@@ -31,4 +31,5 @@ pretty_assertions = "^0.7.2"
 proptest = { version = "0.10.1", default-features = false, features = ["std"] }
 serde_json = { version = "^1.0.73", features=["float_roundtrip"], default-features=true }
 maplit = { version = "^1.0" }
+test-case = "1.2.1"
```
jeffa5, line 34: Is this used?

jeffa5 (changes requested):
I think this makes sense as an optimisation. I guess it could lead to lots of double unwraps but I don't think it makes sense having 'no change' be an error either so I'm happy enough.

alexjg:
I suppose an alternative to returning an `Option<OpId>` would be instead to return the the opid of the existing element.  This would remove the unwraps problem. It would mean that if the caller wanted to detect if a change was made they would need to check the before and after opids - but this may be a rarer usecase and so not worth optimising for.

jeffa5:
Yeah, I think that could be quite surprising to get an opid from the past as a result of calling the function so I'm happy to keep the `Option<>` around.

automerge/src/lib.rs
alexjg: You think we could sometime have a noop increment?

alexjg: Oh, I see what you mean :flushed:

automerge/Cargo.toml
alexjg: removed.

automerge/src/lib.rs
alexjg: Fixed.

automerge/src/lib.rs
alexjg: I'll wait for Orion to weigh in as well just in case we've missed some subtleties of the successor handling logic.

orionz:
I resolved the conflicts and merged this manually.  I thought about doing this at one point but couldn't think of an efficient way to do it - but you realized that the query already had all the info we needed.  Nice.  Seems like we can do the same for `del` and `inc` ... if someone del's something thats not there - or if we try to inc something thats not a counter.

automerge/src/lib.rs
orionz: Yes.  Lets have del() that deletes nothing be a noop.  Also an increment with no matching counter.

orionz: I think right now `inc()` will pred all visible ops - but I think it would make more sense to only pred counters.  This means that if we have conflicts with `( counter a, counter b, string c)` the increment wouldn't clear the conflict (which I like).  If we did an inc(1) we'd get `( counter a + 1, counter b + 1, string c)`  All we would need to do is make sure we only add counters to the pred and return a `None` if we find no counters.

alexjg: Is that the semantics we want? Seems like setting a key should always clear a conflict no?

PR #248, opened 2021-12-19
Add criterion benchmarks for Rust

jeffa5:


edit-trace/benches/main.rs
```diff
@@ -0,0 +1,63 @@
+use automerge::{Automerge, Value, ROOT};
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
+use std::fs;
+
+fn replay_trace(commands: &[(usize, usize, Vec<Value>)]) -> Automerge {
+    let mut doc = Automerge::new();
+
+    let text = doc.set(ROOT, "text", Value::text()).unwrap();
+    for (pos, del, vals) in commands {
+        doc.splice(text, *pos, *del, vals.clone()).unwrap();
```
alexjg, line 10: Does the `vals.clone()` contribute at all to the benchmark? I'm wondering if it's worth cloning the whole `commands` in setup so that we don't need to allocate on each splice call?

jeffa5: Ah, yes that probably might have a small effect. I've moved it to a setup phase now.

alexjg (approved):

PR #249, opened 2021-12-19
Add Automerge::visualise for debugging the optree

alexjg:
The OpTree is a reasonably complicated data structure. This change adds
a method on the OpTree which is then exposed through
`Automerge::visualise` which prints the structure of the OpTree in
graphviz format. Each node in the tree is printed as a simple table of
operations with outgoing edges to tables representing the children of
the node. This is behind a compile time flag (optree-visualisation)
because it's probably not useful except when debugging.

Example output:
optree


alexjg:
Updated to work with the `OpSet` representation which has multiple trees. example output now looks like:

graphviz

orionz:
I want to try this ... How do I generate one of these visualizations?

alexjg:
Right, it's a bit irritating. I typically use it when I'm testing, so I'll do something like 

```rust
doc = automerge::Automerge::new();
...
println!(doc.visualise_optree());
```

Then I'll run the tests with `cargo test --features optree-visualisation`

Then I'll copy the output from the terminal and paste it into a text file to render with `dot`. Alternatively you can paste it into https://dreampuf.github.io/GraphvizOnline

alexjg:
I want a terminal where I can just directly print pictures. But we're not there yet.

PR #250, opened 2021-12-20
Add some rust tests

alexjg:
This commit adds a bunch of testing infrastructure as well as most of the tests from the `legacy_tests.js` file in the js codebase that seem applicable to this codebase.

There are a few small changes in `automerge/src/lib.rs` which should be checked for correctness. The majority of the work here though is in adding some macros to make it maintain these tests going forward.

### DIFF
--- a/automerge/Cargo.toml
+++ b/automerge/Cargo.toml
@@ -31,9 +31,7 @@ version = "^0.3.55"
 features = ["console"]
 
 [dev-dependencies]
-pretty_assertions = "^0.7.2"
+pretty_assertions = "1.0.0"
 proptest = { version = "0.10.1", default-features = false, features = ["std"] }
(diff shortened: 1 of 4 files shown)


alexjg:
It appears that some of the optree visualisation stuff was not quite working in the previous PR, this PR also fixes that. This is entirely accidental as I was maintaining several long running branches and it looks like changes got mixed up between them.

orionz:
Looks good - i see that one panic about `del()` and `inc()` needing to make an op.  Since we've come this far we might as well finish that.  A del that deletes nothing should be a NOOP.  And an increment that hits no counter (pred) should be a NOOP.  Can you squeeze those in this PR?  Or I can do it.

alexjg:
I reckon it's best to do those in another PR. I can do that but not until tomorrow evening as I'm travelling most of tomorrow.

PR #251, opened 2021-12-20
Ci ci oh how i love ci

alexjg:
This adds a bunch of CI steps. There were a few changes I had to make to the code to get clippy and rustfmt to stop complaining. I've also added a deny.toml and run `cargo deny *` to make sure we're not duplicating dependencies or pulling in insecure dependencies.

The biggest thing to review is that I've added an MIT license to all the crates. I think we all (me, @orionz, @jeffa5 ) need to agree to that here. This PR is my endorsement of the license.

### DIFF
--- a/.github/workflows/advisory-cron.yaml
+++ b/.github/workflows/advisory-cron.yaml
@@ -1,0 +1,17 @@
+name: ci
+on:
+  schedule:
+    - cron: '0 18 * * *'
+jobs:
+  cargo-deny:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        checks:
+          - advisories
+          - bans licenses sources
+    steps:
+      - uses: actions/checkout@v2
+      - uses: EmbarkStudios/cargo-deny-action@v1
+        with:
+          command: check ${{ matrix.checks }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,97 @@
+name: ci
+on: [push, pull_request]
+jobs:
+  fmt:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          components: rustfmt
+      - uses: Swatinem/rust-cache@v1
+      - run: ./scripts/ci/fmt
+        shell: bash
+
+  lint:
+    runs-on: ubuntu-latest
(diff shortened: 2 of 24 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,97 @@
...
+      - uses: Swatinem/rust-cache@v1
+      - run: ./scripts/ci/docs
+        shell: bash
+
+  cargo-deny:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        checks:
+          - advisories
```
jeffa5, line 47: Would the advisories be something that could be useful to run periodically, even if we don't add code, since things may get discovered?

automerge/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 name = "automerge"
 version = "0.1.0"
 edition = "2018"
+license = "MIT"
```
jeffa5, line 5: Should we also add the `LICENSE` file to this repo?

jeffa5 (approved):
I'm happy with the MIT license as it fits with Automerge repo's one

.github/workflows/ci.yaml
alexjg: I did not realize you could run actions on a schedule, that's an excellent idea, I'll set it up to run daily.

automerge/Cargo.toml
alexjg: Yep, will do

.github/workflows/ci.yaml
alexjg: Done.

.github/workflows/advisory-cron.yaml
```diff
@@ -0,0 +1,18 @@
...
+    - cron: '0 18 * * *'
+jobs:
+  cargo-deny:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        checks:
+          - advisories
+          - bans licenses sources
+    continue-on-error: ${{ matrix.checks == 'advisories' }}
```
jeffa5, line 13: This does more than just advisories, which is fine but I think it should fail if there are any issues so we can be alerted more easily.

jeffa5 (changes requested):

.github/workflows/advisory-cron.yaml
alexjg: Good shout, removed the `continue-on-error`.

jeffa5 (approved):

PR #252, opened 2021-12-21
`query_at(heads)`

orionz:
I got all the `query_at(heads)` implemented.  Wrote a bunch of tests and they all pass.  All of these do sequential scans so they aren't super fast but they get the job done.  Also removed a bunch of dead code.

### DIFF
--- a/automerge/src/query.rs
+++ b/automerge/src/query.rs
@@ -1,5 +1,5 @@
 use crate::op_tree::{OpSetMetadata, OpTreeNode};
-use crate::{ElemId, Op, OpId, OpType, ScalarValue};
+use crate::{Clock, ElemId, Op, OpId, OpType, ScalarValue};
 use fxhash::FxBuildHasher;
 use std::cmp::Ordering;
 use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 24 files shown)


automerge/src/query.rs
```diff
@@ -185,6 +185,108 @@ impl Default for Index {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Default)]
+pub(crate) struct VisWindow {
+    counters: HashMap<OpId, CounterData>,
+}
+
+impl VisWindow {
+    fn visible(&mut self, op: &Op, pos: usize) -> bool {
```
alexjg, line 194: Can we make the clock for `visible_at` optio Al and then just pass `None` instead of nearly duplicating the implementation?

alexjg (commented):
I really like the `VisWindow` thing, makes it much easier to read. I think the tests could be made more readable once the testing macros I built are merged with a slight modification to the macro to allow specifying the heads to test the document at. I'll have a shot at that tomorrow though.

automerge/src/query.rs
orionz: I could - I guess I was trying to not add any extra "if" statements to the fast path.  But now that I think about it one of the goals for this codebase I have is not adding any optimizations to the code that add complexity if we can't show a measurable difference in the benchmarks.  I suspect that extra "if" wont move the needle at all.

orionz: Well I tried and combining the two functions into one cost me about 20ms on the benchmark (375ms -> 395ms).  Alas...

alexjg: Ah well, at least we know it's worth it.

PR #254, opened 2021-12-28
Expose a non-document-specific opid

alexjg:
# Errrrgonomics

The current API contains a footgun: OpIds which are generated by different documents cannot be used interchangeably. This PR attempts to fix this by introducing some "external" types, which are intended to be exposed to users of the API in the place of the current OpId based types. We also change the representation of object IDs and element IDs to be enums with a `Root` and `Head` branch respectively, which is more idiomatic in Rust and thus hopefully more readable.

A detailed description of the design of the external types is documented in the comments at the top of `automerge/src/external_types.rs`. The short story is that we expose an `ExternalOpId` type which contains the counter and the actual bytes of the OpId, whilst all derivative types which may contain an OpId (e.g. ElemIds, or ObjectIds, or Keys) just contain a refereence to an `ExternalOpid`. 

This PR just contains the work necessary to expose the external types and line up the types on the external API. Once this change is in it will probably make sense to split `automerge::Automerge` into a facade which just performs the work of translating opids and an `Inner` type which actually performas the logic of interacting with the OpSet.

You can already see examples of how much this cleans up the API if you look at the tests. 

## Interior Mutability

The biggest change which is required here is that `OpSet::m: OpSetMetadata` has  become an `Rc<RefCell<OpSetMetadata>>`. This is because now that we are using external opids we must import these opids into the metadata even for "read only" operations - otherwise we have no basis for comparing query operations with internal operations. This does expose us to the risk of violating the contract of `borrow` calls on `Refcell`. For now we rely on tests to catch these situations (and they did several times in the writing of this code) but once the split into Facade and Inner I spoke about above is complete then we could move the `OpSetMetadata` to be held by the facade and passed in by reference to the `OpSet`. This would allow us to easily reason about when mutable borrows are active as they would only be active in the scope of functions in the facade.

## Performance Impact

The benchmarks for save and load are unchanged. The replay benchmark is affected, on my machine it is anywhere from 5-15% slower. The majority of this additional time seems to be spent allocating and pushing into the vector of operations IDs which are returned by the `splice` call. This may be to do with external OpIds being unable to benefit from cache locality of some kind due to being larger? I don't know. Regardless I don't think this is too important. We're still talking about 380ms to _replay_ the entire document and if we want to recoup that performance we can add an option to `splice` to not return the OpIds and thus skip all the tedious allocating and copying. In my opinion this is not worth it until we have actual examples of users who would be impacted by this.

### DIFF
--- a/automerge-wasm/src/lib.rs
+++ b/automerge-wasm/src/lib.rs
@@ -8,6 +8,7 @@ use std::collections::{HashMap, HashSet};
 use std::convert::TryFrom;
 use std::convert::TryInto;
 use std::fmt::Display;
+use std::str::FromStr;
 use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
 
@@ -149,7 +150,7 @@ impl Automerge {
     }
 
     pub fn keys(&mut self, obj: JsValue, heads: JsValue) -> Result<Array, JsValue> {
-        let obj = self.import(obj)?;
+        let obj: automerge::ObjId = self.import(obj)?;
         let result = if let Some(heads) = get_heads(heads) {
             self.0.keys_at(obj, &heads)
         } else {
@@ -162,7 +163,7 @@ impl Automerge {
     }
 
     pub fn text(&mut self, obj: JsValue, heads: JsValue) -> Result<JsValue, JsValue> {
-        let obj = self.import(obj)?;
+        let obj: automerge::ObjId = self.import(obj)?;
         if let Some(heads) = get_heads(heads) {
             self.0.text_at(obj, &heads)
         } else {
@@ -179,7 +180,7 @@ impl Automerge {
         delete_count: JsValue,
         text: JsValue,
     ) -> Result<(), JsValue> {
-        let obj = self.import(obj)?;
+        let obj: automerge::ObjId = self.import(obj)?;
         let start = to_usize(start, "start")?;
         let delete_count = to_usize(delete_count, "deleteCount")?;
         let mut vals = vec![];
@@ -214,7 +215,7 @@ impl Automerge {
         value: JsValue,
         datatype: JsValue,
     ) -> Result<JsValue, JsValue> {
-        let obj = self.import(obj)?;
+        let obj: automerge::ObjId = self.import(obj)?;
         //let key = self.insert_pos_for_index(&obj, prop)?;
         let index: Result<_, JsValue> = index
             .as_f64()
@@ -235,7 +236,7 @@ impl Automerge {
         value: JsValue,
         datatype: JsValue,
     ) -> Result<JsValue, JsValue> {
-        let obj = self.import(obj)?;
+        let obj: automerge::ObjId = self.import(obj)?;
         let prop = self.import_prop(prop)?;
         let value = self.import_value(value, datatype)?;
         let opid = self.0.set(obj, prop, value).map_err(to_js_err)?;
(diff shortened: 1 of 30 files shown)


orionz:
Ok.  A bunch of things jump out at me right now.

We don't need interior mutability.  Read only actions can call `m.lookup` instead of `m.cache`.

There is no need for an External Key or External ElementId.  Key and ElemId are internal only. 

On my machine the performance impact of this patch is around 50ms (360ms -> 410ms).  Given that my goal is to get the edit trace running in 50ms total, this is too high a cost.  Now that said I think something else is going on.  I made a local edit where I exposed a `splice_internal()` method which behaved exactly like the old function and it was also running 50ms slower.  Something else is responsible for the slowdown.

I really liked the only externally facing type being OpId (since Obj, Element and Key all became internal).  Is there some way we can simplify things down to a single type again?

I don't like changes that add code complexity and slow things down.  Feels like we're going the wrong way.  Worth talking over but if we HAVE to go this way I'd like to consider one kind of opid we can use efficiently and correctly internally and externally.  Certainly we can cut a lot of code if the IndexCahces, Importable, Exportable, lamport_cmp code could go away.

Also is there any other way we can fix the footgun with opid's not working between databases?  One idea (not very serious but I want to get the ideas going) would be to change OpId from `OpId( counter, actor_index )` to `OpId( counter, actor_index, automerge_id)` where automerge id is a random number generated by the document a creation time.  When one is passed in we just check to make sure the ideas match so we can throw an error if someone passes one if from another db.  Again - not super serious but want to make sure we consider other solutions to the problem.

orionz:
One experiment would be to change OpId to `OpId(usize, Rc<ActorId>)` and use that internal and external and see what the performance impact would be.  It would certainly result in less code overall.  

Another experiment would be to stick the IndexedCache metadata in some globally accessible context so all automerge documents share the same cache's and therefore all the OpId's will be common between documents.

alexjg:
> We don't need interior mutability. Read only actions can call m.lookup instead of m.cache

:man_facepalming: I'll give that a shot

> On my machine the performance impact of this patch is around 50ms (360ms -> 410ms). Given that my goal is to get the edit trace running in 50ms total, this is too high a cost. Now that said I think something else is going on. I made a local edit where I exposed a splice_internal() method which behaved exactly like the old function and it was also running 50ms slower. Something else is responsible for the slowdown.

Confusing, I'll dig a little more. I was operating under the assumption that the additional time was due to allocations in the result vector - but that's not the case if you're using splice_internal I think.

> I really liked the only externally facing type being OpId (since Obj, Element and Key all became internal). Is there some way we can simplify things down to a single type again?

I agree that getting rid of element and key is good. I'm not sure about `Obj`. I think I have an aesthetic preference for object ID being a different type to the OpID. The root object ID is not really an operation in any sense, it's not a thing that anyone has done, it has no concept of an actor. I find it clearer to pick that out in the types. I also think it has pedagogical value because when someone sees `Into<ObjectId>` in the type signatures on `Automerge::set` etc. they can then go to the documentation for `ObjectId` where we can explain how objects are identified by the operation that creates them and what the role of the root object ID is in automerge. The `Into<ObjectId>` implementation for `Opid` means that there is no syntactic overhead to doing this. Finally, I think it's more idiomatic Rust. All that said, it's not a hill I want to die on, if these arguments don't sway you I can accept just having the one OpID type.

> automerge id is a random number generated by the document a creation time

I thought about this initially. It does fix part of the footgun. But I think it's still not a great API. In particular it would not allow for serializing and deserializing OpIDS - which people will definitely want to do - so we would need to have an import/export API anyway and so I don't think it buys us a great deal.

> One experiment would be to change OpId to OpId(usize, Rc<ActorId>) and use that internal and external and see what the performance impact would be. It would certainly result in less code overall.~

I'll give this a try and report back.

alexjg:
In the end we decided to solve this problem a little differently.

PR #256, opened 2022-01-08
Sync fixes

orionz:
Got all the javascript sync tests passing - found one bug in the rust - cleaned up and added some features to the wasm api - hand ported all the old sync tests over to wasm.  Mostly js changes but if felt pretty beefy and wanted to get feedback before merging.

### DIFF
--- a/automerge-wasm/src/lib.rs
+++ b/automerge-wasm/src/lib.rs
@@ -1,13 +1,19 @@
-extern crate web_sys;
 use automerge as am;
-use automerge::{Change, ChangeHash, ObjId, Prop, Value, ROOT};
-use js_sys::{Array, Object, Reflect, Uint8Array};
-use std::collections::{HashMap, HashSet};
-use std::convert::TryFrom;
+use automerge::{Change, ObjId, Prop, Value, ROOT};
+use js_sys::{Array, Object, Uint8Array};
 use std::convert::TryInto;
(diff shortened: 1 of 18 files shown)


automerge-wasm/src/lib.rs
```diff
@@ -686,15 +739,100 @@ fn to_js_err<T: Display>(err: T) -> JsValue {
...
+fn get<J: Into<JsValue>>(obj: J, prop: &str) -> Result<JS, JsValue> {
+    Ok(JS(Reflect::get(&obj.into(), &prop.into())?))
 }
 
 fn set<V: Into<JsValue>>(obj: &JsValue, prop: &str, val: V) -> Result<bool, JsValue> {
     Reflect::set(obj, &prop.into(), &val.into())
 }
 
 struct JS(JsValue);
+struct AR(Array);
```
alexjg, line 748: This `JS` and `AR` stuff feels like it could be split into a separate file in this crate. I think it would make things a little more readable, especially as you could move all the `From`, `TryFrom`, `Into` stuff there as well.

orionz: Thats a really good point.  The file is getting super cluttered.  Broke it up into 4.  Still not sure the right way to write Javascript iterop code but its evolving and I think its getting better.

alexjg (approved):

PR #257, opened 2022-01-13
Fix bug with deleting a counter from a list - remove vis window 

orionz:
There was a subtle bug where the index being kept for list length couldn't properly compute length because it never knew how many of the successors were increments and how many were overwrites or deletes.  By moving an increment count and counter total into the counter object itself we fix this, also remove the need for a vis window outside of `_at(clock)` queries, and simplify the code.  This also fixes a subtle bug where a `values()` call including a counter will come out in the wrong order (taking the ordering of the final inc() instead).  All told I saw about a 7% speedup on the edit trace with this patch.


automerge/src/value.rs
```diff
@@ -158,29 +158,47 @@ pub(crate) enum DataType {
...
+#[derive(Serialize, Debug, Clone)]
 #[serde(untagged)]
 pub enum ScalarValue {
     Bytes(Vec<u8>),
     Str(SmolStr),
     Int(i64),
     Uint(u64),
     F64(f64),
-    Counter(i64),
+    Counter(i64, i64, usize),
```
alexjg, line 169: Given that this is three integer fields rather than just a wrapper now I think it might be worth naming the fields, otherwise anytime someone reads this they need to go and look at the places the counter is used to figure out which of the fields means what. Named fields would also make the various pattern matches on this enum throughout the codebase a little easier to read.

I'm also a little concerned that it's in our public API and wonder whether it would be worth having a separate counter struct with crate private fields which we put in the `ScalarValue::Counter`, that way users don't know or care about these details.

orionz: Good call - I made a struct `Counter` to contain this data.

PR #260, opened 2022-01-18
Add length tests to props tests

jeffa5:
Getting the length of a `Map` object doesn't seem to work.

At some point we'll also want to add a test for a case where an object gets changed from a list to an object and vice-versa to check we get the length of each respective one in history correctly.

### DIFF
--- a/automerge-wasm/src/interop.rs
+++ b/automerge-wasm/src/interop.rs
@@ -289,7 +289,7 @@ pub(crate) fn get_heads(heads: JsValue) -> Option<Vec<ChangeHash>> {
     JS(heads).into()
 }
 
-pub(crate) fn map_to_js(doc: &mut am::Automerge, obj: &ObjId) -> JsValue {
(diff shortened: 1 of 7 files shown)


jeffa5:
@orionz I've had a go at fixing it but not sure if it might cause other issues.

alexjg (approved):

orionz:
Oh this is interesting.  I was thinking length that `length` or `length_at` was list only and undefined when called on a map.  If what you're going for is `get_keys().len()` we'll need to fix more than this.  The non `at` version just looks at the index which only tracks elemids.  And the solution here would double count a key if it has 2 conflicting values.   I think the correct fix here would be to - in the top level `len` and `len_at` method detect if the object in question is a map and if so run `keys` or `keys_at` and call `len` on the result and return it.  Another solution would be to throw an error if length is called on a map.

alexjg (approved):

PR #262, opened 2022-01-19
Experimental View over document

jeffa5:
This allows us (at least in the Rust side) to traverse the document a bit more idiomatically and also edit it!

### DIFF
--- a/automerge/src/automerge.rs
+++ b/automerge/src/automerge.rs
@@ -9,8 +9,8 @@ use crate::types::{
     ActorId, ChangeHash, Clock, ElemId, Export, Exportable, Key, ObjId, Op, OpId, OpType, Patch,
     ScalarValue, Value,
 };
-use crate::value_ref::MapRef;
-use crate::value_ref::MapRefMut;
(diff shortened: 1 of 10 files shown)


alexjg:
I like the `Borrow<Exid>` change on the core API. I also like the `*View` stuff. However, I think it might make more sense for that to be in another library. My feeling is that we should keep the `automerge-rs` core API pretty low level and minimal. Is it possible to build the `*View` stuff using the public API?

jeffa5:
@alexjg thanks for the feedback, I agree that it isn't core to this repo. I've had a go at doing it another library which is mostly building but currently just gets stuck on `get_heads` requiring a mutable self due to the hidden transactions, this is a bit of a strange signature as it clearly shouldn't be mutating the document from its name. Perhaps we could make transactions more explicit in this rust API so we can have more natural signatures and then things like the hidden transactions can come at a different level? What are your thoughts too @orionz ?

jeffa5:
Here is the current pr for the view stuff https://github.com/jeffa5/automergeable/pull/8

PR #271, opened 2022-02-16
Transaction API

jeffa5:
In Rust we can leverage the borrow-checker to ensure that unsafe mid-change operations don't compile, for instance calling `merge` or `apply_changes` whilst we have an open change. This was originally being done manually by using `ensure_transaction_closed` in each function call. This means that methods that don't mutate the document no longer need to take a `&mut self`, rather just a `&self`.

This PR also adds a version of the original `Automerge`, using an internally stored transaction for use with interop such as wasm/js (`AutoTxn`, happy for suggestions for the name).

Overall, these changes make the library more idiomatically Rust, whilst still retaining the convenient automatic transactions behaviour from before, allowing more things to be built on-top of the library.


orionz:
Man - this is great.  I really hated those `ensure_transaction_closed()` statements.   As best as I can tell there are three ways to make an op now.  Explicitly make a transaction and then perform an op against it.  Have AutoTxn which behaves like the old Automerge.  And then there is now a third way where the base Automerge object makes a one op transaction for each op.  I really would like to avoid this third scenario.  Given how costly each transaction is I don't want it to be easy to commit each op on its own.  I would much rather we just rip out all of the write functions from Automerge and make you either pick explicit transaction or AutoTxn.   Im curious as to why you added this behavior?  Was it to make the tests simpler?  Maybe instead switch all the tests to `AutoTxn::init()` and support the old api only there?  

Also I noticed there are a few of the underscore functions still laying around - such as `fn _get_changes` - given these changes I'd love to see these all taken out - I only added them to avoid calling ensure_transaction_closed over and over in calls that needed these methods.

jeffa5:
Yep, you're right with those three ways of making an op, and yes, I was basically having the top level set etc for ease of testing. I've switch most over to the AutoTxn now, using the explicit tx for others.

Also removed the underscore functions.

I've updated the commit API to be more Rusty too as it is annoying having to pass `(None, None)` in every time. This leaves the js interop the old way though, not sure if that might be better realised as a dict of optional things?

Another addition is the `transact` and `transact_with` functions (and associated types) so that Rust users don't have to forget to manually commit the transaction.

I've also added in a `rollback` on drop for the explicit transactions to ensure we don't break the document (this is skipped if manually calling `commit` or `rollback`).

automerge/src/transaction/result.rs
```diff
@@ -0,0 +1,54 @@
+use crate::ChangeHash;
+
+/// The result of a successful, and committed, transaction.
+#[derive(Debug)]
+pub struct TransactionSuccess<O> {
+    pub(crate) result: O,
+    pub(crate) heads: Vec<ChangeHash>,
+}
```
jeffa5, line 8: Should we just make these `pub` rather than having the accessor functions?

orionz:
Awesome - love all these changes - Nothing else jumps out at me.  I'd like to get @alexjg's input here as he's a consumer of the rust api as well.  What did the benchmarks say as far as the performance impact of all this?

alexjg:
In general this looks really nice. I'll review in detail over the weekend.

jeffa5:
> What did the benchmarks say as far as the performance impact of all this?

About 25ms difference on my machine (474ms for manual, 499 for auto)

automerge/src/transaction.rs
```diff
@@ -0,0 +1,203 @@
...
+        heads: &[ChangeHash],
+    ) -> Result<Vec<(Value, ExId)>, AutomergeError> {
+        self.doc.values_at(obj, prop, heads)
+    }
+}
+
+// If a transaction is not commited or rolled back manually then it can leave the document in an
+// intermediate state.
+// This defaults to rolling back the transaction to be compatible with `?` error returning before
+// reaching a call to `commit`.
```
alexjg, line 196: This means we could be doing quite a bit of work in the Drop implementation so we should document that `Drop` behaviour on the `Transaction` I think.

automerge/src/transaction.rs
```diff
@@ -0,0 +1,203 @@
...
+        self.inner.as_ref().unwrap().pending_ops()
+    }
+
+    /// Commit the operations performed in this transaction, returning the hashes corresponding to
+    /// the new heads.
+    pub fn commit(mut self) -> Vec<ChangeHash> {
+        self.inner.take().unwrap().commit(self.doc, None, None)
+    }
+
+    pub fn commit_with(mut self, options: CommitOptions) -> Vec<ChangeHash> {
```
alexjg, line 46: We should add an example here of using the `CommitOptions` builder

automerge/src/autotxn.rs
```diff
@@ -0,0 +1,417 @@
...
+use crate::types::Patch;
+use crate::{
+    change::export_change, transaction::TransactionInner, ActorId, Automerge, AutomergeError,
+    Change, ChangeHash, Prop, Value,
+};
+use crate::{SyncMessage, SyncState};
+
+/// An automerge document that automatically manages transactions.
+#[derive(Debug, Clone)]
+pub struct AutoTxn {
```
alexjg, line 12: I wonder if this might be better called `AutoCommit` to evoke autocommit options on various SQL database adapters? I don't feel strongly about this, just a thought.

automerge/src/autotxn.rs
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    pub fn pending_ops(&self) -> usize {
+        self.transaction
+            .as_ref()
+            .map(|t| t.pending_ops())
+            .unwrap_or(0)
+    }
+
+    fn try_start_transaction(&mut self) {
```
alexjg, line 65: I think a better name for this would be `ensure_transaction_open` for the symmetry with `ensure_transaction_closed`

automerge/src/autotxn.rs
```diff
@@ -0,0 +1,417 @@
...
+
+    pub fn commit_with(&mut self, options: CommitOptions) -> Vec<ChangeHash> {
+        self.try_start_transaction();
+        self.transaction
+            .take()
+            .map(|tx| tx.commit(&mut self.doc, options.message, options.time))
+            .unwrap_or_else(|| self.doc.get_heads())
+    }
+
+    pub fn ensure_transaction_closed(&mut self) {
```
alexjg, line 117: Could this `ensure_transaction_closed` stuff be avoided by making `AutoTxn` an enum over two possible states. I.e.

```rust
enum AutoTxn {
    InTransaction{
        doc: Automerge,
        transaction: TransactionInner,
    },
    Clean(Automerge),
}
```

This way any code which needs access to the `TransactionInner` would first have to transition the state to `InTransaction`.

automerge/src/lib.rs
```diff
@@ -38,12 +39,14 @@ mod op_set;
 mod op_tree;
 mod query;
 mod sync;
+pub mod transaction;
```
alexjg, line 42: Does this module need to be part of our public API or can we just expose the `Transaction` type?

automerge/tests/test.rs
```diff
@@ -8,11 +8,11 @@ use helpers::{
 };
 #[test]
 fn no_conflict_on_repeated_assignment() {
-    let mut doc = Automerge::new();
+    let mut doc = AutoTxn::new();
     doc.set(&automerge::ROOT, "foo", 1).unwrap();
     doc.set(&automerge::ROOT, "foo", 2).unwrap();
     assert_doc!(
-        &doc,
+        doc.document(),
```
alexjg, line 15: Rather than calling `doc.document()` on every test I think it might be better to modify the `assert_doc` macro to make this call.

automerge/tests/helpers/mod.rs
```diff
@@ -6,12 +6,12 @@ use std::{
 
 use serde::ser::{SerializeMap, SerializeSeq};
 
-pub fn new_doc() -> automerge::Automerge {
-    automerge::Automerge::new_with_actor_id(automerge::ActorId::random())
+pub fn new_doc() -> automerge::AutoTxn {
```
alexjg, line 9: I think it might be worth changing these tests to use the transaction API rather than the `AutoTxn` API. There are several tests which I'm not really sure are semantically equivalent with the `AtoTxn` API because the boundaries of the changes are different. This may be quite manual work though which I am happy to take on. I.e. this is more of a note to myself than a request for a change in this PR 🙂

edit-trace/benches/main.rs
```diff
@@ -1,26 +1,44 @@
...
 
-fn replay_trace(commands: Vec<(usize, usize, Vec<Value>)>) -> Automerge {
+fn replay_trace_tx(commands: Vec<(usize, usize, Vec<Value>)>) -> Automerge {
     let mut doc = Automerge::new();
+    let mut tx = doc.transaction();
+    let text = tx.set(&ROOT, "text", Value::text()).unwrap().unwrap();
+    for (pos, del, vals) in commands {
+        tx.splice(&text, pos, del, vals).unwrap();
+    }
+    tx.commit();
```
alexjg, line 12: Is this really a valid benchmark? This generates one commit right at the end of all the edits, which seems like a pretty unlikely workflow. On the other hand, it probably doesn't make sense to generate a transaction on every single change (although I'm sure there are users who will). Probably the most rigorous way to approach this is to parameterise this benchmark by some kind of transaction frequency and measure performance across all of them. Is this something we care about doing or should we just view this benchmark as somewhat synthetic?

alexjg (commented):
I really like this work, removing the `&mut` on so many methods is fantastic. I've left a few comments and I also have one larger thought: We now have duplicated methods on `AutoTxn` and on `Transaction`. I think it would make sense to put these into a trait which is implemented for both structs. At the very least this would give us a single place to document things, but it also ensures that the method signatures are the same so the client code which starts out using `AutoTxn` doesn't have to change much when introducing transactions. I think that's worth doing now rather than leaving until later because these interfaces could quickly drift if we're not careful.

automerge/src/transaction.rs
jeffa5: Added.

automerge/src/autotxn.rs
jeffa5: Yes! That makes much more sense.

edit-trace/benches/main.rs
orionz: There is no reason to cut a transaction unless you are doing real time collaboration or saving the doc to disk.  So this (in my mind) represents someone editing offline and hitting "save" at the end of the edit session.

alexjg: Hmm. Yeah I think that makes sense, nice!

PR #272, opened 2022-02-16
Add docs workflow for gh-pages

jeffa5:
Before we release to crates.io and whilst we are developing changes it is useful to be able to visit the built docs. This PR adds the github action to build and push them to the `gh-pages` branch.

@orionz this would need the github pages to be setup under settings, I think you have access to do that if you'd be able to, just needs something like this 
image

An example is available [on my branch](http://jeffas.io/automerge-rs/automerge/)

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,0 +1,38 @@
+on:
+  push:
+    branches:
+      - main
+
+name: Documentation
+
+jobs:
+  deploy-docs:
+    concurrency: deploy-docs
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+
+      - name: Cache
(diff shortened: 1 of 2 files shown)


okdistribute:
https://automerge.github.io/ is currently taken by the Automerge docs (github.com/automerge/automerge.github.io)

I dont think that will work on a subdomain.

jeffa5:
Pushed but doesn't seem to have done anything :/

orionz:
Ok - i pressed the button i needed to press and now it hits the index.html in gh-pages which redirects it to `./automerge/` which is a 404 - theres a bunch of stuff in `./docsrs`  - seems like we're almost there

orionz:
> https://automerge.github.io/ is currently taken by the Automerge docs (https://github.com/automerge/automerge.github.io)
> 
> I dont think that will work on a subdomain.

This will be at `https://automerge.github.io/automerge-rs/` - shouldn't conflict

orionz:
Wow!  This is really nice to read just from an `auditing our public facing interfaces` standpoint!! 😍 

Get the clippy passing so I can merge this!

jeffa5:
Fixed!

PR #274, opened 2022-02-21
Cleaned up some warnings and errors running against Rust 1.58.1

dpp:
Cleaned up `main` by compiling against Rust 1.58.1 and:

* Cleaning up warnings
* Upgrading packages to most recent version
* Fixed some errors

All tests pass


orionz:
That feeling when you sit down to do some work and find a PR for it.  =D 

Thank you so much!

PR #275, opened 2022-02-21
Add Doxygen documentation to the experimental C API

jkankiewicz:
@orionz, I'm submitting the changes that I've fully tested so far in order to find out if I'm on the right track or not.

The HTML documentation pages can be built with CMake and viewed with Firefox like so:
```
cd automerge-rs/automerge-c
mkdir build
cd build
cmake -S .. -DCMAKE_BUILD_TYPE=Release
cmake --build . --target automerge_docs
firefox src/html/index.html
```

Changes:
* Added Doxygen documentation generation because, IMHO, one header containing everything is too busy to also serve as documentation for the C API.
* Changed `AmDataType::Null` to have integer value `0` because `NULL == 0`.
* Alphabetized the other tags within `AmDataType`.
* Renamed `AMclone()` to `AMdup()` because it's similar to the C standard `strdup` function.
* Renamed `AMfree()` to `AMdestroy()` because there's no `AMmalloc()` and "destroy" is an antonym of "create".
* Renamed `AMDatatype` to `AmDataType` because "data type" is two words.


jkankiewicz:
The following copy of the generated documentation is for those who'd like to review its contents without having to build it for themselves.

It's in PDF form only because I don't know of a cross-platform way to pack a website into a single file that can be viewed from within a web browser.
[refman.pdf](https://github.com/automerge/automerge-rs/files/8111867/refman.pdf)

orionz:
Oh no - I am **SO** sorry.  I did a bunch of work changing the whole C API around to look more like postgresql and didnt push it - you're working against the old version.  Can you get on our discord so we can chat about this off of github.  We need to be in tighter sync to avoid these kinds of issues.

jkankiewicz:
> Oh no - I am SO sorry. I did a bunch of work changing the whole C API around to look more like postgresql and didnt push it - you're working against the old version.

@orionz, it's okay; any outside developer who opens a PR runs the risk that it will become irrelevant.

jkankiewicz:
> I did a bunch of work changing the whole C API around to look more like postgresql...

@orionz, I've adapted my changes to those you've most recently pushed.

> Can you get on our discord so we can chat about this off of github.

I can't find any invitations for the Discord server, I'm only aware of the Slack. I sent you a private message on the Slack just in case you were referring to it.

orionz:
This all looks great!  I'll look for your slack message and lets get you that discord invite.

PR #278, opened 2022-02-25
Make keys and keys_at iterators

jeffa5:
This helps reduce unneeded work such as allocation of `Vecs`, especially intermediate ones and also allows us to iterate straight from the back.


orionz:
This is excellent and very idiomatic!  I wonder where else we can use this pattern.

alexjg:
I made a comment on this but somehow it got lost. Is it possible to implement `ExactSizeIterator` as well?

jeffa5:
I considered implementing `ExactSizedIterator` too but wasn't sure. We may be able to do it for Keys if there is a len stored somewhere but for KeysAt I think we'd need to iterate through all of the values to get the length which I wasn't too sure on. Happy to add it as a todo.

jeffa5:
> This is excellent and very idiomatic! I wonder where else we can use this pattern.

Thanks! I think we might be able to add some proper iterator system to the OpTrees but it might require some unsafe so not immediately easy.

alexjg:
> I considered implementing `ExactSizedIterator` too but wasn't sure. We may be able to do it for Keys if there is a len stored somewhere but for KeysAt I think we'd need to iterate through all of the values to get the length which I wasn't too sure on. Happy to add it as a todo.

I don't think this is a big deal, it was more a question of if it's easy then it would be worth doing.

PR #283, opened 2022-03-02
Separate scalars and objects in transaction API

jeffa5:
This introduces a separation between the creation of objects and scalars in the API. This makes it safer to use the object id that gets returned when creating an object, as it will always be returned, and nothing will be returned for using a scalar.

### DIFF
--- a/automerge/src/autocommit.rs
+++ b/automerge/src/autocommit.rs
@@ -248,12 +248,12 @@ impl AutoCommit {
     /// ```
     /// # use automerge::transaction::CommitOptions;
     /// # use automerge::transaction::Transactable;
-    /// # use automerge::Value;
     /// # use automerge::ROOT;
     /// # use automerge::AutoCommit;
+    /// # use automerge::ObjType;
     /// # use std::time::SystemTime;
     /// let mut doc = AutoCommit::new();
-    /// doc.set(ROOT, "todos", Value::list()).unwrap();
+    /// doc.make(ROOT, "todos", ObjType::List).unwrap();
     /// let now = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).unwrap().as_secs() as
     /// i64;
     /// doc.commit_with(CommitOptions::default().with_message("Create todos list").with_time(now));
--- a/automerge/src/autocommit.rs
+++ b/automerge/src/autocommit.rs
@@ -253,7 +253,7 @@ impl AutoCommit {
     /// # use automerge::ObjType;
     /// # use std::time::SystemTime;
     /// let mut doc = AutoCommit::new();
-    /// doc.make(ROOT, "todos", ObjType::List).unwrap();
+    /// doc.set_object(&ROOT, "todos", ObjType::List).unwrap();
     /// let now = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).unwrap().as_secs() as
     /// i64;
     /// doc.commit_with(CommitOptions::default().with_message("Create todos list").with_time(now));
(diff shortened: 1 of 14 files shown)


orionz:
Alternatively we could just have `make()` and `set()` and pass insert as a param.  This is how the old version of automerge worked but I was never very fond of it.  That said im unsure which of these options is best.  Open to suggestions.

jeffa5:
I've worked a bit more on this and think the separation is quite useful, mainly in knowing when to expect an object id and when not to. Rather than the `make` naming I think `{op}_object` serves the user a bit better, telling them more explicitly that it is for objects.

So now:
1. `set` and `insert` only take a scalar value and and don't return anything in the successful case
2. `set_object` and `insert_object` are new functions for explicitly making objects, they return the object id in case of success
3. `splice` takes a list of scalar values, rather than allowing objects and doesn't return anything (previously returned a list of the IDs which seemed a bit awkward to me).

I've implemented the corresponding functions and changes in the wasm api. If this seems like a good direction then I can have a go at updating the js side to use these in the proxy, which is currently causing the js tests to fail.

automerge/src/autocommit.rs
```diff
@@ -321,28 +321,50 @@ impl Transactable for AutoCommit {
...
     /// - The key is the wrong type for the object
     /// - The key does not exist in the object
-    fn set<P: Into<Prop>, V: Into<Value>>(
+    fn set<P: Into<Prop>, V: Into<ScalarValue>>(
         &mut self,
         obj: &ExId,
         prop: P,
         value: V,
-    ) -> Result<Option<ExId>, AutomergeError> {
+    ) -> Result<(), AutomergeError> {
```
alexjg, line 329: Can we modify the `set` and `insert` methods to only accept a `ScalarValue` and likewise check in the `set_object` and `insert_object` that the arguments are in fact objects? Otherwise I think we run the risk of people calling `insert_object` with a non-object value and getting back something they think is an object ID but which is actually just an opid?

alexjg: Oh wait, I think that's already covered by the `Into`s, my bad.

jeffa5: Yep, I think those are covered

orionz:
Ergonomics-wise - how's the feeling for the two of you?  Better API?

jeffa5:
I like that from a Rust point of view but what would be the analogous js side style? Would we still have to do a check on the result for opid or null in js?

orionz:
> I like that from a Rust point of view but what would be the analogous js side style? Would we still have to do a check on the result for opid or null in js?

Well - we dont NEED to change anything in the js api - its working fine now - but yeah if we keep the make() i'll need to figure out how to insert make or just leave it as is - you can only do so much with js

jeffa5:
Having a look at implementing the generic version, it isn't playing nicely with the `Into`s. Getting a lot of `type annotations needed` when trying to add the `Into` behaviour back and without it we have to manually do `ScalarValue::Str("mystring".into())` rather than just `"mystring"`. I've pushed what I've got so far to https://github.com/automerge/automerge-rs/pull/290, happy for any comments on it or ways to make it nicer (there are some things that will need cleaning up after it works).

The other side of the coin is readability, without the IDE hints it is difficult to see whether we should be getting an id back or not (if the compiler can even get the types). I think this may be hurting readability a bit.

alexjg:
Ha, I had a play and encountered basically the same issues :joy:. Additionally, we would need to expose `OpId` in our public API as the `CanSet` trait is public, and I really really want to avoid exposing `OpId` if we can.

Apologies for the blind alley, I just can't resist a fancy type.

jeffa5:
> Ha, I had a play and encountered basically the same issues joy. Additionally, we would need to expose `OpId` in our public API as the `CanSet` trait is public, and I really really want to avoid exposing `OpId` if we can.

Yeah, it is very neat but I think asking a bit too much of the compiler, I don't think it's that bad having the separation between objects and scalars.
 
> Apologies for the blind alley, I just can't resist a fancy type.

Haha, no problem, was fun to play with.

jeffa5:
I've fixed up the js proxy now too so all tests pass on my machine.

PR #284, opened 2022-03-02
Always have an actor on the document

jeffa5:
This helps to clean up the Rust API where getting the actor required `&mut self`.

### DIFF
--- a/automerge/src/autocommit.rs
+++ b/automerge/src/autocommit.rs
@@ -44,13 +44,6 @@ impl AutoCommit {
         self.doc.get_actor()
     }
 
-    pub fn new_with_actor_id(actor: ActorId) -> Self {
-        Self {
-            doc: Automerge::new_with_actor_id(actor),
-            transaction: None,
-        }
-    }
-
     fn ensure_transaction_open(&mut self) {
         if self.transaction.is_none() {
             let actor = self.doc.actor;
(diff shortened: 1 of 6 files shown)


automerge/src/automerge.rs
```diff
@@ -27,76 +27,43 @@ pub struct Automerge {
...
+        am
     }
 
     pub fn set_actor(&mut self, actor: ActorId) {
-        self.actor = Some(self.ops.m.actors.cache(actor))
+        self.actor = self.ops.m.actors.cache(actor)
     }
 
-    fn random_actor(&mut self) -> ActorId {
+    fn set_random_actor(&mut self) {
```
alexjg, line 55: I wonder if this should actually be something like `with_random_actor(self) -> Self`. Would make the uses of it read a little nicer.

automerge/src/autocommit.rs
```diff
@@ -40,24 +40,13 @@ impl AutoCommit {
         self.doc.set_actor(actor)
     }
 
-    pub fn get_actor(&mut self) -> ActorId {
+    pub fn get_actor(&self) -> &ActorId {
         self.doc.get_actor()
     }
 
-    pub fn maybe_get_actor(&self) -> Option<ActorId> {
```
alexjg, line 47: So glad to get rid of this :slightly_smiling_face:

alexjg (approved):
Very nice!

jeffa5:
Turns out that this optional functionality may be required to avoid adding actors without changes to the doc on serialization so I'll have a think about this a bit more.

automerge/src/automerge.rs
jeffa5: This isn't there any more, resolving

orionz:
Nice!

PR #286, opened 2022-03-02
import cli

orionz:
I imported the automerge-cli code over.  The "change" module will require significant changes do I commented it out for now.  For the rest I was able to get them compiling and running with minimum effort.

Hopefully this will be good enough to bring the tool back into the fold.


alexjg:
CI is failing because this introduces a transitive dependency on `colored_json`, which is licensed under EPL-2.0. I don't think this should be a problem for us as based on my understanding of the EPL this does not require us to do anything as we only depend on `colored_json` and don't modify it. @orionz does that sound reasonable to you? If so I can allow the EPL-2.0 license in `deny.toml`

alexjg:
Having said that, I think this would make us incompatible with anyone who wants to depend on us in a GPL licensed work (it's in here: https://www.gnu.org/licenses/license-list.html#GPLIncompatibleLicenses). That's probably okay as this is just the CLI, but I'm not sure.

orionz:
Since the CLI is a standalone app and not a library they people link to this should be fine.  That said since this is a mono-repo we have to allow or deny to all apps.  Would be nice to just find an alternative to the offending library.

alexjg (approved):

PR #287, opened 2022-03-03
AsRef exid to avoid &ROOT everywhere

jeffa5:


jeffa5:
I'm not sure how much of an effect this might have on the C library growing in size...

alexjg:
What's the rationale for using `Borrow` instead of `AsRef`? Are there places where we need the extra invariants `Borrow` implies?

orionz:
Man this generic method params is some clever rust!  Love it.  Is there a cost to doing this?  Code readability?  Documentation readability?  Executable size?  Performance?  Overall seems like a great strategy.  Surprised I don't see this in more libraries.

alexjg:
Probably the main downside is executable size. Generic functions are sort of implicitly inlined (due to Rust monomorphizing everything) so you end up with a copy of each of these functions for each implementor of `AsRef`. If there are lots of different types which implement this `AsRef` then you typically would use tricks like calling `.as_ref()` in the API and immediately dispatching to a function which does the actual work and knows the concrete type. I don't think that will be a problem here though.

Some good notes on this here: https://matklad.github.io/2021/07/09/inline-in-rust.html

alexjg (approved):

jeffa5:
> Man this generic method params is some clever rust! Love it. Is there a cost to doing this? Code readability? Documentation readability? Executable size? Performance? Overall seems like a great strategy. Surprised I don't see this in more libraries.

I agree with Alex that the main downside is likely code size. You could probably argue that readability gets a bit worse at first glance but it eases usability a lot.

PR #288, opened 2022-03-03
Add test for overwriting a map and getting value from old one

jeffa5:
This test shows that we can still access an object using its ID, even though it is no longer accessible in the document tree (using the paths). I assume that this is useful for cases such as conflicts but I wonder if there might be a way to make it more footgun-like? 

I'm ok with being able to look at old objects for historical reasons but I think trying to modify them should be an error.

As for how to detect that the object is not in the current document I'm not sure, maybe keep a running set of all the object ids in the document which we can alter as changes are made, or add a modifiable bool to each object?

### DIFF
--- a/automerge/src/automerge.rs
+++ b/automerge/src/automerge.rs
@@ -1224,6 +1224,14 @@ mod tests {
         );
 
         // but we can still access the old one if we know the ID!
-        assert_eq!(doc.value(&map1, "b").unwrap(), None);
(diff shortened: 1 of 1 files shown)


alexjg:
I think this is working as intended. I think of conflicting values as a bit like branches in git. You may want to evolve multiple conflicting versions independently and at some later point merge them, or compare them etc. In that sense being able to retrieve and update values from conflicting values is a feature.

jeffa5:
Ah, thanks. I've updated the test case to ensure this continues to work.

alexjg (approved):

PR #295, opened 2022-03-04
Make decode_change an associated function

jeffa5:
This should be associated with the change type, not just being a free function.


alexjg (approved):
Looks good.

Just FYI I have a giant patch incoming that completely reworks the encoding/decoding logic.

PR #296, opened 2022-03-04
Remove index from ExId

jeffa5:


alexjg:
Does this impact performance at all? My understanding was that carrying around an index was necessary in order to avoid slow actor ID lookups on repeated interactions with the same object ID.

jeffa5:
Benchmarking shows ~1% slowdown which to me doesn't seem too conclusive given the limited scope of the benchmark (this may be a lot worse if there are lots of actors for instance). I'll close this until another time when we can re-evaluate.

PR #297, opened 2022-03-04
Add failing tests for deleting nothing

jeffa5:
I expect we shouldn't be able to delete something that doesn't exist??

### DIFF
--- a/automerge/src/automerge.rs
+++ b/automerge/src/automerge.rs
@@ -1258,7 +1258,7 @@ mod tests {
         let last_change = doc.get_last_local_change().unwrap();
         assert_eq!(last_change.len(), 0);
 
-        let bytes = doc.save().unwrap();
+        let bytes = doc.save();
         assert!(Automerge::load(&bytes).is_ok());
 
         let mut tx = doc.transaction();


jeffa5:
I think the question is whether this should return an error or just do nothing (and not include this op in the change, effectively a noop)?

orionz:
Hmmm @alexjg what's your opinion.  Doesn't bother me that deleting nothing is not an error.  In rust, HashMap doesn't error if you delete a non-existent key.  Meanwhile Vec will panic if you remove an index that's not there...  This feels like a philosophical decision that we should make sure is followed all over the api.  Do valid but nonsensical inputs result in an error or a noop.  For instance if you ask a list instead of a map for its keys - do you get an empty set, or an Error?

I generally find interfaces that default to noop for harmless but useless input "feel" a little more approachable to new developers...  Maybe thats why javascript took off ... 😰

alexjg:
I'm inclined to think it should be a noop, I don't see an enormous downside to it.

alexjg:
I think to the philosophical point, I don't think a delete of a not present element is nonsensical, algebraically that makes a lot of sense. Where I would get worried is where a noop can't be reasoned about like this, e.g. only making a delete a noop if it deletes something that hasn't been committed in a transaction (no reason we would do that, just an example), I feel like this would create a situation where users just have to know that sometimes an op is created and sometimes it isn't.

alexjg (approved):

PR #305, opened 2022-03-09
add paths/materialize to api

orionz:
New api features of the wasm api.  Paths can be used instead of objid's. Also a materialize call has been added that recursively constructs a json object from a given path/objid.

### DIFF
--- a/automerge/src/autocommit.rs
+++ b/automerge/src/autocommit.rs
@@ -298,7 +298,7 @@ impl Transactable for AutoCommit {
         self.doc.length_at(obj, heads)
     }
 
-    fn object_type<O: AsRef<ExId>>(&self, obj: O) -> Result<ObjType, AutomergeError> {
+    fn object_type<O: AsRef<ExId>>(&self, obj: O) -> Option<ObjType> {
         self.doc.object_type(obj)
     }
 
--- a/automerge/src/autocommit.rs
+++ b/automerge/src/autocommit.rs
@@ -298,6 +298,10 @@ impl Transactable for AutoCommit {
         self.doc.length_at(obj, heads)
     }
 
+    fn object_type<O: AsRef<ExId>>(&self, obj: O) -> Result<ObjType, AutomergeError> {
+        self.doc.object_type(obj)
+    }
+
     // set(obj, prop, value) - value can be scalar or objtype
     // del(obj, prop)
     // inc(obj, prop, value)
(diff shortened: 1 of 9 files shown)


automerge/src/autocommit.rs
```diff
@@ -298,6 +298,10 @@ impl Transactable for AutoCommit {
         self.doc.length_at(obj, heads)
     }
 
+    fn object_type<O: AsRef<ExId>>(&self, obj: O) -> Result<ObjType, AutomergeError> {
```
jeffa5, line 301: Should this be `pub` too?

automerge/src/automerge.rs
```diff
@@ -236,6 +236,11 @@ impl Automerge {
         }
     }
 
+    pub fn object_type<O: AsRef<ExId>>(&self, obj: O) -> Result<ObjType, AutomergeError> {
```
jeffa5, line 239: Why not return an option since the error is just `Fail` anyway and the opset function returns an option too?

jeffa5 (changes requested):

automerge/src/error.rs
```diff
@@ -7,6 +7,8 @@ use thiserror::Error;
 pub enum AutomergeError {
     #[error("invalid opid format `{0}`")]
     InvalidOpId(String),
+    #[error("invalid actor `{0}`")]
+    InvalidActor(String),
```
alexjg, line 11: Is this necessary to have in the public API? It would be nice if we could leave it out as it exposes the internal structure of the OpIDs, which I'm really keen to avoid.

orionz: The problem was the upwelling team was feeding a valid opid into the system (but it was from the wrong document) and they were seeing "invalid opid format" - i needed a different error for when its a valid format but for an actor not in the document.   Suggestions?

automerge/src/autocommit.rs
orionz: This is part of a trait, so public by default.

automerge/src/error.rs
alexjg: Ah interesting. i think two things

1. Just having "opid not part of this document" would be as informative right? (and possibly preferable because a user who doesnt know much about actors would atill have a chance of knowing what to do)
2. We should add aome logs (via `tracing` for debugging these conditions when its a programming wrror we have made.

jeffa5 (approved):

alexjg (approved):

PR #309, opened 2022-03-11
Add `parent_object` and `path_to_object` functions

jeffa5:
In a scenario where we want to detect what key in the document has changed we can iterate through the ops in changes as they are applied, however these contain an obscure object id. This functionality helps the user by adding a method of going back up the tree from a found object to build the path to the original object.

There are currently some awkward bits that I'm not sure how we want to address such as the operation having an ObjectId whereas we need an ObjId for the document so we have to convert it to string first then import it.

### DIFF
--- a/automerge/src/automerge.rs
+++ b/automerge/src/automerge.rs
@@ -295,7 +295,7 @@ impl Automerge {
             } else {
                 self.ops
                     .parent_object(&obj)
-                    .map(|(id, key)| (self.id_to_exid(id.0), self.export_key(obj, key)))
+                    .map(|(id, key)| (self.id_to_exid(id.0), self.export_key(id, key)))
             }
         } else {
             None
@@ -308,7 +308,7 @@ impl Automerge {
             Key::Seq(opid) => {
                 let i = self
                     .ops
-                    .search(&obj, query::OpIdSearch::new(opid.0))
+                    .search(&obj, query::ElemIdPos::new(opid))
                     .index()
                     .unwrap();
                 Prop::Seq(i)
--- a/automerge/src/automerge.rs
+++ b/automerge/src/automerge.rs
@@ -287,6 +287,31 @@ impl Automerge {
     // PropAt::()
     // NthAt::()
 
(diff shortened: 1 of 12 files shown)


automerge/src/automerge.rs
```diff
@@ -182,6 +182,31 @@ impl Automerge {
     // PropAt::()
     // NthAt::()
 
+    pub fn parent_object<O: AsRef<ExId>>(&self, obj: O) -> Option<(ExId, crate::legacy::Key)> {
```
alexjg, line 185: I don't actually understand how this is compiling given that `crate::legacy::Key` is private :thinking: . In any case, I'm worried about exposing OpIDs as keys via `Key::ElementID`. Can we use indices instead? This would require a separate public `Key` enum I think.

jeffa5: I think it is public but opaque. I've actually now updated it to use `Prop` since that seems to make more sense when returning the index. Note though that the index returned currently is the index of the op in the tree, not in the users list, I'm not sure how we could handle this.

jeffa5: Seems like I could use `Nth` to convert the correct index to the one in the optree but I don't think we currently have a way of doing the inverse?

jeromegn:
I wasn't sure where to post this, but I'm having some issues with this functionality in conjunction with my #331 PR.

It seems like once I pop the patches, as I go through them the objects referred to in the patches don't have a tree and I can't get their parent. I found this out by adding a bunch of `println!` in all the functions being called from `parent_object(obj)`. The issue is `self.trees.get(obj)` returns `None` here: https://github.com/automerge/automerge-rs/pull/309/files#diff-397994bdf8ff993666ddbc597fe755520f0c7442f1b66d8d47114ed2ed579c9aR49

For example: if I get a `Patch::Assign(AssignPatch { obj, key, value, .. })`, I can find the parents for `obj`, but not for `value.1` (which is the new obj being assigned at `key`).

Is this expected?

I'm using an `Automerge` (not `AutoCommit`) document with transactions, just doing `tx.put(obj, Prop::Map("key"), am::ScalarValue::Str("foo".into()))`

Issue #311, opened 2022-03-18
Suspected incorrect list index computation


ept:
I suspect there is a bug in the computation of list indexes, which appears in [`Nth`](https://github.com/automerge/automerge-rs/blob/ebe7bae992ea715246219cf037497d1b24c0f66b/automerge/src/query/nth.rs#L42-L57) and [`InsertNth`](https://github.com/automerge/automerge-rs/blob/ebe7bae992ea715246219cf037497d1b24c0f66b/automerge/src/query/insert.rs#L61-L73). This is just based on reading the code — I've not yet written a test case for it since I'm not sure how to write the test that triggers the necessary condition.

When a list element is overwritten with a new value multiple times (possibly concurrently), the opTree will contain multiple operations for the same list element. A list element is visible if it has at least one such operation that has not been overwritten; a list element has a conflict if there are two or more non-overwritten operations for the same element.

Each tree node contains as metadata the number of visible list elements it contains. However, it is possible that the operations for the same list element don't all fall within the same tree node, i.e. they may be split across tree nodes.  If there is a conflict on the list element, the visible operations may fall within different tree nodes, and therefore the same list element would be counted towards the visible element count multiple times. To compensate for this double-counting, we need to detect whether the last visible list element in one subtree is in fact the same as the first visible element in the next subtree, and if so, subtract 1 from the visible element count.

The logic in `Nth` and `InsertNth` attempts to do this, but I think it does so incorrectly. I think there are two bugs:

1. Am I correct in thinking that `child.index` contains all list elements, regardless of whether they are visible or not? If so, the condition `child.index.has(&self.last_seen)` is incorrect: we want it to be true only if `last_seen` is visible in the tree node `child`, but in fact it will always be true if `last_seen` appears in `child`, even if it is not visible.
2. The variable `last_seen` is set to `child.last().elemid()`, which I believe is also incorrect: we want it to be the elemId of the last *visible* list element, but if I understand the code correctly, it will be set to the elemId of the last operation in the subtree `child`, regardless of whether it is visible or not.

To fix this, we need some way of determining whether the last visible list element of one subtree is the same as the first visible element of another. One way of doing this would be to scan over the ops (backward from the last op of the first subtree, and forward from the first op of the second subtree) until we find one that is not overwritten. Another option (which is what I did in the JS implementation) would be to have some extra metadata on every tree node that tells us the elemId of the first and last visible list element in that subtree, and to update this metadata as nodes are updated and split.

[...]

PR #312, opened 2022-03-18
Generate patches

ept:
With lots of help from @orionz (thank you Orion!) I've made a start at generating patches when applying remote changes. To avoid slowing down users who don't need the patches, they are only enabled if you explicitly opt in. The patches can be used to maintain a copy of a document as plain JS objects, for situations when it's undesirable to query directly into the wasm state. Patches can also help keep other state, such as UI libraries for text editing, up-to-date with the CRDT state.

Not finished yet, but I wanted to post it so that others can take a look. Known issues:

* Also need to generate patches for local updates
* Fix patch generation for operations on counters (currently panics because you can’t call `op.value()` on an increment operation)
* Index calculation is wrong when the ops for a list element are split across tree nodes, see #311

Since I am only just getting started with Rust and this codebase I would be grateful for suggestions on how to improve the code :)


alexjg:
Hooray! Martin is writing rust! 🥳

My first impression here is that there is some duplication of logic between the "with patches" and "without patches" branches - `insert_op` vs `insert_op_with_patch` and `SeekOp` vs `SeekOpWithPatch`. If I understand things correctly the logic in both these cases is identical except for the additional state needed to track the patches. I think it makes sense to allow users who aren't interested in patches to avoid the cost of tracking them, but I have some ideas which seem to me to provide a more flexible API and also help us to avoid the logic duplication. To the API. Rather than the `enable_patches()/pop_patches()` combo, I think it might be nicer to use a kind of observer which is passed to `apply_change`. Something like this maybe:

```rust

pub trait OpObserver {
    fn insert(&mut self, obj_id: ObjId, index: usize, tagged_value: (Value, ObjId)) {
    }
    fn assign(&mut self, obj_id: ObjId, key: Prop, tagged_value (Value, ObjId), conflict: bool) {
    }
    fn delete(&mut self, obj_id: Objid, key: Prop) {
    }
}

pub struct Automerge { ...

    pub fn apply_changes<O: OpObserver>(&mut self, changes: Vec<Change>, observer: Option<O>) -> Result<(), AutomergeError> { }
...
}
```

We can then switch on the `Some(observer)/None` to select whether or not patches should be collected. This has a few nice advantages to my mind:

- The choice of data structure for collecting patches is up to the user. The PR currently hard codes a `Vec<Patch>`, which implies allocating on every change. This trait based API is a bit more flexible in that it allows the user to do things like update a boolean flag if a certain path is modified and otherwise ignore changes. Or to collect changes directly into some structure which is appropriate for the user's usecase without going through a `Vec` first.
- It's easier to make backwards compatible changes. With the `Patch` data structure any additional branch in the enum will be a breaking change, with the trait we can add default methods for new branches and remain backwards compatible.
- It opens the door to API's like "filtering" ops, which I am interested in, where the methods on the trait actually return a value which the `SeekOp` uses to decide whether the operation is invalid and should cause the change to be rejected for some reason.
- We don't require the user to remember to `enable_patches` before calling any mutating methods or remember to `disable_patches` if you are no longer interested in them, so it's harder to shoot your toes off.

Can we also make a similar modification which might obviate the `SeekOp/SeekOpWithPatch` split where instead of having two separate queries with the same logic but some additional state, we just have some optional state on the `SeekOp`?:

```rust
#[derive(Debug, Clone, PartialEq)]
pub(crate) struct SeekOp<const B: usize> {
    op: Op,
    pub pos: usize,
    pub succ: Vec<usize>,
    found: bool,
    value_tracker: Option<ValueTracker>
}

struct ValueTracker {
    had_value_before: bool,
    values: Vec<Op>,
}
```

Then in the various places where we use that state we could switch on the `Some(ValueTracker)`. (I wonder if it would make sense to pass in an `&mut Vec<Op>` for the `values` which can be reused between operations to avoid allocating a vec for each operation). BTW I _love_ all the comments in the `SeekOpWithPatch` implementation.

ept:
Hi @alexjg, thank you for your detailed comments! Good call on the code duplication — I will try to tidy it up.

@orionz and I also discussed whether to do a callback. I'd be happy with a callback-based API too; here are a few more trade-offs to consider:

* One future extension of the patch format might be to turn consecutive insertions of list elements/text characters (like pasting a big chunk of text) into a single multi-insertion, rather than a separate patch object per character, and similarly for consecutive deletions. If we have a vector of patches, it's easy to check whether the latest operation is an extension of the previous patch, and to update it. In a callback-based API, the responsibility for merging consecutive operations would presumably be moved into the callback? Alternatively, we'd need to maintain some additional state in the document somewhere that would allow us to merge consecutive ops before invoking the callback.
* If an error occurs while applying a change (e.g. something in the change is malformed), some callbacks may have already fired, so their effect may need to be rolled back in order to get into a consistent state again. With a vector, the patches aren't exposed to the application until the change has been fully applied, so a rollback is just a matter of discarding the vector.
* If we also want to expose the callback-based API via the C API and Wasm, we'd need to figure out how to do that, since currently I don't think any of the existing APIs use callbacks.

Those might be reasonable trade-offs to accept, but I just wanted to put them up for discussion as well.

alexjg:
Interesting. Can we avoid some of the impacts of the tradeoffs you describe by including an implementation of `OpObserver` for `&mut Vec<Patch>` as part of `automerge-rs`? This is basically the same as passing `&mut Vec<Patch>` to `apply_change` to add patches to rather than adding the patches to an internal list of patches which is later retrieved. I.e. the API would become

```rust
let doc: automerge::Automerge = ...
let change: automerge::Change = ...
let mut patches = Vec::new();
match doc.apply_change(change, Some(&mut patches)) {
    Ok(_) => { 
        // do something with the patchs 
    },
    Err(e) => {
        // discard the patches
    }
}
```

* This doesn't directly address the patch coalescing issue, but it does mean that for users who aren't worried about the performance implications they can just use the included implementation and not think about it. For users who are concerned about performance this would be a reference implementation they can read to understand the intended semantics.
* I think error handling is still possible with some fiddling. The user has to ensure that they discard any changes collected by an `OpObserver` during an operation which threw an error. This does feel like something which could be a potential footgun.
* The C and WASM APIs would use the `&mut Vec<Patch>` implementation for now, although I suppose the WASM API could have it's own implementation which directly creates the JS objects rather than going through a Vec at some point.

This API is a bit clunky, but my personal view is that `automerge-rs` is most useful as a low level library where we implement the complicated and performance critical parts of automerge and more opinionated and ergonomic layers can be built on top; so I tend to err on the side of flexibility over usability if forced to choose.

ept:
Yeah I think that would work. I'll have a go at implementing it and will come back to you :)

orionz:
@ept - I have a PR waiting that fixes the Value type so we can get rid of those `(Value,ExId)` tuples everywhere.  I had to break compatibility with the js implementation and now `Automerge.getConflicts()` returns an array of values without the opids per conflict.  Anyhow - be aware and hopefully this will make your code much simpler if it lands.

alexjg:
> @alexjg - the way Martin is doing it there's a flag marking weather or not we want patches so the impact on people who don't want them is a single if statement checking to see if the flag is on in `apply_changes()`

Yeah that works for users who want a simple in/out for the `Vec<Patch>` based workflow. But it seems to me that it's quite likely users will want more control over the performance/patch complexity tradeoff. I already have usecases where I want to do things like "tell me if there is an operation for this object ID", using the `Vec<Patch>` is a performance hit when I don't need most of what it's doing. I think the structure I've sketched above allows us to have the best of both worlds no? A default which just works for people who don't mind about performance or for whom `Vec<Patch>` is a good fit, but making it a trait means users who have more specific requirements can do what they need to do. But maybe I've missed something important about it? I'm curious as to what you think of it.

> https://github.com/automerge/automerge-rs/pull/312/files#diff-60370539c096503497f862671d06693f8480a82c645f5b57eb4fe816ce8c2fceL422-R516
> 
> All the new code happens only after that if is satisfied. As for a single seek op implementation, as we were writing it the seek op with patches code substantially different. For instance after finding the place to insert the op, SeekOp is done, where as SeekOpWithPatches needs to seek to the next element to collect information on conflicting values. Given that these seek functions are the most performance critical of all I'd rather have two fully different functions then one with a many if branches. There could be ample opportunity to roll up many of the simple tasks repeated in many of the different queries into helper functions for sure.

Yeah I missed that the `SeekOpWithPatches` does more work so yeah, maybe it's just a question of splitting out helper methods then.

orionz:
I like the idea of what kind of patch information you're getting being more configurable.  But this feels like a good first pass - perhaps we can mark the api is in development - and if someone can spec out a more flexible api we can evolve it.

orionz:
I see a lot of ways to improve this feature but it wont break anything to merge it and it might be a few weeks before @ept can look at it again and if we dont merge it it'll fall too far behind.  I'm going to merge it and we can treat it as a feature under development.

jeromegn:
I noticed patches didn't "trigger" when using transactions.

I made the following changes, I'm not sure if they're "good" though:

```rust
impl TransactionInner {
    /// Commit the operations performed in this transaction, returning the hashes corresponding to
    /// the new heads.
    pub fn commit(
        mut self,
        doc: &mut Automerge,
        message: Option<String>,
        time: Option<i64>,
    ) -> ChangeHash {
        if message.is_some() {
            self.message = message;
        }

        if let Some(t) = time {
            self.time = t;
        }

        let change = export_change(self, &doc.ops.m.actors, &doc.ops.m.props);
        let hash = change.hash;
        // CHANGED: call `apply_change` here instead of manually updating the history
        doc.apply_change(change);
        debug_assert_eq!(doc.get_heads(), vec![hash]);
        hash
    }
}
```

jeromegn:
Indeed. I ended up with a bad change head. I fixed it in my PR: #331

PR #313, opened 2022-03-22
return touched objects from apply_changes

orionz:
Upwelling was asking for the ability to know what objects changed when doing an applyChanges(), receiveSyncMessage() or merge() and I thought that just returning those objects was a great idea and just generally more useful.  To underscore how little we use the current return values no tests broke when I made this change 😆 

Would love feedback.

### DIFF
--- a/automerge-wasm/index.d.ts
+++ b/automerge-wasm/index.d.ts
@@ -104,7 +104,7 @@ export class Automerge {
 
   // transactions
   commit(message?: string, time?: number): Heads;
-  merge(other: Automerge): Heads;
+  merge(other: Automerge): ObjID[];
   getActorId(): Actor;
   pendingOps(): number;
   rollback(): number;
@@ -112,14 +112,14 @@ export class Automerge {
   // save and load to local store
   save(): Uint8Array;
   saveIncremental(): Uint8Array;
-  loadIncremental(data: Uint8Array): number;
+  loadIncremental(data: Uint8Array): ObjID[];
 
   // sync over network
-  receiveSyncMessage(state: SyncState, message: SyncMessage): void;
+  receiveSyncMessage(state: SyncState, message: SyncMessage): ObjID[];
   generateSyncMessage(state: SyncState): SyncMessage | null;
 
   // low level change functions
-  applyChanges(changes: Change[]): void;
+  applyChanges(changes: Change[]): ObjID[];
   getChanges(have_deps: Heads): Change[];
   getChangesAdded(other: Automerge): Change[];
   getHeads(): Heads;
--- a/automerge-wasm/src/lib.rs
+++ b/automerge-wasm/src/lib.rs
@@ -89,12 +89,9 @@ impl Automerge {
     }
 
     pub fn merge(&mut self, other: &mut Automerge) -> Result<Array, JsValue> {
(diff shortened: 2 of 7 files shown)


orionz:
There is some overlap for sure.  The patch interface is quite computationally expensive and gives you 100% of the data you need to rebuild state in userspace.  Other than the patch all we have is `heads` which just tells you "something changed" and requires you to re-walk the whole document without really knowing which things are different.  This is a nice middle ground so you know what changed and what objects to walk as a result, but is very cheap to generate since we just need to grab the `obj` field off of the ops passing through.

ept:
To me this seems quite arbitrary. Are there really many apps that will want to know the object IDs that changed, but not the keys within those objects that changed? Is this really an API that we would want to commit to long-term? Maybe, like @alexjg suggested on #312, we could make the patch interface such that you can cheaply get only the objIds if that's all you're interested in, while also supporting full patches for apps that want them.

okdistribute:
I am not sure I can speak much about the patch interface -- I've never used it. 

The objectId change has been useful so far. I set up a React component that is listening to the object id, and is responsible for rendering it. It's quite cheap to re-render a slice of the object that is in view, since only so many items can be visible on a screen at once. Many renderers nowadays also handle property changes so that any objects with the same key/id (that haven't changed) aren't re-rendered. So as long as the component can listen to when the object changed, it can re-render the visible parts of that object relatively quickly since retrieval of the relevant pieces of that object (e.g., range of a list for a virtualized list view or visible ids in a map) is fast.

okdistribute:
Furthermore, the current return behavior from the `apply_changes` function is the heads, which is not that useful since I can get these easily if I need them with `getHeads()`, so I believe `apply_changes` should return something more useful than heads.

alexjg:
I can certainly see that this is useful - I have many of the same requirements in applications I am building. My concern with this PR is more that I think we can handle things in a more general manner with the approach outlined in #312. This specific feature could then be supported at a higher level - for example it could be implemented in `automerge-wasm`. My concern is that I see the core `automerge` crate as a low level building block so I think we should try and be as flexible as possible at that level.

okdistribute:
I can't tell, does #312 return something from `apply_changes` as well? Are they conflicting PRs?

ept:
#312 doesn't return anything directly when applying changes (the patches are currently made available through a separate function, and we're discussing turning it into a callback), so there's no conflict API-wise. If several people here feel that the affected objIds are the best thing to return here, I'm happy to go along with it.

alexjg:
True, it's not conflicting, I'm not really against merging it then, although I personally find the API strange. The only other question I have is does allocating a `Vec` on every change not have performance implications?

okdistribute:
Yeah, I am hearing that there is a speculation that this may not be something that people want to commit to maintaining. Alternatively, `apply_changes` could return `void` -- right now it is returning `heads` which can be retrieved in constant time outside of `apply_changes`, so the return value currently isn't useful. Is there something else more useful than `opIds` that could be returned from `apply_changes` that isn't accessible after the function call and directly pertains to the applied changes? Would you like to return a list of patches or something instead? Just trying to understand the issue a bit deeper here to not shoot our future selves in the foot!

PR #315, opened 2022-03-22
fix the return type of values

orionz:
The return type of value() and values() has long been an ugly tuple.  With the new `set()` vs `set_object()` the only thing standing in the way of fixing this was the js `getConflicts()` needing the opid of each conflict.  I decided to break convention here and make the js implementation return a list of values instead of a map with opids.  Now the Value type can just contain all the information it needs in one place.

### DIFF
--- a/automerge-cli/src/export.rs
+++ b/automerge-cli/src/export.rs
@@ -7,15 +7,15 @@ pub(crate) fn map_to_json(doc: &am::Automerge, obj: &am::ObjId) -> serde_json::V
     for k in keys {
         let val = doc.value(obj, &k);
         match val {
-            Ok(Some((am::Value::Object(o), exid)))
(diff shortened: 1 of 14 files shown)


ept:
With this change, apps will have no way of telling which user made which of the conflicted changes, which seems like it might be useful in some apps (e.g. for an attribution UI that says "Alice changed the title to X, Bob changed the title to Y"). Are you really sure it's a good idea to remove this information?

orionz:
This is a good point - as it is you could only get info on who changed what when you're in a conflicted state and there is no information when not conflicted.   Maybe I could add a low level api call to give this information per object/property.  

Im going to try adding a new api call which will allow the old `getConflicts()` tests to pass but also fixes the Value() type problem... more to come...

PR #322, opened 2022-03-30
Add an insert buffer for transactions

jeffa5:
This cuts the number of searches we need to do for insert positions drastically.

Takes edit-trace applying the ops from ~650ms to ~370ms.

I think we'd be able to do more but this is a small first step.

### DIFF
--- a/automerge/src/transaction/inner.rs
+++ b/automerge/src/transaction/inner.rs
@@ -20,6 +20,45 @@ pub(crate) struct InsertBuffer {
     last_id: OpId,
 }
 
+impl InsertBuffer {
+    pub fn new(id: OpId, index: usize, tree_index: usize) -> Self {
+        Self {
+            target_index: index,
+            num_actions: 1,
+            tree_index,
+            last_id: id,
+        }
+    }
+
+    pub fn push(
+        &mut self,
+        doc: &Automerge,
+        obj: &ObjId,
+        id: OpId,
+        index: usize,
+    ) -> Result<Key, AutomergeError> {
+        if self.target_index + self.num_actions == index {
+            // this is an insert into the same object and at the next index so we can group it
+            self.num_actions += 1;
+            // key is the id of the last valid insert, since these are sequential inserts we
+            // can just use the last insert's id
+            let key = Ok(Key::Seq(ElemId(self.last_id)));
+            self.last_id = id;
+            key
+        } else {
+            // this buffer is not valid for this insert so start a new one
+            let query = doc.ops.search(obj, InsertNth::new(index));
+            *self = InsertBuffer {
+                last_id: id,
+                target_index: index,
+                num_actions: 1,
+                tree_index: query.pos(),
+            };
+            query.key()
+        }
+    }
+}
+
 #[derive(Debug, Clone)]
 pub struct TransactionInner {
     pub(crate) actor: usize,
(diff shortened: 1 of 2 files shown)


automerge/src/transaction/inner.rs
```diff
@@ -184,13 +198,40 @@ impl TransactionInner {
         index: usize,
         action: OpType,
     ) -> Result<Option<OpId>, AutomergeError> {
-        let id = self.next_id();
-
-        let query = doc.ops.search(&obj, query::InsertNth::new(index));
+        let key = if let Some(buffer) = self.insert_buffer.as_mut() {
```
alexjg, line 201: What if, instead of an `Option<InputBuffer>`, the `InputBuffer` was an enum with an empty state, e.g:

```rust
enum InputBuffer{
    Empty,
    Buffer {
        /// object that these inserts are operating on
        obj: ObjId,
        /// index that the first op in this buffer wanted to insert at
        target_index: usize,
        /// number of actions this buffer has seen
        num_actions: usize,
        /// index in the optree that the first op in this buffer was inserted in
        tree_index: usize,
    }
}
```
If we do this and move the logic here into an `InputBuffer::next_op` then I think you would be able to remove the `unwrap` below and this whole chunk of code would become something like:

```rust
doc.ops
            .insert(buffer.next_op(&mut this.ops, obj, index, action));
```

orionz:
Ok!  I'm really happy to see this happening!   I've been meaning to do something like this for ages - yjs uses something like this via a 10 item LRU that it manages to short circuit lookups near previous inserts or deletes.

I was laying the groundwork to do something like this in a slightly different place - LMK if you like this and might want adjust or if not thats fine too...  here's what I was thinking ...

```rust
#[derive(Debug, Clone, PartialEq)]
pub(crate) struct OpSetInternal<const B: usize> {
    trees: HashMap<ObjId, (ObjType, OpTreeInternal<B>), FxBuildHasher>,
```

I wanted  to change this HashMap to be `HashMap<ObjId, ObjectOpTreeData>` where this would be an enum that could contain the OpTreeInternal, the OpType, but also the Input Buffer stuff - this keeps the input buffer local to the object in question, and avoids on extra has lookup on each op since the tree and the cache are kept together.

Then here 

```rust
    pub fn search<Q>(&self, obj: &ObjId, query: Q) -> Q
    where
        Q: TreeQuery<B>,
    {
        if let Some((_typ, tree)) = self.trees.get(obj) {
            tree.search(query, &self.m)
        } else {
            query
        }
    }
```

this function becomes more like 

```rust
    pub fn search<Q>(&self, obj: &ObjId, query: Q) -> Q
    where
        Q: TreeQuery<B>,
    {
        if let Some(obj_op_tree_data) = self.trees.get(obj) {
            if !obj_op_tree_data.cache_lookup(&mut query) {
                let query = tree.search(&mut query, &self.m);
                obj_op_tree_data.cahce_update(&mut query)
           }
        }
        query
    }
```

This way the caching is transparent to the caller - and we can put all of our caching logic for maps or text or lists in one place.  Those methods compile down to no ops for ObjOpTreeData that dont support caching.    The downside to doing this way is that the cahce isn't stored with the transaction and then isn't blow away when the transaction closes - rather we'd need the `cache_lookup` method to blow the cache when specific queries are passed in.

Lastly if we get could this working for insert Nth and delete Nth (which is almost all of the edit trance I bet we'd get another HUGE speedup)

Exciting! =D

jeffa5:
Ooh, yeah that looks nice. I thought about doing it but went the easier way instead to see how effective it would be.

I think it makes sense but my main question is what do the caches look like? Do you think there is a general structure that would suit all queries or should we make it dynamic for the queryer? If we go the per-queryer caches route, when do the things in the cache get invalidated (e.g. caching an insert and then a delete happens may invalidate the insert's cache).

I think having them not be on the transaction will actually be better as they can be cached across transactions as well as during things like `apply_changes` from loading.

Did you have anything in a branch yet we could look at?

orionz:
I dont have any code written yet - the code I quoted before I wrote to try and create a good place to insert it when the time came.

The query caching is going to be different for maps vs lists.  So there might be some kind of 

```rust
enum ObjOpSetData {
     Map( ObjType, OpTreeInternal, MapOpSetCache )
     Seq ( ObjType, OpTreeInternal, SeqOpSetCache )
}
```

Or something similar. 

The data wouldn't be per query.  I was assuming we'd do an LRU cache for seq - that might look like. `Vec<(ListPos,OpTreePos)>` (both usize).  And when we do an insert we update it and add add +1 to all the positions larger (and -1 in the case of a del).  We reorder it so the newest goes on top and the oldest goes on the bottom.  This way Nth, InsertNth, del, set, etc, operate on the same cache.  The only tricky one is when we insert an op from an apply changes .  If its not an insert we need to know if it has a concurrent value so we could either figure that out or blow the cache (which could be fine too).

jeffa5:
I've refactored some of my stuff into that style: https://github.com/automerge/automerge-rs/pull/325 feel free to push to it as I've only done insert things!

PR #326, opened 2022-04-01
Add range and values queries

jeffa5:
Range provides an efficient way of getting just some values in the range.

Values provides a handy way of iterating through all things in the object.

This also renames `value` and `values` functions to `get` and `get_conflicts` respectively (along with their `_at` counterparts). I think this also helps to make the intent of the functions clearer.


alexjg:
I like `put` and `get` but agree on the multivalue naming stuff. I think my preference would be to flip it. `get` returns all the values and `get_winner` returns a single value. `winner` being intended to evoke things like "last write wins" as a hint that really there may be may values.

jeffa5:
Nice, I agree that the `conflicts` wording should probably be avoided. I've gone with a hybrid: `get_winner` and `get_all`. This also leaves the `get` name free should we want to effectively alias or rename one of them later.

orionz:
I suspect that 99% of users really want `get_winner` vs `get_all` and I generally like it when the "golden path" of the api has the simplest most terse api names.  I suggests to a new user whos just learning what to use, and for someone willing to flip through multiple values - they can deal with 4 extra characters in the method name.   So my vote would still be `get` and `get_all`

ept:
Don't want to turn this into bikeshedding, but if you want my opinion, I'm with @orionz and would also prefer `get` and `get_all`. My concern with `get_winner` is that I feel it introduces unnecessary cognitive overhead for someone who doesn't know or care about CRDTs (winning what? who is the loser?).

jeffa5:
That's fine, changed the winner variants to be plain `get` :)

alexjg (approved):

PR #327, opened 2022-04-01
Rename `del` and `inc` to `delete` and `increment`

jeffa5:
Rename `del` and `inc` to `delete` and `increment`.

I've also changed them in js/wasm but if that causes a conflict with js internals then we can revert the delete one.

### DIFF
--- a/automerge-cli/src/import.rs
+++ b/automerge-cli/src/import.rs
@@ -22,31 +22,31 @@ fn import_map(
     for (key, value) in map {
         match value {
             serde_json::Value::Null => {
-                doc.set(obj, key, ())?;
+                doc.put(obj, key, ())?;
             }
             serde_json::Value::Bool(b) => {
-                doc.set(obj, key, *b)?;
+                doc.put(obj, key, *b)?;
             }
             serde_json::Value::String(s) => {
-                doc.set(obj, key, s.as_ref())?;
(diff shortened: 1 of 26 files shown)


ept:
I like this. I think `delete` is a keyword in JS, but AFAIK all JS engines except for Internet Explorer 8 and older nevertheless allow `delete` as a function name. (This information is based on the fact that [Immutable.js has a function called `delete`](https://immutable-js.com/docs/v4.0.0/Map/#delete()), which is documented to be unavailable only in IE8.) Given that IE8 doesn't support Wasm either, I think we can safely use `delete`.

Since we're renaming operations, another one I've considered is whether we should rename `set` to `assign`, because the word "set" is ambiguous: as a verb it refers to updating something, whereas as a noun it refers to a collection ("the set of objects"). What do folks think of that?

jeffa5:
Interesting, `delete` may be ok then.

I agree `set` could be ambiguous, I'll throw in `put` as another option which, in my mind at least, is commonly used for overwriting things if they exist.

ept:
I'd be fine with `put` as well.

alexjg:
I like this! I'm also fine with `put`.

alexjg (approved):

alexjg (approved):

PR #329, opened 2022-04-04
Have queries be able to return references to scalars

jeffa5:
Lots of the queries were cloning operations that they were going to return. This allows them to keep just the reference to the op that exists in the tree.

Additionally, rather than returning owned values of scalars that live in the ops we should just return a reference to them to make calls to `value` cheaper, and future iteration methods.


orionz:
This looks good.  My only request would be this...

`ScalarValue(v.into_owned())`

Cant we just redefine ScalarValue as having a lifetime so we don't need to copy here?   Seems like we should be able to since it typically only lasts long enough to be converted to a JsValue

jeffa5:
Yep, the `Cow` helps us not clone values out of the ops but does still take quite a bit of space since it is an enum, if we can go just for `&ScalarValue` then that would cut the size of a `Value` by 3x (I think). 

Updated the wasm `ScalarValue` to take a `Cow` so we don't have to do the `to_owned` and added a `to_owned` function on a `Value`.

alexjg (approved):

PR #330, opened 2022-04-07
Remove grapheme splitting internally

jeffa5:
Currently, we do an automatic split of strings into grapheme clusters on some calls but users can still insert normal strings into text. Since grapheme clusters and the algorithm behind splitting them is able to change with different revisions of the unicode spec I think we should leave this to users, as they require.

When we need to split things (`splice_text` and from js) we can just use the chars that make up the rust string, these are unicode scalars.

### DIFF
--- a/automerge-cli/src/import.rs
+++ b/automerge-cli/src/import.rs
@@ -28,7 +28,7 @@ fn import_map(
                 doc.put(obj, key, *b)?;
             }
             serde_json::Value::String(s) => {
-                doc.put(obj, key, s.as_ref())?;
+                doc.put(obj, key, s)?;
             }
             serde_json::Value::Array(vec) => {
                 let id = doc.put_object(obj, key, am::ObjType::List)?;
@@ -68,7 +68,7 @@ fn import_list(
                 doc.insert(obj, i, *b)?;
             }
             serde_json::Value::String(s) => {
-                doc.insert(obj, i, s.as_ref())?;
+                doc.insert(obj, i, s)?;
             }
             serde_json::Value::Array(vec) => {
                 let id = doc.insert_object(obj, i, am::ObjType::List)?;
--- a/automerge-wasm/Cargo.toml
+++ b/automerge-wasm/Cargo.toml
@@ -30,7 +30,6 @@ getrandom = { version = "^0.2.2", features=["js"] }
 uuid = { version = "^0.8.2", features=["v4", "wasm-bindgen", "serde"] }
 serde-wasm-bindgen = "0.1.3"
 serde_bytes = "0.11.5"
-unicode-segmentation = "1.7.1"
 hex = "^0.4.3"
 regex = "^1.5"
 
--- a/automerge-wasm/src/interop.rs
+++ b/automerge-wasm/src/interop.rs
@@ -4,7 +4,6 @@ use automerge::{Change, ChangeHash, Prop};
 use js_sys::{Array, Object, Reflect, Uint8Array};
 use std::collections::HashSet;
 use std::fmt::Display;
-use unicode_segmentation::UnicodeSegmentation;
 use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
 
@@ -286,9 +285,9 @@ pub(crate) fn to_objtype(
         Some("text") => {
             let text = value.as_string()?;
             let text = text
-                .graphemes(true)
+                .chars()
                 .enumerate()
-                .map(|(i, ch)| (i.into(), ch.into()))
+                .map(|(i, ch)| (i.into(), ch.to_string().into()))
                 .collect();
             Some((am::ObjType::Text, text))
         }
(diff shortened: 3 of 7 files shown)


alexjg (approved):

orionz:
I'd like to get some input from @ept on this - he was the origin of the graphmemes part of the spec.   I was actually considering a patch thats kind of the opposite of this - where doing a `doc.set(text_id, 10, "hello world")` would convert into a `doc.splice_text(text_id, 10, 1, "hello world")` (if and only if) the object was a text - there by not allowing the user to insert non graphmeme text segments into the document.  I had a very confused user who did a doc set string and then got confusing results when fetching the indexes.

I want to talk to the upwelling team here since they are currently building a text editor on top of automerge and this graphmeme vs char distinction could make or break their project.  

I do concur that automerge and the app disagreeing on where graphmeme boundaries are could cause issues but im not 100% sure this is the best solution.

blaine:
Update from the upwelling team: we support moving to chars instead of graphemes. Overall, it gives more control to users of the api, and for our specific current purpose, we would need to implement character counting separately in any event as prosemirror counts in chars not graphemes.

ept:
Does removing the `unicode-segmentation` dependency have a noticeable effect on the size of the compiled wasm module?

I feel that from a principled point of view, it makes most sense to treat each grapheme cluster as an atomic unit. For example, a single emoji character may consist of a sequence of several code points, and it wouldn't make sense to insert something in the middle of that sequence, since it would break the emoji character. However, splitting it into the individual code points would still be okay, as long as editors delete all of the code points in a grapheme cluster if it is deleted, and as long as they don't insert new code points into the middle of an existing grapheme cluster.

There is another point of view, which is ease of integration with external APIs such as text editors. For example, it would be convenient for the index in the Automerge text object to correspond exactly to the index of the same character in the text editing widget. However, the problem here is that different platforms handle strings differently:

* [JavaScript indexes strings by UTF-16 code unit](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charAt), not by code point. For example, the character 🙃 (U+1F643) is a single code point, but in JavaScript it is a string of length 2, because it is encoded as a surrogate pair.
* [Swift indexes strings by grapheme cluster](https://developer.apple.com/documentation/swift/string) by default (if I understand correctly), although it also supports UTF-8 and UTF-16 views.
* [Rust strings are UTF-8](https://doc.rust-lang.org/std/string/struct.String.html).
* I can't find documentation on how Kotlin indexes strings right now.

This means that no matter how we split strings into characters, there will always be some platforms on which we need to perform some awkward translations of indexes because, for example, something that is a length-2 string in the app programming language is a single element in the Automerge array, or vice versa.

@blaine, when you say ProseMirror counts in chars, what does that mean exactly? My guess is that it counts UTF-16 code units since that's what JS does. What happens if you add emoji to an upwelling document?

blaine:
> I feel that from a principled point of view, it makes most sense to treat each grapheme cluster as an atomic unit. For example, a single emoji character may consist of a sequence of several code points, and it wouldn't make sense to insert something in the middle of that sequence, since it would break the emoji character. However, splitting it into the individual code points would still be okay, as long as editors delete all of the code points in a grapheme cluster if it is deleted, and as long as they don't insert new code points into the middle of an existing grapheme cluster.

I would argue that access to the underlying code units is probably essential in the long term; grapheme access is a convenience, because as you say, editors will/should always correctly insert/delete the correct number of code units.

Maybe a Swift-like optional view would make sense to support platform variance?

> @blaine, when you say ProseMirror counts in chars, what does that mean exactly? My guess is that it counts UTF-16 code units since that's what JS does. What happens if you add emoji to an upwelling document?

Yes, prosemirror is counting UTF-16 code units per Javascript's approach - https://discuss.prosemirror.net/t/special-handling-of-utf8-needed/3243

ept:
I'm okay with changing the API to split a string into characters such that each Automerge list element is one Unicode scalar value.

I like your suggestion of having multiple views to match the indexing used by different platforms. That way, the API could support indexing based on UTF-8, UTF-16, Unicode scalar values, or grapheme clusters, but internally we always use the same representation. Such a view feature is something we could add in the future.

alexjg (approved):

orionz:
Well id be happy to make text resolve the Char for several reasons.  One of those would be we can make a `ScalarValue::Char()` and assume all normal values in a text object are char.  What should we do if someone does a `doc.put(text_id, 100, "It was the best of times... ")`   We could (1) break it into chars and translate this into a `doc.splice(text_id, 100, 1, "It was the best of time...")` effectively splicing this in and deleting 1 char.  Or would could treat the string the same way we would a map or a binary and do object substitution if someone does `doc.text(text_id)`.  Thoughts?

Im happy to merge this - is there any will to add Char scalar value and the splice or object substitution behavior into this PR?

jeffa5:
I think for clarity, this PR does not enforce that text has to use just unicode scalars, users can still use graphemes if they like, just they will have to use `splice` rather than `splice_text`. This avoids us having to depend on a single unicode revision. Something like enforcing it could come later.

For this reason, I think automatically changing `put(textid, 100, "my long text")` into a `splice` of some sort shouldn't be done as what if someone does `put(textid, 100, "<grapheme_cluster>")` and expects that to be a single thing.

Object substitution could be interesting but I think may be something for another PR.

On a related note: since we can insert objects into texts, how is a Text different now from a List? It signals intent to the user but isn't enforced in a reasonable way. If there aren't any special limitations of Text (users can insert arbitrary things in) then should we just use a List instead?

jeffa5:
I like the idea of offering views over the text objects for various encodings/groupings but think it might still be worth keeping the ability to have a grapheme at a position in text objects unless we can really optimise it when we make some restriction.

PR #331, opened 2022-04-07
Add patches support to transactions

jeromegn:
This adds previously-unsupported patches tracking to transactions. Patches only worked for `apply_changes` in the past.

### DIFF
--- a/automerge/src/automerge.rs
+++ b/automerge/src/automerge.rs
@@ -2117,4 +2117,69 @@ mod tests {
         let len = doc.length(&text);
         assert_eq!(len, 4); // 4 chars
     }
+
+    #[test]
+    fn patches_are_populated() {
+        let mut doc = Automerge::new();
+        doc.enable_patches(true);
+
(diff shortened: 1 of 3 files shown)


jeffa5:
Thanks for this! Would you be able to add some tests to check that this works and continues to work?

I also think we should be able to get around having to query the opset again for each patch by storing the `Prop` of each operation in the `TransactionInner::operations` field and then using that to make the patches.

jeromegn:
I added a test(s)!

> I also think we should be able to get around having to query the opset again for each patch by storing the `Prop` of each operation in the `TransactionInner::operations` field and then using that to make the patches.

I've also added this (I think I did it right). I'd like to trigger the test pipeline here to make sure it's all good.

alexjg:
>  I'd like to trigger the test pipeline here to make sure it's all good.

Note that you can run all the tests that CI runs locally by running `./scripts/ci/run`

orionz:
This is superseded by Op Observer

PR #332, opened 2022-04-07
Add OpObserver to the document

jeffa5:
This generalises the patch interface to an observer trait. This allows more flexible ways of tracking operations.

Based off https://github.com/automerge/automerge-rs/pull/312#issuecomment-1073259343

It also tracks operations made during transactions as they occur, it issues `undo_*`s if it is rolled back. This may be better of as just triggering the observer in commit rather than on every individual action. Not sure if the undo behaviour would be useful anywhere else, if not I'd be inclined to remove it.

Related: https://github.com/automerge/automerge-rs/pull/331


jeffa5:
Yep, I think I agree too, the `dyn` makes it quite hard to use and get things out (`Rc<RefCell<T>>` needed).

I'm still unsure about the observing on every local operation, I think it could end up complicating some implementations but I'm happy to give it a go as I can imagine cases where users would want the observer to be updated asap

jeromegn:
For our use case, this might be harder to use.

We need to add observers _after_ the fact. The patches API works well for this because we manage our own observers. When we `pop_patches`, we can traverse the tree based on the obj id and figure out if we need to trigger a notification to a "watcher".

Furthermore, I think both the patches and observers implementation have the same bug here: https://github.com/automerge/automerge-rs/pull/332/files#diff-397994bdf8ff993666ddbc597fe755520f0c7442f1b66d8d47114ed2ed579c9aR174
^ this does not report the "inserted at" index, it seems to report the last index of the array.

alexjg:
To the first point, can you achieve the same semantics as the `pop_patches` approach byt using the `VecOpObserver`? I.e. where you would currently do `pop_patches` you just look at what's in `VecOpObserver` after calling `apply_change` or whichever local mutations you're calling?

alexjg (approved):
This looks good to me. The `NULL_OBSERVER` stuff is annoying, but I can't see a better way of doing it without compromising on flexibility. We'll just have to document it well.

Regarding whether to call the observer on commit vs on each local operation, I wonder if we should do both? That way it's up to the application whether they want to be notified of things which may need rolling back. Regardless, I think that would be future work.

jeffa5:
I suppose another option for the `NULL_OBSERVER` things is having variants such as `put_with(obj, key, value, opobserver)` which means they have to opt into it. I think that would make normal use clearer.

Ah yes, I like that we could do both, perhaps we could add it to the `CommitOptions` struct and then it works with `commit_with`?

jeromegn:
> * However, mine only added the patches when the commit was successful.

The workaround for this is probably to not do anything with the patches if the transaction returns an `Err`.

alexjg:
> * Passing around `Option<&mut Obs>` multiple times is a bit cumbersome. Essentially have to pass `&mut Option<&mut Obs>` and constantly do something like:

Interesting, can you post an example of this, I would have thought you could just pass `obs.as_mut()`

jeromegn:
image

Possibly just a matter of implementing OpObserver for all &mut OpObserver?

jeffa5:
I've updated this so that we just capture local operations on the observer during commit, this simplifies the situation of rolling back local changes, especially as transactions should be short-lived.

If a need to have it be more fine-grained arises then I suggest we tackle it then, with some examples of usage.

jeromegn:
I like the `CommitOptions` change!

Using it in actual code with a `VecObserver` though, I'm pretty sure lifetime parameters should be specified:

```rust
    pub fn transact_with<'a, F, O, E, C, Obs>(&mut self, c: C, f: F) -> transaction::Result<O, E>
    where
        F: FnOnce(&mut Transaction) -> Result<O, E>,
        C: FnOnce(&O) -> CommitOptions<'a, Obs>,
        Obs: 'a + OpObserver,
```

or else the `OpObserver` needs to be alive for a `'static` lifetime.

jeffa5:
Thanks @jeromegn! Just fixed up that lifetime issue, and made use of it in the `watch` example to clean things up a bit.

orionz:
Wow..  This is really really nice work.   Thoughts on a `merge()` and `merge_with()`, `load()`, `load_with()`.  etc... not sure if its worse to have too many method permutations or too many unused arguments

alexjg:
I'm not sure either but one argument in favor of having both `<op>` and `<op>_with()` permutations might be that the `*_with()` methods have different performance characteristics because they traverse more parts of the op tree?

jeffa5:
I'll rebase to fix the conflicts, then I think having the `*_with(*Options)` style will be cleaner.

jeffa5:
Updated to fix conflicts and add `ApplyOptions` for passing the observer, also allows future expansions.

PR #333, opened 2022-04-07
Wasm readme

orionz:
This is mostly a readme and packaging improvements for automerge-wasm.  There was some API cleanup as well.  Most notably the inclusion of a `fork_at()` method to allow a document to be forked at a point in time.

### DIFF
--- a/automerge-wasm/README.md
+++ b/automerge-wasm/README.md
@@ -22,18 +22,17 @@ This is a rust/wasm package and will work in a node or web environment.  Node is
 
 This creates a document in node.  The memory allocated is handled by wasm and isn't managed by the javascript garbage collector and thus needs to be manually freed.
 
-```
+```javascript
   import { create } from "automerge-wasm"
 
   let doc = create()
 
   doc.free()
-
 ```
 
 While this will work in both node and in a web context
 
-```
+```javascript
   import init, { create } from "automerge-wasm"
 
   init().then(_ => {
@@ -51,7 +50,7 @@ Automerge has many scalar types.  Methods like `set()` and `insert()` take an op
 
 These are sets without a datatype
 
-```
+```javascript
   import { create } from "automerge-wasm"
 
   let doc = create()
@@ -69,7 +68,7 @@ Sets with a datatype and examples of all the supported datatypes.
 
 While int vs uint vs f64 matters little in javascript, Automerge is a cross platform library where these distinctions matter.
 
-```
+```javascript
   import { create } from "automerge-wasm"
 
   let doc = create()
@@ -89,7 +88,7 @@ While int vs uint vs f64 matters little in javascript, Automerge is a cross plat
 
 Automerge WASM supports 3 object types.  Maps, lists, and text.  Maps are key value stores where the values can be any scalar type or any object type.  Lists are numerically indexed set of data that can hold any scalar or any object type.  Text is numerically indexed sets of graphmeme clusters.
 
-```
+```javascript
   import { create } from "automerge-wasm"
 
   let doc = create()
@@ -115,7 +114,7 @@ Automerge WASM supports 3 object types.  Maps, lists, and text.  Maps are key va
 
 You can access objects by passing the object id as the first parameter for a call.
 
-```
+```javascript
   import { create } from "automerge-wasm"
 
   let doc = create()
@@ -127,7 +126,7 @@ You can access objects by passing the object id as the first parameter for a cal
 
 Anywhere Object Id's are being used a path can also be used.  The following two statements are equivelent:
 
-```
+```javascript
   // get the id then use it
 
   let id = doc.value("/", "config")[1]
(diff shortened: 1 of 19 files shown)


automerge-wasm/README.md
```diff
@@ -1,696 +1,478 @@
...
+
+CRDT stands for Conflict Free Replicated Data Type.  It is a data structure that offers eventual consistency where multiple actors can write to the document independently and then these edits can be automatically merged together into a coherent document that, as much as possible, preserves the intent of the different writers.  This allows for novel masterless application design where different components need not have a central coordinating server when altering application state.
+
+### Terminology
+
+The term Actor, Object Id and Heads are used through this documentation.  Detailed explanations are in the glossary at the end of this readme.  But the most basic definition would be...
+
+An Actor is a unique id that distinguishes a single writer to a document.  It can be any hex string.
+
+An Object id uniquely identifies a Map, List or Text object within a document.  This id comes as a string in the form of `{number}@{actor}` - so `"10@aabbcc"` for example.  The string `"_root"` or `"/"` can also be used to refer to the document root.  These strings are durable and can be used on any descendant or copy of the document that generated them.
```
alexjg, line 17: do we want to talk about the structure of the object ID at all? I've been thinking of object IDs as opaque from the users perspective  and think it would be better to just say "this is an opaque string". Maybe that's naive and people will use the vector clock structure regardless of what we encourage them to do, but in that case we should probably add _more_ documentation and better API support.

automerge-wasm/README.md
```diff
@@ -1,696 +1,478 @@
...
+    doc.insertObject(items, 1, [1,2])             // insert a list with 2 values at pos 1
+
+    doc.materialize(items)                        // returns [ "bat", [1,2], { hello : "world" }, true, "bag", "brick"]
+    doc.length(items)                             // returns 6
+    doc.free()
+```
+
+### Text
+
+Text is a specialized list type intended for modifying a text document.  The primary way to interact with a text document is via the slice operation.  Non text can be inserted into a text document and will be represented with the unicode object replacement character.
```
alexjg, line 190: Should we add something here around the way strings are split in javascript vs other platforms. Feels like this is a common footgun so worth documenting.

automerge-wasm/README.md
```diff
@@ -1,696 +1,478 @@
...
+    doc1.merge(doc2)
+
+    doc1.materialize("_root")  // returns { number: 10, total: 33 }
+
+    doc1.free(); doc2.free()
+```
+
+### Transactions
+
+Generally speaking you don't need to think about transactions when using Automerge.  Normal edits queue up into an in-progress transaction.  You can query the number of ops in the current transaction with `pendingOps()`.  The transaction will commit automatically on certains calls such as `save()`, `saveIncremental()`, `fork()`, `merge()`, `getHeads()`, `applyChanges()`, `generateSyncMessage()`, and `receiveSyncMessage()`.  When the transaction commits the heads of the document change.  If you want to roll back all the in progress ops you can call `doc.rollback()`.  If you want to manually commit a transaction in progress you can call `doc.commit()` with an optional commit message and timestamp.
```
alexjg, line 262: Should we document when you _would_ want to worry about transactions?

automerge-wasm/README.md
```diff
@@ -1,696 +1,478 @@
...
+
+    doc1.materialize("_root")       // returns { key1: "val1", key2: "val2", key3: "val3" }
+    doc2.materialize("_root")       // returns { key1: "val1", key3: "val3" }
+
+    doc1.free(); doc2.free()
+```
+
+Note that calling `a.merge(a)` will produce an unrecoverable error from the wasm-bindgen layer which (as of this writing) there is no workaround for.
+
+### Saving and Loading
```
alexjg, line 341: We should document somewhere in this section that the concatenation of two automerge documents is the merge of those documents because it's just too neat to not shout about it.

automerge-wasm/README.md
```diff
@@ -1,696 +1,478 @@
...
+
+  doc1.materialize("_root")  // returns { key1: "value1", key2: "value2" }
+  doc2.materialize("_root")  // returns { key1: "value1", key2: "value2" }
+  doc3.materialize("_root")  // returns { key1: "value1", key2: "value2" }
+  doc4.materialize("_root")  // returns { key1: "value1", key2: "value2" }
+
+  doc1.free(); doc2.free(); doc3.free(); doc4.free()
+```
+
+### Syncing
```
alexjg, line 386: I think this could do with a small narrative section about the sync protocol needing a number of round trips. Especially the fact that `generateSyncMessage` returns `null` when sync is complete.

alexjg (approved):
Some small thoughts on the README. I'm fine to merge this either way, these comments were just my thoughts as I read through.

automerge-wasm/README.md
```diff
@@ -1,696 +1,478 @@
...
+  doc.put("/", "prop2", 3.14) // f64
+  doc.put("/", "prop3", "hello world")
+  doc.put("/", "prop4", new Date())
+  doc.put("/", "prop5", new Uint8Array([1,2,3]))
+  doc.put("/", "prop6", true)
+  doc.put("/", "prop7", null)
+  doc.free()
+```
+
+Sets with a data type and examples of all the supported data types.
```
jeffa5, line 69: Should this be `Puts` rather than `Sets`?

automerge-wasm/README.md
```diff
@@ -1,696 +1,478 @@
...
+  let config = doc.putObject("/", "config", { align: "left", archived: false, cycles: [10, 19, 21] })
+
+  doc.put(config, "align", "right")
+
+  // Anywhere Object Ids are being used a path can also be used.
+  // The following two statements are equivalent:
+
+  // get the id then use it
+
+  let id = doc.value("/", "config")[1]
```
jeffa5, line 134:
```suggestion
  let id = doc.value("/", "config")
```

automerge-wasm/README.md
```diff
@@ -1,696 +1,478 @@
...
+                              // make a new empty object and assign it to the key `sub` of mymap
+
+    doc.keys(mymap)           // returns ["bytes","foo","sub"]
+    doc.materialize("_root")  // returns { mymap: { bytes: new Uint8Array([1,2,3]), foo: "bar", sub: {}}}
+    doc.free()
+```
+
+### Lists
+
+Lists are index addressable sets of values.  These values can be any scalar or object type.  You can manipulate lists with `insert()`, `put()`, `insertObject()`, `pubObject()`, `push()`, `pushObject()`, `splice()`, and `del()`.
```
jeffa5, line 170: Is it `delete` now?

automerge-wasm/README.md
```diff
@@ -1,696 +1,478 @@
...
+    let doc1 = create("aaaaaa")
+    doc1.put("_root", "number", 0)
+    doc1.put("_root", "total", 0, "counter")
+
+    let doc2 = doc1.fork("bbbbbb")
+    doc2.put("_root", "number", 10)
+    doc2.inc("_root", "total", 11)
+
+    doc1.put("_root", "number", 20)
+    doc1.inc("_root", "total", 22)
```
jeffa5, line 251: `inc` is now `increment` I think

automerge-wasm/README.md
```diff
@@ -1,696 +1,478 @@
...
+  let doc6 = load(doc4.save(), "00aabb11")
+
+  let actor = doc1.getActor()
+
+  doc1.free(); doc2.free(); doc3.free(); doc4.free(); doc5.free(); doc6.free()
+```
+
+### Glossary: Object Id's
+
+Object Ids uniquely identify an object within a document.  They are represented as strings in the format of `{counter}@{actor}`.  The root object is a special case and can be referred to as `_root`.  The counter is an ever increasing integer, starting at 1, that is always one higher than the highest counter seen in the document thus far.  Object Id's do not change when the object is modified but they do if it is overwritten with a new object.
```
jeffa5, line 457: Perhaps worth noting that using the old object id still accesses the old object though it is no longer visible in the document?

automerge-wasm/index.d.ts
```diff
@@ -109,10 +109,10 @@ export class Automerge {
   keys(obj: ObjID, heads?: Heads): string[];
   text(obj: ObjID, heads?: Heads): string;
   length(obj: ObjID, heads?: Heads): number;
-  materialize(obj?: ObjID): any;
+  materialize(obj?: ObjID, heas?: Heads): any;
```
jeffa5, line 112:
```suggestion
  materialize(obj?: ObjID, heads?: Heads): any;
```

automerge/src/automerge.rs
```diff
@@ -198,6 +198,31 @@ impl Automerge {
         f
     }
 
+    /// Fork this document at the give heads
```
jeffa5, line 201:
```suggestion
    /// Fork this document at the given heads
```

jeffa5 (commented):
Looks great! Just a few comments.

automerge-wasm/README.md
orionz: This one is hard for me because I can't think of many real world uses for an explicit commit.

alexjg: The only thing I can think of is if you wanted to sign a change and ship it off somewhere. But that's really complicated to explain.

PR #341, opened 2022-04-18
Don't queue or apply a change more than once

### LINKED ISSUE #337: Duplicate sync messages issues in `experiment` branch
**Context**

I've been using automerge's `experiment` branch in my project. It's a bit different in that it doesn't synchronize with a centralized server, instead it synchronizes with N peers which are expected to end up with the same state eventually (weakly-consistent).

Effectively, I'm sending `Change`s (compressed) instead of using the sync protocol. This is because given our "mesh" graph, it would be very intensive to do run the sync protocol loop with every nodes every time there's a change.

Sending `Change`s-only actually works quite well. I was hoping to use the sync protocol on an interval as an extra guarantee of weak consistency.

I'm using the sync protocol on startup, to get up to speed fast.

I'm using `automerge-persistent` w/ a [slightly modified `experiment` branch](https://github.com/jeromegn/automerge-persistent/tree/experiment-observer) to support the latest changes in `automerge` and the `sled` store.

**Problem**

After a few updates, I start getting the same non-empty sync messages over and over. That is, the node generating a sync protocol message for another node keeps sending the same data.

This creates an infinite loop because `generate_sync_message` always produces `Some(message)` and `receive_sync_message` always works too, so there's no way of breaking the loop.

Here's how I'm using the sync protocol:

```
A
|--> generate_sync_message( B.id )
    |--> if Some(message)
        |--> http/2 request with the encoded sync message
    |--> if None
        |--> break loop

(may receive a `Change` and apply it)

B
|--> receive_sync_message( A.id, message )
|--> generate_sync_message( A.id )
    |--> if Some(message)
        |--> http/2 response with encoded sync message
    |--> if None
        |--> http/2 response with "done" message

(may receive a `Change` and apply it)

A
|--> if "done"
    |--> break loop
|--> else
    |--> receive_sync_message( B.id, message )
    |--> continue loop

(may receive a `Change` and apply it)
```

Now, it's worth noting that `Change` message change happen between sending and receiving a sync message. I've been told this did not matter.

What am I doing wrong?

I've worked around it by breaking the loop if I detect I've received the same message twice in a row from the same peer.

I've also turned off the sync protocol interval, so my nodes don't ever sync. All they do is exchange `Change`s. **Surprisingly**, this results in a better consistency in the end. That is to say: if I use the sync protocol at all, in addition to my `Change` propagation, the state becomes desynchronized. I assume because I'm using the sync protocol wrong.

jeromegn:
Fixes #337 

This checks that a `Change` hasn't already been queued in addition to checking that it's not in the history index.

Without this, a few bad things can happen that will corrupt a document's history:
- A change can be queued twice if it isn't causally ready twice
- A change might be applied if it's causally ready, even if the same change had been previously queued

Since the queue is looped through after applying changes, then the 2 aforementioned situation can lead to a change being applied twice, bypassing the duplicate check.

### DIFF
--- a/automerge/src/automerge.rs
+++ b/automerge/src/automerge.rs
@@ -657,13 +657,17 @@ impl Automerge {
         dup
     }
 
+    fn history_indexed_or_queued(&self, hash: &ChangeHash) -> bool {
+        self.history_index.contains_key(hash) || self.queue.iter().any(|c| hash == &c.hash)
+    }
+
     /// Apply changes to this document.
     pub fn apply_changes(
         &mut self,
         changes: impl IntoIterator<Item = Change>,
     ) -> Result<(), AutomergeError> {
         for c in changes {
-            if !self.history_index.contains_key(&c.hash) {
+            if !self.history_indexed_or_queued(&c.hash) {
                 if self.duplicate_seq(&c) {
                     return Err(AutomergeError::DuplicateSeqNumber(
                         c.seq,


orionz:
Thank you for this.  Looks like we duplicated each others work here.

https://github.com/automerge/automerge-rs/pull/342

PR #342, opened 2022-04-18
duplicate changes in the queue could corrupt internal state

orionz:
If a change got out of order queued multiple times it could be applied multiple times and corrupt internal state.  Added a test to expose the problem and a fix to fix it.  Code to prevent the problem at queue push time instead of pop time could be more memory efficient but as this is an unusual edge case I figured the simple fix would be good enough for now.

### DIFF
--- a/automerge/src/automerge.rs
+++ b/automerge/src/automerge.rs
@@ -678,7 +678,9 @@ impl Automerge {
             }
         }
         while let Some(c) = self.pop_next_causally_ready_change() {
-            self.apply_change(c);
+            if !self.history_index.contains_key(&c.hash) {
+                self.apply_change(c);
+            }
         }
         Ok(())
     }
--- a/automerge/tests/test.rs
+++ b/automerge/tests/test.rs
@@ -831,10 +831,10 @@ fn handle_repeated_out_of_order_changes() -> Result<(), automerge::AutomergeErro
     doc1.commit();
     doc1.insert(&list, 3, "d")?;
     doc1.commit();
-    let changes = doc1.get_changes(&[]);
-    doc2.apply_changes(changes[2..].iter().cloned().cloned().collect())?;
-    doc2.apply_changes(changes[2..].iter().cloned().cloned().collect())?;
-    doc2.apply_changes(changes.iter().cloned().cloned().collect())?;
+    let changes = doc1.get_changes(&[]).into_iter().cloned().collect::<Vec<_>>();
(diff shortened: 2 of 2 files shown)


jeffa5 (approved):

orionz:
Im a little bothered this can put the same change multiple times into the queue and eat memory.  

Option 1: don't worry about it - it's a rare edge case.

Option 2: check before pushing in the queue

`   } else if !self.queue.iter().any(|q| q.hash == c.hash) {`

This adds a linear search term any time a change arrives out of order - I dont like it. 

Option 3: change queue from a `Vec<Change>` to a `HashMap<ChangeHash,Change>`. This means that changes can only be added once, it also means `pop_next_causally_ready_change()` is essentially walking the queue at random which would matter in some cases (but no others).  Id rather not overengineer and make queue be a dep tree but we could (popping here based on change.deps after applying).  We'd also need make sure a change applied (that has a copy in queue) doesn't get double applied.  This could be a `queue.remove(c.hash)` with each `apply_change()` which is a small by steady tax on the speed of each apply change even when nothing ever enters the queue

jeromegn:
Sounds like regardless of the solution, queue length can become a problem.

Possibly adding a configurable limit. When reached, this would trigger an error "changes queue full" or something.

It sounds like an error that can be acted upon. Probably by triggering a sync or a full load from another peer.

Thinking about it, inspecting the queue length during program execution might also be useful to determine when to sync!

When I did my PR, I was looking for a crate that provided a HashSet-like structure that kept its order. Such things do exist, if we're not averse to adding dependencies.

jeromegn:
I also wondered if it would make sense for "generate sync message" to consider items in the queue as "haves". But this is way over my head.

jeffa5:
@orionz I think we should merge this as is, its a nice and self-contained minimal fix. I agree that we we should limit duplicates coming into the queue too but think it could be a separate PR. I'd probably go the `HashMap` route, which we can optimise if needed with an indexmap or a more sophisticated graph.

This does remind me of conversations had here (https://github.com/automerge/automerge/issues/400) about the difficulty with the queue and whether we should make the queue external to this automerge library (could always add a version in the wasm lib), then @jeromegn can manage things like the limit externally.

alexjg:
I agree that we should merge this as is, it's strictly an improvement. I definitely think we should split the queue out though and just bundle it in the `AutoCommit` at some point though.

alexjg (approved):

PR #346, opened 2022-04-20
Prevent increment on non-counter

### LINKED ISSUE #344: Local operations allow increment on non-counters
Currently, we can call `increment` on a non-counter value without getting an error back. I don't think this should be allowed as it doesn't make logical sense and could lead to some confusing behaviour.

jeffa5:
Fixes #344

### DIFF
--- a/automerge/src/decoding.rs
+++ b/automerge/src/decoding.rs
@@ -52,7 +52,60 @@ pub enum Error {
     Io(#[from] io::Error),
 }
 
-#[derive(thiserror::Error, Debug)]
+impl PartialEq<Error> for Error {
+    fn eq(&self, other: &Error) -> bool {
+        match (self, other) {
+            (
+                Self::WrongType {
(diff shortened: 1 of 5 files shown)


automerge/src/decoding.rs
```diff
@@ -52,7 +52,60 @@ pub enum Error {
     Io(#[from] io::Error),
 }
 
-#[derive(thiserror::Error, Debug)]
+impl PartialEq<Error> for Error {
```
alexjg, line 55: What's the reason for having to write this out by hand?

jeromegn:
Ah! I just faced this issue yesterday when I realized a "number" wasn't necessarily a counter...

automerge/src/decoding.rs
jeffa5: The leb and io errors don't implement PartialEq unfortunately

PR #350, opened 2022-04-22
Optimise prop query

### LINKED ISSUE #349: Slow counter performance in map
When creating a counter in a map we have to do a sequential scan through the ops, effectively making it `O(n)` per increment operation, nasty!

I think we should be able to do this if we change how prop works a bit: using the last item in each node for a key comparison and then using that to decide whether to descend or skip that node.

jeffa5:
Results from critcmp using the new benchmarks:
```
group                       main                                    opt-prop
-----                       ----                                    --------
decreasing put 1000         1.00  1267.9±80.15µs        ? ?/sec     1.13  1438.7±39.22µs        ? ?/sec
decreasing put 10000        1.00     16.0±1.14ms        ? ?/sec     1.24     19.9±1.24ms        ? ?/sec
increasing put 1000         1.00  1460.2±103.95µs        ? ?/sec    1.20  1750.3±16.75µs        ? ?/sec
increasing put 10000        1.00     19.6±1.08ms        ? ?/sec     1.28     25.1±1.18ms        ? ?/sec
repeated increment 1000     9.17     11.7±1.14ms        ? ?/sec     1.00  1277.1±14.38µs        ? ?/sec
repeated increment 10000    98.99 1708.5±55.31ms        ? ?/sec     1.00     17.3±0.56ms        ? ?/sec
repeated put 1000           7.30     11.7±1.03ms        ? ?/sec     1.00  1605.7±43.10µs        ? ?/sec
repeated put 10000          79.08 1655.9±11.38ms        ? ?/sec     1.00     20.9±0.53ms        ? ?/sec
```

When there is a unique op for the key the old version is faster (currently, I have ideas to improve it later). But, when there are multiple updates to a key (`repeated *`) there is a massive speedup, basically by skipping non-visible ops and tracking visibility in the index, even for maps.

Fixes #349

### DIFF
--- a/automerge/Cargo.toml
+++ b/automerge/Cargo.toml
@@ -43,3 +43,12 @@ proptest = { version = "^1.0.0", default-features = false, features = ["std"] }
 serde_json = { version = "^1.0.73", features=["float_roundtrip"], default-features=true }
 maplit = { version = "^1.0" }
 decorum = "0.3.1"
+criterion = "0.3.5"
+
+[[bench]]
+name = "range"
+harness = false
+
+[[bench]]
+name = "map"
+harness = false
--- a/automerge/benches/map.rs
+++ b/automerge/benches/map.rs
@@ -1,0 +1,84 @@
+use automerge::{transaction::Transactable, Automerge, ScalarValue, ROOT};
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+
+fn repeated_increment(n: u64) -> Automerge {
+    let mut doc = Automerge::new();
+    let mut tx = doc.transaction();
+    tx.put(ROOT, "counter", ScalarValue::counter(0)).unwrap();
+    for _ in 0..n {
+        tx.increment(ROOT, "counter", 1).unwrap();
+    }
+    tx.commit();
+    doc
+}
+
+fn repeated_put(n: u64) -> Automerge {
+    let mut doc = Automerge::new();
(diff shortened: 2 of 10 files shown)


orionz:
Seems like this has a 25% speed penalty for normal puts and sets - is there any way to fix this without that baseline cost?
What if we kept `visible: Map<Key,Vec<Value>>` in the index?  Couldn't we dramatically reduce the search code for all query operations in non `_at` functions then?  Somewhat similar to the `OpRange<>` idea I was kicking around but would be super simple to implement and should give similar performance results..  All the nth queries would just be about converting pos to elemid and then looking in the index for the answer.

orionz:
Also I cant merge until the conflicts get cleared =)

jeffa5:
Conflicts fixed, I've also added a skip method to the searching which should bring back the performance for singles

jeffa5:
> Seems like this has a 25% speed penalty for normal puts and sets - is there any way to fix this without that baseline cost? What if we kept `visible: Map<Key,Vec<Value>>` in the index? Couldn't we dramatically reduce the search code for all query operations in non `_at` functions then? Somewhat similar to the `OpRange<>` idea I was kicking around but would be super simple to implement and should give similar performance results.. All the nth queries would just be about converting pos to elemid and then looking in the index for the answer.

Yeah, I think this could also work in the long run.

PR #355, opened 2022-04-28
Storage v2

alexjg:
This is a draft of the rewrite of the storage subsystem I've been working on. All the tests are now passing and there is a 40% improvement in the load benchmark. Unfortunately there are some performance regressions, mostly in the save benchmark but also a small regression in the replay benchmark. I'm confident these can be addressed as I've done no optimisation yet.

I'm posting this PR for people who are curious. This is a lot of code and I am planning on adding a lot of documentation and splitting the patch into a smaller series of commits to make it easier to review. I'm also happy to synchronously walk through the code with anyone who is interested.

Due to this being a very large change I have implemented this behind a feature flag, which should make it easier to experiment with without comitting us to a giant new change immediately. To use it enable the `automerge/storage-v2` or `automerge-wasm/storage-v2` feature flags.

The top level objective of this PR is the following:

- Improve the performance of the load benchmark by streaming operations directly into the OpTrees as they are read from the document, and by reconstructing changes without an extra encode/decode step to get the change graph. The entry point for this is the `Automerge::load_with` method and the `automerge::storage::load::reconstruct_document` module
- Make the parsing logic more robust in the face of incorrect encodings (this mostly means introducing a new `Decodable` trait which returns `Result<T, ..>`  rather than just `T`.
- More completely implement the columnar storage format. This includes types which allows to iterate over columns we don't expect, which leads on to the next point:
- Lay a basic set of types (those in `columnar_2` and particularly in `columnar_2::column_range::generic`) which give us an API for more easily representing arbitrary columns at runtime. Using the columnar representations at runtime is the next step once this work is merged. This will hopefully give us performance improvements but more importantly (IMO) will mean that we can operate on documents with unknown columns.
- Making compression of document chunks optional (via `Automerge::save_nocompress`)

Still on my todo list:

- Documentation
- Fixing performance regressions
- Improve the API of `DecodeColumnError`
- Improve the `storage::parse` module:
    - Make the error API easier to use
    - Add a wrapper around the input type which records the position in the
      input slice so that it's easier to write parsers which produce
      `Range<usize>`
- Really a lot more documentation


ept:
I got a code walk-through from @alexjg and I am excited about this change. I'm too much of a Rust beginner to comment on the actual code, but there are several things I like:

* Adding support for preserving unknown columns, which is an important part of the forward-and-backward-compatibility story for Automerge docs, and which is one thing (perhaps the last thing?) that is currently keeping automerge-rs from feature parity with the JS implementation
* Laying the groundwork for using columnar encoded data in the OpTree, which has the potential to make it a lot more memory-efficient
* Faster document loading

BTW Alex, if you need a test case for a change containing unknown columns, you can find one [here](https://github.com/automerge/automerge/blob/main/test/new_backend_test.js#L1857-L1905).

alexjg:
Alright folks, this PR is ready for review. I've tried to split it up into hopefully discrete enough commits. Particular things I would like review of:

* @jkankiewicz I've made a few changes to `automerge-c` to accomodate a new `ScalarValue::Unknown` variant. Hopefully what I've done is sensible but would be great to get your eyes on it. I've also not made any attempt to run the tests for `automerge-c` with a `storage-v2` feature flag enabled.
* @jeromegn I've introduced an `Automerge::save_nocompress` method. I recall that compression was costing you quite a lot - if that's still the case I would be curious to see if this helps.

Once we merge this I would like to have a few people test it with realistic workloads and then fairly quickly remove the old storage code and the feature flag. The main reason for the feature flag is to make it easy for people to test alongside the existing implementation. I don't want it to stick around long term.

jkankiewicz:
> I've made a few changes to automerge-c to accomodate a new ScalarValue::Unknown variant. Hopefully what I've done is sensible but would be great to get your eyes on it. I've also not made any attempt to run the tests for automerge-c with a storage-v2 feature flag enabled.

@alexjg, if I have to revise your changes to the C API then should I submit a PR against the "storage-v2" branch?

alexjg:
That's probably the best way yeah.

.github/workflows/ci.yaml
```diff
@@ -78,6 +78,15 @@ jobs:
...
         run: ./scripts/ci/wasm_tests
 
+  wasm_tests_storage_v2:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install wasm-pack
+        run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+      - name: run tests
+        run: ./scripts/ci/wasm_tests
```
jeffa5, line 88: Should this be the v2 script?

automerge/src/change.rs
```diff
@@ -407,6 +427,13 @@ impl Change {
         self.bytes.compress(self.body_start);
     }
 
+    pub fn compressed_bytes(&self) -> &[u8] {
```
jeffa5, line 430: This name seems a bit confusing to me if we have the uncompressed variant. Perhaps something just like `bytes`?

automerge/src/columnar.rs
```diff
@@ -592,6 +592,9 @@ impl ValEncoder {
                 let len = (*n).encode(&mut self.raw).unwrap();
                 self.len.append_value(len << 4 | VALUE_TYPE_IEEE754);
             }
+            ScalarValue::Unknown { type_code, bytes } => {
+                panic!("unknown value")
```
jeffa5, line 596: Is there a plan for encoding unknown values in the future?

automerge/src/types/opids.rs
```diff
@@ -74,6 +94,38 @@ impl OpIds {
...
+    let mut last = match opids.next() {
+        Some(e) => e,
+        None => return true,
+    };
+
+    for next in opids {
+        if f(last, next) == Ordering::Greater {
+            return false;
+        }
+        if last == next {
```
jeffa5, line 123: Could this be a `Ordering::Equal` check on the above if statement?

jeffa5 (commented):
Had a very quick skim. With the current size I don't think I'll really get time to go through this fully so not sure what to do with it really. As long as it passes the tests, you're happy with it and confident in the work then I wouldn't be strongly against getting it in.

jkankiewicz:
> That's probably the best way [to revise my changes to the C API] yeah.

@alexjg, I've opened PR #415.

alexjg:
Right, I've updated this. Here are the changes:

* Remove the storage-v2 feature flag and all the old parsing and decoding/encoding logic
* Allow incremental loads in `automerge::Automerge::load_incremental` to
  continue in the presence of partial loads.
* Rename `Change::compressed_bytes` -> `Change::bytes`

Obviously the biggest change is removing the feature flag. This is as per a discussion with @orionz where we decided that it didn't make sense to maintain the feature flag going forward (although it has been useful whilst developing this patch). I've elected to leave the feature flag in for most of the commits and add a commit at the end which removes it. This makes it a bit easier to follow the reasoning of the commits.

.github/workflows/ci.yaml
alexjg: It should have been, I actually fixed this in the latest rebase but it's also somewhat irrelevant now that we're removing the storage-v2 feature flag.

automerge/src/change.rs
alexjg: Done!

automerge/src/columnar.rs
alexjg: This code would always have panicked in the case of unknown values so this isn't strictly a downgrade. The new code does handle this case and now that we've removed the old encoding/decoding code (i.e. this code) it doesn't matter anyway.

automerge/src/types/opids.rs
alexjg: Done!

alexjg:
Whelp, appears that a bunch of the `automerge-c` tests are failing - I must have messed up the rebase. Will try and fix this weekend.

alexjg:
I've incorporated the changes @jkankiewicz helped me figure out to get the C tests passing.  This is now ready to merge.

Issue #353, opened 2022-04-26
Automerge::range and thus Automerge::values do not support lists


luleyleo:
I ran into an error in my app and reproduced it with the following test:
```rust
#[test]
fn get_list_values() {
    let mut doc = AutoCommit::new();
    let list = doc.put_object(ROOT, "list", ObjType::List).unwrap();

    // insert elements
    doc.insert(&list, 0, "First").unwrap();
    doc.insert(&list, 1, "Second").unwrap();
    doc.insert(&list, 2, "Third").unwrap();

    // list elements
    let elements = doc
        .values(list)
        .map(|(key, _value, _id)| key)
        .collect::<Vec<_>>();

    println!("{:?}", elements);
    assert_eq!(elements.len(), 3);
}
```

Which results in:
```
thread 'get_list_values' panicked at 'found list op in range query', automerge/src/query/range.rs:43:36
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/panicking.rs:116:14
   2: <automerge::query::range::Range<R> as core::iter::traits::iterator::Iterator>::next
             at ./src/query/range.rs:43:36
   3: <automerge::values::Values as core::iter::traits::iterator::Iterator>::next
             at ./src/values.rs:22:9
   4: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::next
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/iter/adapters/map.rs:103:9
   5: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter_nested::SpecFromIterNested<T,I>>::from_iter
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/alloc/src/vec/spec_from_iter_nested.rs:23:32
   6: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/alloc/src/vec/spec_from_iter.rs:33:9
   7: <alloc::vec::Vec<T> as core::iter::traits::collect::FromIterator<T>>::from_iter
[...]

PR #356, opened 2022-04-29
fixed panic in doc.values() - fixed concurrency bugs in range

orionz:
`doc.values()` which was intended to iterate over the values in a list was calling range (for maps) and panicking.  There were no tests for `doc.values()` so this slipped through.  I've added tests and created a new query `list_range.rs` (possible `range.rs` should be renamed to `map_range.rs`).   While copying the code I found a bug in in Range.  When values were concurrent it would pick the first in the op range and not the last (but got it right when going in reverse).   I added breaking tests and then got them to pass.  I noticed the double ended iterator tracks a `last_key` and a `last_key_back`.  I have a feeling it would not behave correctly if we reversed a reverse in the middle an iteration (tests needeed) and also noticed that range iterates over all the elements - not making any effort to jump to the start of the range - or exiting when reaching the end of the range.  Seems needlessly slow.

Also seems like we should either merge range and values into a `range_values( .., Range<Prop>)` function or name them `map_range_values()` and `list_range_values()`.

  Also the `range_at` and `list_range_at` will not produce correct results if iterating over counters and increments.   I have some ideas on how to clean up the query code to address these issues (inspired by @jeffa5's iterator work).  Coming soon... but this fixes a serious bug so I wanted to get it in...


luleyleo:
Thanks, this solves my issue! Closes #353

luleyleo:
@orionz I ran into an issue with this. The following test fails:
```rust
#[test]
fn insert_at_index() {
    let mut doc = AutoCommit::new();
    
    let list = &doc.put_object(ROOT, "list", ObjType::List).unwrap();
    doc.insert(list, 0, 0).unwrap();
    doc.insert(list, 0, 1).unwrap();

    assert_eq!(doc.length(list), 2);
    assert_eq!(doc.keys(list).count(), 2);
    assert_eq!(doc.values(list).count(), 2); // it's just 1!
}
```

automerge/src/query/list_range.rs
```diff
@@ -0,0 +1,55 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        let mut result = None;
+        for i in self.index..self.index_back {
+            let op = self.root_child.get(i)?;
+            self.index += 1;
+            if op.visible() {
+                if self.range.contains(&self.pos) {
+                    result = self.next_result.replace((self.pos, op.value(), op.id));
+                }
+                if Some(op.key) != self.last_key {
```
luleyleo, line 44: Apparently, this condition is at fault. It seems to me like we would have to differentiate between inserts and puts here? 

```suggestion
                if op.insert || Some(op.key) != self.last_key {
```

With this, both the tests you added, and mine, pass.

orionz: Excellent find! - actually the problem is I based this on the range code which using maps relies on keys - list ops are a little more complicated and we need to go off `op.elemid()`.  So glad you caught this.

automerge/src/automerge.rs
```diff
@@ -310,7 +310,7 @@ impl Automerge {
     /// For a list the keys are the element ids (opids) encoded as strings.
     pub fn values<O: AsRef<ExId>>(&self, obj: O) -> Values<'_> {
         if let Ok(obj) = self.exid_to_obj(obj.as_ref()) {
-            let iter_range = self.ops.range(obj, ..);
+            let iter_range = self.ops.list_range(obj, ..);
```
jeffa5, line 313: Does this now remove support for `values` on a map?

orionz: Did you want something like this?

```
#[derive(Debug)]
pub(crate) enum ValuesIter<'a> {
    None,
    List(query::ListRange<'a, RangeFull>),
    Map(query::MapRange<'a, RangeFull>),
}


#[derive(Debug)]
pub struct Values<'a> {
    range: ValuesIter<'a>,
    doc: &'a Automerge,
}

impl<'a> Values<'a> {
    pub(crate) fn new<T: Into<ValuesIter<'a>>>(doc: &'a Automerge, range: T) -> Self {
        Self {
            range: range.into(),
            doc,
        }
    }
}

impl<'a> Iterator for Values<'a> {
    type Item = (Value<'a>, ExId);

    fn next(&mut self) -> Option<Self::Item> {
        match &mut self.range {
            ValuesIter::None => None,
            ValuesIter::List(l) => l
                .next()
                .map(|(_idx, value, id)| (value, self.doc.id_to_exid(id))),
            ValuesIter::Map(m) => m
                .next()
                .map(|(_prop, value, id)| (value, self.doc.id_to_exid(id))),
        }
    }
}```

File: automerge/src/automerge.rs [hunk 3]
```diff
@@ -279,52 +281,91 @@ impl Automerge {
...
+        range: R,
+    ) -> ListRange<'_, R> {
+        if let Ok(obj) = self.exid_to_obj(obj.as_ref()) {
+            ListRange::new(self, self.ops.list_range(obj, range))
+        } else {
+            ListRange::new(self, None)
+        }
+    }
+
+    /// Historical version of [`range`](Self::list_range_at).
```
jeffa5, line 325:
```suggestion
    /// Historical version of [`list_range`](Self::list_range).
```

jeffa5 (approved):
Think CI needs fixing but looks ok.

PR #357, opened 2022-04-30
Make the OpSet iterator faster

alexjg:
The opset iterator was using `OpTreeInternal::get(index)` to fetch each
successive element of the OpSet. This is pretty slow. We make this much
faster by implementing an iterator which is aware of the internal
structure of the OpTree.

This speeds up the save benchmark by about 10%.

Signed-off-by: Alex Good <alex@memoryandthought.me>

### DIFF
--- a/automerge/src/op_set.rs
+++ b/automerge/src/op_set.rs
@@ -1,7 +1,7 @@
 use crate::clock::Clock;
 use crate::exid::ExId;
 use crate::indexed_cache::IndexedCache;
-use crate::op_tree::OpTree;
+use crate::op_tree::{self, OpTree};
 use crate::query::{self, OpIdSearch, TreeQuery};
 use crate::types::{self, ActorId, Key, ObjId, Op, OpId, OpType};
 use crate::{ObjType, OpObserver};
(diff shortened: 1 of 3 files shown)


automerge/src/op_set.rs
```diff
@@ -45,13 +45,11 @@ impl OpSetInternal {
     }
 
     pub(crate) fn iter(&self) -> Iter<'_> {
-        let mut objs: Vec<_> = self.trees.keys().collect();
-        objs.sort_by(|a, b| self.m.lamport_cmp(a.0, b.0));
+        let mut objs: Vec<_> = self.trees.iter().collect();
+        objs.sort_by(|a, b| self.m.lamport_cmp((a.0).0, (b.0).0));
```
jeffa5, line 49: I wonder if it would be worth trying out a btreemap instead of a hashmap for the trees at some point so that we don't have to do an intermediate collect and sort.

jeffa5 (approved):
Looks good!

automerge/src/op_set.rs
alexjg: O like it, although o suppose for that to work we would have to use the `ExId` as a key?

jeffa5: Maybe, so it might not be so feasible.

PR #362, opened 2022-05-06
Add tests and fixes for double ended map range iterator

jeffa5:


orionz:
Great!  Did you want to implement double ended iterator for values() and list_range() as well?  I left an `unimplemented!()` in values.rb that should be resolved either by adding the functionality or removing it.

PR #365, opened 2022-05-07
Implement OpTreeIter::nth correctly

alexjg:
The previous implementation of nth was incorrect, it returned the nth
element of the optree but it did not modify the internal state of the
iterator such that future calls to `next()` were after the nth element.
This commit fixes that.

Signed-off-by: Alex Good <alex@memoryandthought.me>


orionz:
Love this - can you fix the fmt and I'll merge?

alexjg:
The `fmt` was failing due to the CI toolchain being out of date. I've rebased this on top of #366 which fixes the problem

PR #366, opened 2022-05-09
Update CI toolchain to 1.60

alexjg:


jeffa5:
If we're bumping this should we also update the rust-version in the Cargo.tomls? e.g. https://github.com/automerge/automerge-rs/blob/409189e36a1587b9c796bbc92aef8c2c86e5a451/automerge/Cargo.toml#L8

PR #368, opened 2022-05-16
Don't remove last actor when there are none

jeromegn:
This causes a panic if there are no actors in the optree.

### DIFF
--- a/automerge/src/transaction/inner.rs
+++ b/automerge/src/transaction/inner.rs
@@ -87,7 +87,7 @@ impl TransactionInner {
         }
 
         // remove the actor from the cache so that it doesn't end up in the saved document
-        if doc.states.get(&self.actor).is_none() {
+        if doc.states.get(&self.actor).is_none() && doc.ops.m.actors.len() > 0 {
             let actor = doc.ops.m.actors.remove_last();
             doc.actor = Actor::Unused(actor);
         }


alexjg (approved):

jeromegn:
@jeffa5 I didn't include one, but I think this would trigger it:

```rust
let mut doc = automerge::Automerge::new();
let tx = doc.transaction();
tx.rollback(); // or let it drop
```

jeffa5:
That isn't triggering it for me, when creating a transaction we explicitly put the actor in the cache, which is why I'm concerned that this was reached without one there.

jeffa5:
https://github.com/automerge/automerge-rs/blob/409189e36a1587b9c796bbc92aef8c2c86e5a451/automerge/src/automerge.rs#L116 calls `get_actor_index`, which should cache the actor id if it isn't already https://github.com/automerge/automerge-rs/blob/409189e36a1587b9c796bbc92aef8c2c86e5a451/automerge/src/automerge.rs#L96-L100

jeromegn:
That's odd.

I wonder if I was creating UB. I'm using `rquickjs` and I'm not entirely sure about the soundness of its bindings. Mostly a feeling that my mental model of Rust breaks down when passing and retrieving values with lifetime from the JS context.

Let me try and recreate this.

PR #369, opened 2022-05-16
Webpack

orionz:
moved example code into automerge-wasm folder - added webpack example code


ept:
Hi @orionz, I just tried this and it seems like webpack isn't actually doing anything here… it emits a 5kB JS file that simply requires `automerge-wasm` and `./src/index.js` and then runs them. I assumed a JS bundler was supposed to make a self-contained copy of the code and place it in the `dist` directory, but that's not happening here – the file in `dist` contains neither the Wasm module, nor the the JS wrapper for it, nor the hello world code from `src/index.js`.

ept:
Oh, I see now that you've intentionally configured the Node bundler to require the external libraries, and the browser config does seem to be bundling everything together into the `public` directory. That should be fine then.

JS bundling is such a pain…😖

PR #371, opened 2022-05-17
Convert automerge-js to typescript

orionz:
Converting `automerge-js` to typescript to enable easy packaging as we'll inherit the `automerge-wasm` packaging issues,  Step one is done but the js is littered with any's.  Fully getting `tslint` happy and no any's is a long way away but before im done with this I want to finish the npm package building and have the autogenerated types as close to the original typescript types file.


automerge-js/package.json
```diff
@@ -4,10 +4,21 @@
   "main": "src/index.js",
   "license": "MIT",
   "scripts": {
-    "test": "mocha --bail --full-trace"
+    "lint": "tslint --project tsconfig.json",
```
scotttrinh, line 7: Due to [deprecation of tslint in 2019](https://blog.palantir.com/tslint-in-2019-1a144c2317a9), any reason not to prefer [`typescript-eslint`](https://typescript-eslint.io)? I can help with any setup issues that might come up, we've recently done a migration.

orionz: Thank you!  I haven't done much typescript work so I'll take all the help i can get.  Since I did a pretty fast and rough javascript to typescript conversion were a long way from the lint passing but I really wanted to have it in here as a bar to reach.

orionz:
Ok - the typescript is now compiling into cjs and mjs and I have can build a single npm package that works for node and web with optional async wasm loading.  Very close to an actual package.  Lots of clean up work to be done tho - the typescript is littered with `any`s.

ept:
I just had a thought. If I understand correctly, your motivation for using TypeScript is to allow us to build separate Node and browser versions of the high-level JS API, in order to deal with the fact that the low-level Wasm API also has separate Node and browser versions. However, a different way of achieving this would be to ask users to load the high-level API and the Wasm module separately, and then register the loaded Wasm module with the high-level API. Something like this (rough sketch):

```js
import * as Automerge from 'automerge'
import init from 'automerge-wasm'

init().then(lowLevelAPI => {
  Automerge.use(lowLevelAPI)
})
```

If we do this, the `automerge` library doesn't need to know about the differences between Node and browser. It also leaves open the option of having a pure-JS implementation of the low-level API in case some people have problems using Wasm.

Of course we can still use TypeScript anyway, and I'm sure the TS enthusiasts who have previously campaigned for switching to TS will be delighted. But I'm not sure that building two different versions of the high-level API for Node and browser is the right way forward.

orionz:
That's a cool idea... My first order goal was to make a package that could be a drop in replacement for `automerge` proper.  Adding an extra `Automerge.use()` violates this in a small way but I like it.  

We could do two packages - one that loads the wasm for you (which is just a tiny wrapper around the one that doesnt).  Or just make people call `Automerge.use()` - thats fine too I guess.  There will be unavoidable differences in this package no matter what (such as the need to call `init().then()` in the browser.

orionz:
Ill give it a shot - it'll make the typescript a little more tricky since the Automerge code will no longer have direct access to the `automerge-wasm` types...

automerge-js/package.json
orionz: Hey Scott!  I mean... this PR is going to get merged with probably 200+ lint errors in the TODO list... A PR that fixed any one of those would be warmly welcomes 😂

orionz:
Ok.  Cleaned up most but not all of the typescript errors.  Have (almost) successfully removed `automerge-wasm` as a dependency.  Only including it for its types currently.  Might need to make a separate `@types/automerge-wasm` package with just type info they can both depend on.  One ugly thing is needing a type for the top level module.  Currently I have this

```ts
export function create(actor?: Actor): Automerge;
export function load(data: Uint8Array, actor?: Actor): Automerge;
export function encodeChange(change: DecodedChange): Change;
export function decodeChange(change: Change): DecodedChange;
export function initSyncState(): SyncState;
export function encodeSyncMessage(message: DecodedSyncMessage): SyncMessage;
export function decodeSyncMessage(msg: SyncMessage): DecodedSyncMessage;
export function encodeSyncState(state: SyncState): Uint8Array;
export function decodeSyncState(data: Uint8Array): SyncState;
export function exportSyncState(state: SyncState): JsSyncState;
export function importSyncState(state: JsSyncState): SyncState;

export class API {
  create(actor?: Actor): Automerge;
  load(data: Uint8Array, actor?: Actor): Automerge;
  encodeChange(change: DecodedChange): Change;
  decodeChange(change: Change): DecodedChange;
  initSyncState(): SyncState;
  encodeSyncMessage(message: DecodedSyncMessage): SyncMessage;
  decodeSyncMessage(msg: SyncMessage): DecodedSyncMessage;
  encodeSyncState(state: SyncState): Uint8Array;
  decodeSyncState(data: Uint8Array): SyncState;
  exportSyncState(state: SyncState): JsSyncState;
  importSyncState(state: JsSyncState): SyncState;
}
```
Where I export the top level functions and then define an interface which exports these exact same functions.  Im sure there's a better way to do this so any typescript pointers would be wonderful.

orionz:
Thoughts @scotttrinh ?

scotttrinh:
@orionz Sorry for the delay on the ping here, I'll take a look today and see how I can help! Should be able to reduce some of the duplication there.

PR #375, opened 2022-05-19
Get changes opt

jeffa5:
In order to improve sync performance I've taken a look at the `generate_sync_message` functionality and it was quite largely spending time in getting the changes, this can be especially bad when we can't use the fast version...

This PR optimises the `get_changes` function by doing the following:
- Maintain a vector clock for each change (hash) to allow constant time calculation of the vector clock
- Use this new `clocks` state to speed up finding the clock at any given hash
- Change `get_changes` to use the clocks under the hood, converting the hashes first to a single clock and then building the resulting changes from that
- Precalculate the number of operations in a change and store it alongside, otherwise we have to individually count the operations every time

This does add overhead to each document's memory to store the clocks, I'm sure we can optimise this somehow but I'm basically trying to have syncing not take too long, even though it is something that shouldn't be done too frequently it still shouldn't take multiple seconds!

**Note**: speeding up the `clock_at` function also speeds up all of the `_at` queries for historical content so this is kind of a double win!

And this turned up a bug in the wasm tests which is fixed here.

### DIFF
--- a/automerge/src/automerge.rs
+++ b/automerge/src/automerge.rs
@@ -4,6 +4,7 @@ use std::num::NonZeroU64;
 use std::ops::RangeBounds;
 
 use crate::change::encode_document;
+use crate::clock::ClockData;
 use crate::exid::ExId;
 use crate::keys::Keys;
 use crate::op_observer::OpObserver;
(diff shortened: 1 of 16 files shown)


automerge/src/automerge.rs
```diff
@@ -995,11 +972,26 @@ impl Automerge {
...
+        let actor_index = self.ops.m.actors.cache(change.actor_id().clone());
         self.states
-            .entry(self.ops.m.actors.cache(change.actor_id().clone()))
+            .entry(actor_index)
             .or_default()
             .push(history_index);
 
+        let mut clock = Clock::new();
+        for hash in &change.deps {
+            let c = self.clocks.get(hash).unwrap();
```
alexjg, line 983: How do we know this unwrap is okay?

automerge/src/clock.rs
```diff
@@ -1,30 +1,53 @@
...
+pub(crate) struct ClockData {
+    /// Maximum operation counter of the actor at the point in time.
+    pub(crate) max_op: u64,
+    /// Sequence number of the change from this actor.
+    pub(crate) seq: u64,
+}
+
+/// Vector clock mapping actor indices to the max op counter of the changes created by that actor.
+#[derive(Default, Debug, Clone, PartialEq)]
+pub(crate) struct Clock(HashMap<usize, ClockData, FxBuildHasher>);
```
alexjg, line 15: Does this perform any differently with a `BTreeMap`? `BTreeMap` feels like a more natural fit here where we have integer keys and most of the time a fairly small amount of data.

automerge/src/automerge.rs
jeffa5: Updated to an expect.

automerge/src/clock.rs
jeffa5: I've not tried but I'm just leaving that as it was for now. May be worth looking at later if it shows up as a bottleneck.

orionz:
There is a small change in behavior here where `get_changes()` returns an error instead of an empty set when given an unknown hash.  Seems reasonable.   Just want to make sure we discuss weather this is the most useful behavior.  I'm sad to see more places where we need unwraps but think this is probably a good change.  Thoughts?

PR #376, opened 2022-05-20
Use BTreeSet for sync::State to allow deriving Hash

jeffa5:


orionz:
Help me understand what this fixes?

jeffa5:
`HashSet` doesn't implement `Hash` which means we can't store it in some scenarios (I'm particularly trying out some model checking), `BTreeSet` does implement `Hash` so we can use it as a substitute.

jeffa5:
Maybe I should just create a new branch for my stuff since it might be making different tradeoffs from the core logic (though I want to keep it very close).

orionz:
No this is fine.  I just wanted to know.  As long as we're not making the code overly complex or slow its all good.

PR #386, opened 2022-06-07
Patch v2

orionz:
Re work the patch interface to include paths that are current making updating local state trivial.  Still several shortcomings id like to work through.   `OpObserver` is no longer a trait.  `AutoCommit` is now handling its own `OpObserver`.  And manual transactions need to be told to collect data on creation AND given an observer on commit which seems wrong.


alexjg:
What's the reasoning behind removing the `OpObserver` trait?

jeffa5 (commented):
This seems to have rather a lot of unrelated changes in, making it quite hard to review.

From what I can see I think we should still keep the `OpObserver` trait and have `VecOpObserver` be a separate type that users can use (as well as the wasm impl).

I'm also not sure why the patch format is now containing the entire path, I thought the issue was just in the index of lists?

I'm actually not sure exactly what this was trying to fix now, It seems that the old commit would have given the index of the actual insert at the time of the operation anyway. Is this an issue with the applying changes logic when calculating the patches?

PR #391, opened 2022-06-14
Add `AMactorId` to the C API

jkankiewicz:
@orionz, I've exposed the `automerge::ActorId` struct to C so that the logic to convert from a hexadecimal string to bytes and _vice versa_ wouldn't be repeated for multiple structs and so that a client application/library can synthesize a conforming actor identifier without guesswork.

@lightsprint09, this PR removes `AMgetActorHex()`, `AMsetActorHex()` and updates `AMgetActor()`, `AMsetActor()` and `AMchangeActorId()`.

### DIFF
--- a/automerge-c/src/CMakeLists.txt
+++ b/automerge-c/src/CMakeLists.txt
@@ -51,6 +51,7 @@ add_custom_command(
     MAIN_DEPENDENCY
         lib.rs
     DEPENDS
+        actor_id.rs
         byte_span.rs
         change_hashes.rs
         change.rs
--- a/automerge-c/src/actor_id.rs
+++ b/automerge-c/src/actor_id.rs
@@ -1,0 +1,132 @@
+use automerge as am;
+use std::cell::RefCell;
+use std::ffi::{CStr, CString};
+use std::os::raw::c_char;
+use std::str::FromStr;
+
+use crate::byte_span::AMbyteSpan;
+use crate::result::{to_result, AMresult};
+
+/// \struct AMactorId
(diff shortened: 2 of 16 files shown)


alexjg:
What's the reasoning behind carrying around a cached string representation of the actor ID? Is that idiomatic in C? It feels strange to me because it forces a bunch of memory use on the user when they may not care at all about hex encodings of the actor.

jkankiewicz:
> What's the reasoning behind carrying around a cached string representation of the actor ID?

@alexjg, it's there so that `AMactorIdStr()` can return a `const char*` (the hexadecimal string) instead of taking `char*`(a buffer into which the hexadecimal string is to be copied) and `size_t` (the length of the buffer) arguments and returning a 'size _t` (the length of the hexadecimal string so that one can detect when the buffer is too small).
This simplifies API usage in the following ways:
* The alternate approach requires two `AMactorId()` calls--first with a `NULL` `char*` argument to retrieve the length of the buffer to allocate and second to copy it--because the length of the hexadecimal string varies between `ActorId`s.
* The alternate approach requires a subsequent `free()` call (for the buffer) in addition to the subsequent `AMfree()` call (for the `AMresult` containing the `ActorId`) or it will leak the copy of the string.

> Is that idiomatic in C?

No.

> It feels strange to me because it forces a bunch of memory use on the user when they may not care at all about hex encodings of the actor.

Only if the user actually calls `AMactorIdStr()` on the given `AMactorId`, otherwise it only forces a `RefCell<Option<ActorId>>` of memory use on the user.

PR #392, opened 2022-06-18
Fix TypeScript syntax error in `automerge-wasm` definitions

rf-:
I'm not sure if there are some configurations under which this works,
but I get

    index.d.ts:2:21 - error TS1005: ';' expected.

    2 export default from "automerge-types"
                          ~~~~~~~~~~~~~~~~~

both in my project that depends on `automerge-wasm` and when I run `tsc`
in this repo.

It seems like `export default from` is still a Stage 1 proposal, so I
wouldn't expect it to be supported by TS, although I couldn't really
find hard evidence one way or the other. It does seem like this syntax
should be exactly equivalent based on the proposal doc though.

### DIFF
--- a/automerge-wasm/index.d.ts
+++ b/automerge-wasm/index.d.ts
@@ -1,2 +1,2 @@
 export * from "automerge-types"
-export default from "automerge-types"
+export { default } from "automerge-types"


orionz:
Doesn't the change proposed though export the old default export as a property export named 'default'?   If so this is incorrect.

orionz:
Im somewhat new to typescript.  Im happy to take pointers on how to do this correctly.

rf-:
Yeah this aspect was a bit surprising to me too—it seems like the name `default` is actually special-cased? Check out this part of the `export default from` proposal: https://github.com/tc39/proposal-export-default-from#common-concerns

Issue #394, opened 2022-06-27
Error: there was a decoding problem: Calculated heads differed from actual heads


rf-:
I somehow managed to create a doc ([attached](https://github.com/automerge/automerge-rs/files/8988451/automerge_data.gz)) that can load with vanilla Automerge but not automerge-js:
```
> Automerge.load(require('fs').readFileSync('automerge_data'))
Error: there was a decoding problem: Calculated heads differed from actual heads
    at module.exports.__wbg_new_55259b13834a484c (/Users/rfitz/src/gridlock/gridlock_app/node_modules/automerge-wasm/nodejs/bindgen.js:1430:15)
    at wasm://wasm/00346bd2:wasm-function[1520]:0xc2972
    at wasm://wasm/00346bd2:wasm-function[206]:0x5a450
    at wasm://wasm/00346bd2:wasm-function[1005]:0xb9165
    at module.exports.loadDoc [as load] (/Users/rfitz/src/gridlock/gridlock_app/node_modules/automerge-wasm/nodejs/bindgen.js:229:14)
    at Object.load (/Users/rfitz/src/gridlock/gridlock_app/node_modules/automerge-js/dist/cjs/index.js:138:42)
```
If I then save the doc with vanilla Automerge, it produces a byte-identical file.

alexjg:
I believe the issue here is that automerge-rs is expecting the document to have a sequence of ULEB128 encoded integers representing the indices of the changes at the head of the document in the set of encoded changes - as specified here: https://alexjg.github.io/automerge-storage-docs/#head-indexes

For some reason the data you've posted doesn't include that. How was it generated?

ept:
There are a few strange things about this file:

1. Like @alexjg pointed out, the headIndexes are missing. The JS implementation [tolerates this](https://github.com/automerge/automerge/blob/main/backend/new.js#L1727-L1739), because IIRC previous versions of Automerge did not include this data. However, I think recent versions of Automerge should always include headIndexes. Did you generate this file using an old version?
[...]

PR #401, opened 2022-07-11
Make Op::add_succ sort succ vector after adding

adelsz:
Fixes https://github.com/automerge/automerge-rs/issues/394

Let me know if there is a more elegant approach than sorting after push, maybe using something like a BinaryHeap instead of Vec for the succ list.

Optree before fix (note the ordering of successors on line 5):
image

Optree after fix:
image

### DIFF
--- a/automerge/src/columnar.rs
+++ b/automerge/src/columnar.rs
@@ -754,6 +754,16 @@ struct SuccEncoder {
     ctr: DeltaEncoder,
 }
 
+fn succ_ord(left: &OpId, right: &OpId, actors: &[usize]) -> Ordering {
+    match (left, right) {
+        (OpId(0, _), OpId(0, _)) => Ordering::Equal,
+        (OpId(0, _), OpId(_, _)) => Ordering::Less,
+        (OpId(_, _), OpId(0, _)) => Ordering::Greater,
+        (OpId(a, x), OpId(b, y)) if a == b => actors[*x].cmp(&actors[*y]),
+        (OpId(a, _), OpId(b, _)) => a.cmp(b),
+    }
+}
+
 impl SuccEncoder {
     fn new() -> SuccEncoder {
         SuccEncoder {
@@ -765,7 +775,9 @@ impl SuccEncoder {
 
     fn append(&mut self, succ: &[OpId], actors: &[usize]) {
         self.num.append_value(succ.len());
-        for s in succ.iter() {
+        let mut sorted_succ = succ.to_vec();
+        sorted_succ.sort_by(|left, right| succ_ord(left, right, actors));
+        for s in sorted_succ.iter() {
             self.ctr.append_value(s.0);
             self.actor.append_value(actors[s.1]);
         }


alexjg:
Have you confirmed that this  fixes your issue? I've been sitting on an identical patch for a while because I haven't been able to make a unit test which reproduces the problem which makes me worry we have not understood the problem.

adelsz:
Yes, here is how I am testing:
I have an automerge binary file which I load from disk (using `Automerge::load`) and then serialize (using automerge-rs `doc.save`) to send over wire to my frontend automerge-js. JS frontend always crashes (when calling `Automerge.load`) with "operation IDs are not in ascending order" with upstream version and works fine with this branch.

alexjg:
Hmm, I mean, this code is pretty obviously the correct thing to do and it fixes your problem and all other tests pass. On that basis we should merge it but I want to find a good test if possible. If I haven't got a good reproduction in the next day I'll merge regardless.

adelsz:
Just encountered a different document which also crashes automerge js 😞 .
Tried to dig deeper to understand what the expected succ sorting is supposed to be. Here is a screenshot of the [debugged](https://github.com/automerge/automerge/blob/9dd494baff4aa0f9ad885684b0b90b71e201909a/backend/columnar.js#L519) offending values: 
Screenshot 2022-07-12 at 01 13 27
As I understand it, in automerge-rs we are sorting successors `struct OpId(u64, usize)` by "mapped" `actorId` which are counted `usize` numbers, but automerge js expects them to be sorted as uuid strings?
Shouldn't automerge js use the [actorIdToActorNum](https://github.com/automerge/automerge/blob/9dd494baff4aa0f9ad885684b0b90b71e201909a/backend/columnar.js#L101) before doing the comparison?

Here is a screenshot of the actorId mapping I expected automerge js to use while sorting:
Screenshot 2022-07-12 at 01 17 53

jeffa5:
> As I understand it, in this PR I am sorting successors struct OpId(u64, usize) by "mapped" actorId which are counted usize numbers, but automerge js expects them to be sorted as uuid strings?
Shouldn't automerge js use the [actorIdToActorNum](https://github.com/automerge/automerge/blob/9dd494baff4aa0f9ad885684b0b90b71e201909a/backend/columnar.js#L101) before doing this [comparison](https://github.com/automerge/automerge/blob/9dd494baff4aa0f9ad885684b0b90b71e201909a/backend/columnar.js#L114)?

On the rust side you'll probably need to use the `OpSetMetadata` to get the correct sorting using `lamport_cmp` for the `OpId`s. This will need to be wired up in a few places.

jeffa5:
Having a quick go and following through compile errors: https://github.com/jeffa5/automerge-rs/tree/sort-succ

adelsz:
> So, I tried this and the borrow checker got very upset with me furthermore, it seems a little fragile to me to maintain this invariant (succs are encoded in actor ID order) in the Op, which knows nothing about storage. I think it might be more fruitful to ensure the sort order when encoding


Sounds good, do you mind pointing me to the encoding logic for successors? I can probably take a stab at it today.

adelsz:
> Can you squash this into a single commit with a commit message describing the change and reason for it? (I know our commit log is not a good example of this behavior right now but I want to start being stricter about it).

Sure, done.

PR #402, opened 2022-07-12
Don't build tests for docs

### LINKED ISSUE #398: Docs workflow failing
The docs workflow is failing on main, seems to be an issue finding cmocka which was removed in https://github.com/automerge/automerge-rs/commit/4f7843e00739c58d22a7bb510968037bc256bd04, @jkankiewicz should cmocka be removed from somewhere else or added back into the workflow?

jeffa5:
The test `CMakeLists.txt` brings in cmocka but we don't actually need to
build the tests to get the docs. This just makes the cmake docs script
tell cmake not to build docs.

Fixes #398

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -2,6 +2,9 @@ on:
   push:
     branches:
       - main
+  pull_request:
+    branches:
+    - main
 
 name: Documentation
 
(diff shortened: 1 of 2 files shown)


jeffa5:
Ignoring unrelated advisory error

PR #407, opened 2022-07-14
Add OpIds to enforce ordering of Op::succ and Op::pred

alexjg:
I was thinking a bit more about #401 and I've decided that actually we probably should maintain the op ID ordering invariant in the OpSet. My reasoning is that sorting op IDs when encoding forces us to allocate for every op in every change which has a succ when we commit a transaction and I worry about this affecting performance. This PR introduces an `OpIds` type which enforces the sorting invariant.

### DIFF
--- a/automerge/src/types/opids.rs
+++ b/automerge/src/types/opids.rs
@@ -1,0 +1,126 @@
+use itertools::Itertools;
+
+use super::OpId;
+
+/// A wrapper around `Vec<Opid>` which preserves the invariant that the ops are
+/// in ascending order with respect to their counters and actor IDs. In order to
+/// maintain this invariant you must provide a comparator function when adding
+/// ops as the actor indices in an  OpId are not sufficient to order the OpIds
+#[derive(Debug, Clone, PartialEq, Default)]
+pub(crate) struct OpIds(Vec<OpId>);
+
+impl<'a> IntoIterator for &'a OpIds {
+    type Item = &'a OpId;
+    type IntoIter = std::slice::Iter<'a, OpId>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.iter()
+    }
+}
+
+impl OpIds {
+    pub(crate) fn new<I: Iterator<Item = OpId>, F: Fn(&OpId, &OpId) -> std::cmp::Ordering>(
+        opids: I,
+        cmp: F,
+    ) -> Self {
+        let mut inner = opids.collect::<Vec<_>>();
+        inner.sort_by(cmp);
+        Self(inner)
+    }
+
+    /// Add an op to this set of OpIds. The `comparator` must provide a
+    /// consistent ordering between successive calls to `add`.
+    pub(crate) fn add<F: Fn(&OpId, &OpId) -> std::cmp::Ordering>(
+        &mut self,
+        opid: OpId,
+        comparator: F,
+    ) {
+        use std::cmp::Ordering::*;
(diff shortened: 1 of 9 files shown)


alexjg:
@adelsz I would be grateful if you could test that this change still fixes your issue.

automerge/src/types/opids.rs
```diff
@@ -0,0 +1,126 @@
...
+    ) {
+        use std::cmp::Ordering::*;
+        if self.is_empty() {
+            self.0.push(opid);
+            return;
+        }
+        let idx_and_elem = self
+            .0
+            .iter()
+            .find_position(|an_opid| matches!(comparator(an_opid, &opid), Greater | Equal));
```
adelsz, line 46: What do you think about using [binary_search_by](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.binary_search_by) here to speed it up to logn?

automerge/src/types/opids.rs
```diff
@@ -0,0 +1,126 @@
+use itertools::Itertools;
+
+use super::OpId;
+
+/// A wrapper around `Vec<Opid>` which preserves the invariant that the ops are
+/// in ascending order with respect to their counters and actor IDs. In order to
+/// maintain this invariant you must provide a comparator function when adding
+/// ops as the actor indices in an  OpId are not sufficient to order the OpIds
+#[derive(Debug, Clone, PartialEq, Default)]
+pub(crate) struct OpIds(Vec<OpId>);
```
adelsz, line 10: From an API point of view, maybe it would be good to make `OpIds` have a generic for specifying the comparator func. Otherwise the same `OpIds` instance can be used with different comparators leading to bugs?

automerge/src/types/opids.rs
```diff
@@ -0,0 +1,126 @@
...
+    pub(crate) fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    pub(crate) fn iter(&self) -> std::slice::Iter<'_, OpId> {
+        self.0.iter()
+    }
+
+    pub(crate) fn contains(&self, op: &OpId) -> bool {
+        self.0.contains(op)
```
adelsz, line 75: Fixing one comparator per `OpIds` might also allow to use [binary_search](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.binary_search) here.

adelsz (commented):
@alexjg Confirming that this branch works well with my documents.
Also added some suggestions/questions to this PR that you might find useful.

automerge/src/types/opids.rs
adelsz: That makes sense, thanks for explaining the rationale behind it.

automerge/src/types/opids.rs
```diff
@@ -0,0 +1,126 @@
...
+    pub(crate) fn new<I: Iterator<Item = OpId>, F: Fn(&OpId, &OpId) -> std::cmp::Ordering>(
+        opids: I,
+        cmp: F,
+    ) -> Self {
+        let mut inner = opids.collect::<Vec<_>>();
+        inner.sort_by(cmp);
+        Self(inner)
+    }
+
+    /// Add an op to this set of OpIds. The `comparartor` must provide a
```
jeffa5, line 31:
```suggestion
    /// Add an op to this set of OpIds. The `comparator` must provide a
```

jeffa5 (approved):
I think it looks good, nice to have it be upheld during the in-memory stored version too!

Issue #405, opened 2022-07-12
Release automerge (Rust) to crates.io


jeffa5:
The Rust implementation of Automerge is pretty mature now, with it working with the js suite of tests, having a wasm and js frontend on top of it and it seeing some adoption. Therefore I think it is about time that we cut an initial release to crates.io.

This would enable more uptake as it at least has some semblance of stability and we can keep better track of when things change (rather than having people pin the git and have to bisect things if an update breaks things).

Having it released further helps stabilisation and releases of related projects as they can pin a version rather than use the git versions again.

It is currently set to use version `0.1.0` which I think is reasonable.

Currently we have two crates for automerge on `crates.io`: 
- https://crates.io/crates/automerge
- https://crates.io/crates/automerge-rs
both owned by @alexjg, perhaps you could add some of us as owners of the crate so we can release too? (https://doc.rust-lang.org/cargo/reference/publishing.html#cargo-owner). In future we can work to have this released from github actions on a tag/release in github but let that not delay us now.

I'd think we want to go with automerge, since crates.io is for Rust crates and not js or wasm things anyway.

### Changelog

I think that after this release we should also start keeping a changelog file (e.g. https://keepachangelog.com/en/1.0.0/) so that everyone can inspect what changes in new releases and can update code accordingly!

### Current blockers

This section lists any outstanding issues or PRs that MUST be fixed / merged before the initial release, I don't think heavy performance work is necessary in this, mainly large bug fixes, of which I'm not aware of many.

- https://github.com/automerge/automerge-rs/issues/394

cc: @ept @orionz @alexjg

alexjg:
I'm inclined to agree with this. I'll sort out crates.io stuff tomorrow. One thing I think we should do - apropos of the changelog - is to start being much stricter about commit discipline, having discrete commits with good commit messages makes generating a changelog easy

jeffa5:
Great! For the commits, did you have a particular scheme in mind to help? https://www.conventionalcommits.org/en/v1.0.0/ perhaps?

alexjg:
I personally don't find conventional commits particularly useful at a project of this size. I think it's more just a case of including the commit messages in code review. I'm not against conventional commits if others would like that approach though.

tombh:
[...]

PR #410, opened 2022-07-18
Add range functions to C API

jkankiewicz:
@orionz, I've exposed ten missing methods to the C API through the following six functions:
* `AMfork()`
* `AMlistGet()`
* `AMlistRange()`
* `AMmapGet()`
* `AMmapRange()`
* `AMobjValues()`

@lightsprint09, I've made the following backwards-incompatible changes to the C API:
* Added a `heads` argument to `AMlistGet()`.
* Added a `heads` argument to `AMmapGet()`.
* Renamed `AMdup()` to `AMclone()`.
* Renamed `AMgetActor()` to `AMgetActorId()`.
* Renamed `AMsetActor()` to `AMsetActorId()`.
* Renamed the `AMstrings` struct to `AMstrs`.


alexjg:
@jkankiewicz the commits here seem quite small. Would it be possible to condenses them into a smaller set of commits with a high level description on each one of what's achieved?

jkankiewicz:
> Would it be possible to condenses them into a smaller set of commits with a high level description on each one of what's achieved?

@alexjg, is there an upper limit on the number of commits per PR?

alexjg:
It's more that the commit messages don't tell a good story about what's going on here. Quite a lot of the commits have very similar messages (e.g. there are more than four commits where the commit message contains "Added AMListItem") and don't describe why the change is necessary. It seems to me that those commits would be easier to write messages for if they were combined, with the commit message describing at a slightly higher level what's going on and why. Maybe it's just a case of writing different commit messages though rather than condensing the commits?

I know we haven't worried about commit messages much to date. I'm starting to think about them because this project is picking up contributors and users. I think that having a good commit trail makes it easy to review code but more importantly it makes it easier to understand why code is the way it is two years later when everyone has forgotten. I know people have different opinions about commit messages though and maybe it would be better to discuss that somewhere else, perhaps #405 ?

jkankiewicz:
@alexjg, I'll try to redo the commits for this PR with higher-level descriptions in their messages.

lightsprint09:
Is this ready to merge?

jkankiewicz:
> Is this ready to merge?

@lightsprint09, it's good to hear from you!

This PR won't be merged until @orionz has approved it so you may wish to ask him for an estimate regarding if and when that will happen.

PR #415, opened 2022-08-07
Patch C API changes for "storage-v2"

jkankiewicz:
> I've made a few changes to `automerge-c` to accomodate a new `ScalarValue::Unknown` variant. ... I've also not made any attempt to run the tests for automerge-c with a storage-v2 feature flag enabled.

@alexjg, this patch enables you to do that by appending `-DFEATURE_FLAG_STORAGE_V2=ON` to the CMake invocation.

The C API's unit test suite still passes when the "storage-v2" feature is enabled.


alexjg:
Fantastic, thanks!

PR #416, opened 2022-08-09
Minor issues

simonthum:
Hi,

I'm looking into this project in an effort to build an offline-first app. I am amazed by the overall quality and effort in this project! Thank you all!

But I also found a few glitches, maybe that is because my rust is either too old or too new or I'm not (yet) proficient in rust.

Anyway, my rustc is 1.62.1 (e092d0b6b 2022-07-16) and the CI script is issuing these as errors.

Also, the dev dependencies are missing cargo deny. This in turn has some more funny deps so I ended up installing perl modules I did not know existed. CI is is calling cargo deny, so it should at least be a documented dependency.

### DIFF
--- a/automerge-c/src/doc/list.rs
+++ b/automerge-c/src/doc/list.rs
@@ -314,11 +314,11 @@ pub unsafe extern "C" fn AMlistPutNull(
 ) -> *mut AMresult {
     let doc = to_doc_mut!(doc);
     let obj_id = to_obj_id!(obj_id);
-    let value = ();
+
     to_result(if insert {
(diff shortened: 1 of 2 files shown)


orionz:
Thanks for this - looks like there's a minor collision with the merge I just did - happy to merge this if you get it fixed.

jkankiewicz (commented):
This was already resolved by #414.

jkankiewicz:
> Was required to build with rust 1.62.

@orionz, to clarify, @simonthum's patch of automerge/src/decoding.rs was *not* made redundant by #414.

PR #419, opened 2022-08-21
Eliminate most C API test failures on "storage-v2"

jkankiewicz:
> Whelp, appears that a bunch of the automerge-c tests are failing...

@alexjg, I've eliminated all but one of the failures in the C API's unit tests on the "storage-v2" branch (#355); `test_should_be_able_to_fetch_changes_by_hash()` is still failing on the comparison of two change hashes that have heretofore been identical.


alexjg:
The last failing test should be equivalent to this test (which passes) right?

```
#[test]
fn fetch_changes_by_hash() {
    let (actor1, actor2) = sorted_actors();
    let mut doc1 = new_doc_with_actor(actor1);
    let mut doc2 = new_doc_with_actor(actor2);
    doc1.put(&automerge::ROOT, "a", "b").unwrap();
    doc2.put(&automerge::ROOT, "b", "c").unwrap();

    let hash1 = doc1.get_heads()[0];
    let hash2 = doc2.get_heads()[0];

    let change = doc1.get_change_by_hash(&hash1).unwrap();
    assert_eq!(hash1, change.hash());
    assert!(doc1.get_change_by_hash(&hash2).is_none());
}
```

alexjg:
After some more digging it appears that the problem is solved by the following diff:

```diff
diff --git a/automerge-c/src/byte_span.rs b/automerge-c/src/byte_span.rs
index 939a52c57..3ddd5b152 100644
--- a/automerge-c/src/byte_span.rs
+++ b/automerge-c/src/byte_span.rs
@@ -45,10 +45,13 @@ impl From<&mut am::ActorId> for AMbyteSpan {
 
 impl From<&am::ChangeHash> for AMbyteSpan {
     fn from(change_hash: &am::ChangeHash) -> Self {
-        Self {
-            src: change_hash.0.as_ptr(),
-            count: change_hash.0.len(),
-        }
+        let source = Box::new(change_hash.0.clone());
+        let result = Self {
+            src: source.as_ptr(),
+            count: source.len(),
+        };
+        std::mem::forget(source);
+        result
     }
 }
 ```
 
 I think the problem is something to do with the fact that the memory referred to by `Change::hash` is a little more indirected then in the non-storage-v2 version and so one way or another the `&am::ChangeHash` above ends up referring to memory on the stack. This worked previously when that memory happened to be valid but doesn't work now. In any case, I think the code I've posted here is required to make the `From` impl valid, does it look right?

**Comment by jkankiewicz** [comment]
> I think the code I've posted here is required to make the `From` impl valid, does it look right?

@alexjg, unfortunately, that implementation will leak memory because an `AMbyteSpan` is meant to be just a view onto memory that it doesn't own; the only things one ever has to free when using the C API is an `AMresult` or an `AMresultStack`.

> I think the problem is something to do with the fact that the memory referred to by `Change::hash` is a little more indirected then in the non-storage-v2 version and so one way or another the `&am::ChangeHash` above ends up referring to memory on the stack.

This entails that either `AMchange` must be extended to optionally store the bytes that the returned `AMbyteSpan` references the first time `AMchangeHash()` is called on it or `AMchangeHash()` must now return an `AMresult*` that contains an `AMbyteSpan`.

**Comment by alexjg** [comment]
Like so (this fixes the test)?

```diff
iff --git a/automerge-c/src/change.rs b/automerge-c/src/change.rs
index a63132311..3c29ba96a 100644
--- a/automerge-c/src/change.rs
+++ b/automerge-c/src/change.rs
@@ -23,17 +23,15 @@ macro_rules! to_change {
 pub struct AMchange {
     body: *mut am::Change,
     c_msg: RefCell<Option<CString>>,
+    c_changehash: RefCell<Option<am::ChangeHash>>,
 }
 
 impl AMchange {
     pub fn new(change: &mut am::Change) -> Self {
-        let c_message = match change.message() {
-            Some(c_message) => CString::new(c_message.as_bytes()).ok(),
-            None => None,
-        };
         Self {
             body: change,
-            c_msg: RefCell::new(c_message),
+            c_msg: Default::default(),
+            c_changehash: Default::default(),
         }
     }
 
@@ -53,6 +51,20 @@ impl AMchange {
         }
         std::ptr::null()
     }
+
+    pub fn hash(&self) -> AMbyteSpan {
+        let mut c_changehash = self.c_changehash.borrow_mut();
+        if let Some(c_changehash) = c_changehash.as_ref() {
+            c_changehash.into()
+        } else {
+            let hash = unsafe { (*self.body).hash() };
+            let ptr = c_changehash.insert(hash);
+            AMbyteSpan {
+                src: ptr.0.as_ptr(),
+                count: hash.as_ref().len(),
+            }
+        }
+    }
 }
 
 impl AsMut<am::Change> for AMchange {
@@ -174,8 +186,7 @@ pub unsafe extern "C" fn AMchangeFromBytes(src: *const u8, count: usize) -> *mut
 pub unsafe extern "C" fn AMchangeHash(change: *const AMchange) -> AMbyteSpan {
     match change.as_ref() {
         Some(change) => {
-            let hash: &am::ChangeHash = &change.as_ref().hash();
-            hash.into()
+            change.hash()
         }
         None => AMbyteSpan::default(),
     }
```

jkankiewicz:
> `AMchange` must be extended to optionally store the bytes that the returned `AMbyteSpan` references the first time `AMchangeHash()` is called...

> Like so (this fixes the test)?

@alexjg, yes, like that. 💯

Thanks for not returning an `AMresult*` from a property access function!

alexjg:
Excellent, okay, I'll add this stuff to the storage V2 PR tomorrow. Thanks for your help!

orionz:
Alex manually merged this.

PR #424, opened 2022-09-05
Correct documentation for AutoSerde

bluebear94:
This type claims to implement `Deserialize` but actually implements `Serialize`.


alexjg:
Good spot!

PR #425, opened 2022-09-06
Another rewrite of the patch interface

orionz:
This is another rewrite of the patch system.

The core problem i'm trying to address is that you cannot apply patches to a mirror of the document internal state given only an object id.  Consider this datastructure.

```js
    {  items : [ { id: 1, name: "car" }, { id: 2, name: "house" } ] }
```

And the two edits were `doc.items[1].name = "hovel"` and `doc.items.delete(0)`.

When processing the first patch `Patch::Put { id: "2@aabbcc", key: "name", value: "hovel" }` you might call `doc.path_to_object("2@aabbcc")` and it would tell you the path is `[ "items" , 0 ]` which is correct NOW but is not useful for applying the patch.  The only way to make this work under the current system would be for someone using the system to maintain a vector of elemid's for every list in the system external to the document.  This is a VERY high burden and is exactly what made the old frontend backend implementation's so flimsy and slow.

To fix this I'm changing two things.

The simple one is adding a `path: Vec<Prop>` to the patch and passing a `Parents<'_>` iterator in with the patch interface. 

The more complicated one is moving the `OpObserver` from the commit call to the transaction itself.  The reason for this is that the path needs to be generated at the moment the op is applied and cannot wait till the end of the transaction (as per the original problem).

This has many downstream effects. The first is that ManualTransaction and AutoCommit are now `ManualTransaction<OpObserver>` and `AutoCommit<OpObserver>` as they need to manage the observer itself.  Also I got rid of `ApplyOptions` as it only had an observer in it and everywhere it was used just passing an `&mut OpObserver` made more sense.  I would like feedback on where I put the responsibility of managing the observer generic - this is my fourth attempt at this and im pretty sure this is the simplest way to do it but im open to suggestions.  

There is a little confusion with the use of `Option<Obs>` vs `AutoCommit<()>`. I'm tempted to try to normalize on the Option behavior since there are places where knowing that nothing is observed is better than calling a bunch of no-op functions but this leads to other problems such as still needing a Generic type when not using an observer.  Ideas welcome.

It's a pretty big patch so please comments welcome

Style:  I tried to change the style as little as possible.  I do very much want to change all the `op_observer`s to `observer` and all the `Obs` to `O` because I like brevity but wanted this patch to be a small of a change as possible a deferred that until after discussion.


automerge-wasm/src/observer.rs
```diff
@@ -0,0 +1,323 @@
...
+            self.patches.truncate(0)
+        }
+        self.enabled = enable;
+    }
+
+    fn push(&mut self, patch: Patch) {
+        if let Some(tail) = self.patches.last_mut() {
+            if let Some(p) = tail.merge(patch) {
+                self.patches.push(p)
+            }
```
alexjg, line 29: This was quite confusing to read - I thought for a while that there was a missing `else`. It might read better if `Patch::merge` was called `Patch::attempt_merge`?

automerge-wasm/src/interop.rs
```diff
@@ -457,3 +458,143 @@ pub(crate) fn list_to_js_at(doc: &am::AutoCommit, obj: &ObjId, heads: &[ChangeHa
...
+        Value::Object(o) if o == &am::ObjType::Map || o == &am::ObjType::Table => {
+            Object::new().into()
+        }
+        Value::Object(_) => Array::new().into(),
+        Value::Scalar(v) => ScalarValue(v.clone()).into(),
+    }
+}
+
+pub(crate) fn apply_patch(obj: JsValue, patch: &Patch) -> Result<JsValue, JsValue> {
+    apply_patch2(obj, patch, 0)
```
alexjg, line 479: Do we need `apply_patch2` or can we just consolidate this to `apply_patch`?

automerge/src/transaction/manual_transaction.rs
```diff
@@ -20,14 +20,15 @@ use super::{CommitOptions, Transactable, TransactionInner};
 /// intermediate state.
 /// This is consistent with `?` error handling.
 #[derive(Debug)]
-pub struct Transaction<'a> {
+pub struct Transaction<'a, Obs: OpObserver> {
     // this is an option so that we can take it during commit and rollback to prevent it being
     // rolled back during drop.
     pub(crate) inner: Option<TransactionInner>,
     pub(crate) doc: &'a mut Automerge,
+    pub op_observer: Obs,
```
alexjg, line 28: This feels like it should be private, or at least crate private, we're likely to change this stuff a lot so I think putting this behind an accessor method as in `AutoCommit` would be sensible.

Also, should this be an `Option<Obs>` for cases where you're uninterested in observing?

automerge/src/transaction/inner.rs
```diff
@@ -405,15 +422,48 @@ impl TransactionInner {
...
+            self.do_insert(doc, op_observer, obj, pos, v.clone().into())?;
             pos += 1;
         }
         Ok(())
     }
+
+    fn finalize_op<Obs: OpObserver>(
+        &mut self,
+        doc: &mut Automerge,
+        op_observer: &mut Obs,
```
alexjg, line 438: If we make this an `Option<&mut Obs>` then I think we can skip a lot of work in this method when there is no observer right?

automerge/src/op_observer.rs
```diff
@@ -1,50 +1,105 @@
 use crate::exid::ExId;
+use crate::Parents;
 use crate::Prop;
 use crate::Value;
 
 /// An observer of operations applied to the document.
-pub trait OpObserver {
+pub trait OpObserver: Default + Clone {
```
alexjg, line 7: I don't think these constraints should be here, we should add the constriants where we need them, i.e. in `impl` blocks.

automerge/src/op_observer.rs
```diff
@@ -1,50 +1,105 @@
...
     /// A new value has been inserted into the given object.
     ///
     /// - `objid`: the object that has been inserted into.
     /// - `index`: the index the new value has been inserted at.
     /// - `tagged_value`: the value that has been inserted and the id of the operation that did the
     /// insert.
-    fn insert(&mut self, objid: ExId, index: usize, tagged_value: (Value<'_>, ExId));
+    fn insert(
```
alexjg, line 14: The docs on these methods all need updating to match the new parameters.

automerge/src/op_observer.rs
```diff
@@ -1,50 +1,105 @@
...
     /// - `objid`: the object that has been deleted in.
-    /// - `key`: the key of the value that has been deleted.
-    fn delete(&mut self, objid: ExId, key: Prop);
+    /// - `prop`: the prop of the value that has been deleted.
+    fn delete(&mut self, parents: Parents<'_>, objid: ExId, prop: Prop);
+
+    /// Merge data with an other observer
+    ///
+    /// - `other`: Another Op Observer of the same type
+    fn merge(&mut self, other: &Self);
```
alexjg, line 61: It's not clear from the docs when an observer would be merged or branched. I think it's worth adding a few sentences to explain when this would happen for both this method and `branch`.

automerge-wasm/src/interop.rs
orionz: I clean this up in the pojo_js pull request

PR #428, opened 2022-09-07
query::Prop: don't scan past end of OpTree

### LINKED ISSUE #400: Error: recursive use of an object detected which would lead to unsafe aliasing in rust
If I load [this doc](https://github.com/automerge/automerge-rs/files/9078308/unsafe_aliasing_doc.gz) and mutate it in this specific way, I run into an error:

```javascript
const Automerge = require('automerge-js');
Automerge.use(require('automerge-wasm'));

const doc = Automerge.load(require('fs').readFileSync('unsafe_aliasing_doc'));

Automerge.change(doc, (doc) => {
  doc.versions['c29cccdee60b4803a29b77b3dabdebb8'].fillChoicesBySlot['7,5,down,5'].rejected['test'] = true
});

// Uncaught:
// Error: recursive use of an object detected which would lead to unsafe aliasing in rust
//     at module.exports.__wbindgen_throw (node_modules/automerge-wasm/nodejs/bindgen.js:1547:11)
//     at wasm://wasm/00346bd2:wasm-function[1480]:0xc275c
//     at wasm://wasm/00346bd2:wasm-function[1482]:0xc2778
//     at wasm://wasm/00346bd2:wasm-function[1277]:0xc0e6d
//     at Automerge.rollback (node_modules/automerge-wasm/nodejs/bindgen.js:571:24)
//     at _change (node_modules/automerge-js/dist/cjs/index.js:112:15)
//     at Object.change (node_modules/automerge-js/dist/cjs/index.js:62:16)
```

The doc was created with `automerge-js` 0.1.3 and `automerge-wasm` 0.1.5, but the repro also works on `automerge-js` 0.1.5—my actual app is still on 0.1.3 for now because of #399. Loading the doc and applying this change with upstream Automerge appear to work fine.

alexjg:
The logic in `query::Prop` works by first doing a binary search in the OpTree for the node where the key we are looking for starts, and then proceeding from this point forwards skipping over nodes which contain only invisible ops. This logic was incorrect if the start index returned by the binary search was in the last child of the optree and the last child only contains invisible ops. In this case the index returned by the query would be greater than the length of the optree.

Clamp the index returned by the query to the total length of the opset.

Fixes #400

### DIFF
--- a/automerge/src/query/prop.rs
+++ b/automerge/src/query/prop.rs
@@ -9,7 +9,15 @@ pub(crate) struct Prop<'a> {
     pub(crate) ops: Vec<&'a Op>,
     pub(crate) ops_pos: Vec<usize>,
     pub(crate) pos: usize,
-    start: Option<usize>,
+    start: Option<Start>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+struct Start {
+    /// The index to start searching for in the optree
+    idx: usize,
+    /// The total length of the optree
+    optree_len: usize,
 }
 
 impl<'a> Prop<'a> {
@@ -30,12 +38,21 @@ impl<'a> TreeQuery<'a> for Prop<'a> {
         child: &'a OpTreeNode,
         m: &OpSetMetadata,
     ) -> QueryResult {
-        if let Some(start) = self.start {
+        if let Some(Start {
+            idx: start,
+            optree_len,
+        }) = self.start
+        {
             if self.pos + child.len() >= start {
                 // skip empty nodes
                 if child.index.visible_len() == 0 {
-                    self.pos += child.len();
-                    QueryResult::Next
+                    if self.pos + child.len() >= optree_len {
+                        self.pos = optree_len;
+                        QueryResult::Finish
+                    } else {
+                        self.pos += child.len();
+                        QueryResult::Next
+                    }
                 } else {
                     QueryResult::Descend
                 }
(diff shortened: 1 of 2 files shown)


automerge/tests/test.rs
```diff
@@ -1332,3 +1332,13 @@ fn load_incremental_with_corrupted_tail() {
...
     );
 }
+
+#[test]
+fn insert_after_many_deletes() {
+    let mut doc = AutoCommit::new();
+    let obj = doc.put_object(&ROOT, "object", ObjType::Map).unwrap();
+    for i in 0..100 {
+        doc.put(&obj, format!("{}", i), i).unwrap();
+        doc.delete(&obj, format!("{}", i)).unwrap();
```
jeffa5, line 1342: Could use `to_string`

automerge/src/query/prop.rs
```diff
@@ -30,12 +38,21 @@ impl<'a> TreeQuery<'a> for Prop<'a> {
...
+            idx: start,
+            optree_len,
+        }) = self.start
+        {
             if self.pos + child.len() >= start {
                 // skip empty nodes
                 if child.index.visible_len() == 0 {
-                    self.pos += child.len();
-                    QueryResult::Next
+                    if self.pos + child.len() > optree_len - 1 {
```
jeffa5, line 49:
```suggestion
                    if self.pos + child.len() >= optree_len {
```

jeffa5 (approved):
Just some style things but looks good

PR #434, opened 2022-09-21
Better js package

alexjg:
This is a shot at revamping the javascript packagin story to get it ready for a preview release which will replace the current `automerge` NPM package.

The primary change here is to package for wasm-bindgens `bundler` target rather than the `web` target when distributing code for browsers. This allows us to do away with the async `init` function which `automerge-wasm` exposed - that now being automatically handled by the bundler - which in turn allows us to directly depend on `automerge-wasm` package in the `automerge` package and consequntly not require any initialization steps on the part of users. This seems like such a huge benefit that it is worth requiring everyone to use a bundler.

This PR includes an example of bundling with Webpack 5 and Vite. Unfortunately create-react-app doesn't work because there's no way to tell CRA that we want to use Webpack's experimental webassembly support.

In order to effectively test all this I've added a tool in `automerge-js/e2e` which publishes the various packages to a local NPM repository and then builds against that. This is used in the `scripts/ci/js_tests` and `scripts/ci/wasm_tests` scripts to run the tests against the version of the code in the repository even if that hasn't been published to NPM yet. This script  also includes a way of running the local NPM server so that other projects on the same machine can point at it if for some reason you need to do that - such as debugging without having to edit three levels of transitive dependencies.


pvh:
I'll pull & try it today but isn't CRA still a pretty important thing to support?

alexjg:
Right, this supports `create-react-app` via a bunch of awful hacks - explained in `automerge-js/examples/create-react-app`. I've also renamed the JS version to `2.0.0-alpha.1`. I think we should merge this now and then push a JS alpha release. Please object in the next few hours or forever hold your peace.

PR #435, opened 2022-09-23
add python wrapper

julienfr112:


julienfr112:
Hi Alex. Thanks for you nice response. I completely understand your point of view. I'm sorry but won't have that much time to work on your points. This is still a POC for future python wrapper. IMHO, a python wrapper is a must-have for automerge success. Woud it be possible to have other people involved ?

Just a small question : what do you mean by UI thread ? Something with Kivi or pyqt ? The use case I had in mind was more server side for a web app and persistence / right management.

PR #437, opened 2022-10-04
New Patch System and Incrementally Updated Js Object for `automerge-js`

orionz:
This PR has a few parts

1. In order to implement the incrementally updated js object I needed object paths instead of just obj ids.  Without this I would need to maintain a elemid <-> index lookup table in memory per list and as we saw in automerge 0.14 this is about as computationally expensive as the whole of automerge.
2. For the paths to be correct they need to be computed at the moment of op insert - not at commit time.  This is because an op after the one inserting could insert or delete to a parent list changing the list index.
3. To make this work I had to move the op observer from the commit call to the transaction itself.  I'm mostly but not tally happy with the current implementation especially because there is some confusion between OpObserver being implemented by `()` and `Option<&mut OpObserver>`.
4. The current WASM api (`materialize()`, `registerDatatype()`, `popPatches()`, `enablePatches()`) is a work in progress and not the final API but are sufficient to meet `automerge-js`s needs.
5. AutomergeJS now incrementally updates an in memory representation of the automerge document.  The ability to observer patches is back.  All of this work is done by rust in the `automerge-wasm` `applyPatches()` function.


alexjg:
I rebased this and merged it manualley

PR #439, opened 2022-10-08
Add TypeScript type for PatchCallback

alexjg:


pvh:
This looks good but as a general style nit I'd encourage you to use `patchCallback?: Foo` instead of `patchCallback: Foo | undefined`. As far as I know this is completely equivalent.

PR #440, opened 2022-10-11
Repo reorg

alexjg:
We're moving towards this repository becoming the primary collaboration point for automerge. i think it's worth making sure the repo layout and README point people towards the things they care about. As such I've done the following things:

* Moved all the rust code into a `./crates` subdirectory. This is idiomatic in Rust codebases and keeps the crates out of the way of the top level organisation of the repo.
* Moved the JS code into `wrappers/automerge-js`. The introduction of the `wrappers` subdirectory a) provides a place for people to put in PRs for new wrappers and b) describes the role of the JS (and other wrapper) codebases in this project.
* Updated the README to try and describe in more detail what this project is, how it is organised, and where to find other resources.


orionz:
Maybe we should do "wrappers/javascript" and "wrappers/wasm"

alexjg:
Good shout. I've renamed `wrappers/automerge-js` to `wrappers/javascript`. As to `automerge-wasm` and `automerge-c`. I've actually left them in `crates/` for now as they are both primarily Rust codebases. This is a bit messy but there isn't a neat way to split these things up unfortunately as `wasm-bindgen` and `cbindgen` are designed to have a single directory with mixed Rust + WASM/C code in them.

PR #444, opened 2022-10-15
make freeze work recursively

orionz:
Enhancing the JavaScript wrapper to respect the `{ freeze: true }` flag recursively and behave properly on `load()` and `clone()`.

### DIFF
--- a/javascript/examples/vite/README.md
+++ b/javascript/examples/vite/README.md
@@ -21,6 +21,13 @@ import topLevelAwait from "vite-plugin-top-level-await"
 
 export default defineConfig({
     plugins: [topLevelAwait(), wasm()],
+    
+    // This is only necessary if you are using `SharedWorker` or `WebWorker`, as 
+    // documented in https://vitejs.dev/guide/features.html#import-with-constructors
+    worker: { 
+        format: "es",
+        plugins: [topLevelAwait(), wasm()] 
+    },
 
     optimizeDeps: {
         // This is necessary because otherwise `vite dev` includes two separate
--- a/javascript/examples/vite/vite.config.js
+++ b/javascript/examples/vite/vite.config.js
@@ -5,6 +5,13 @@ import topLevelAwait from "vite-plugin-top-level-await"
 export default defineConfig({
     plugins: [topLevelAwait(), wasm()],
 
+    // This is only necessary if you are using `SharedWorker` or `WebWorker`, as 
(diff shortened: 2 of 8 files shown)


crates/automerge-wasm/src/interop.rs
```diff
@@ -370,20 +370,23 @@ impl Automerge {
         datatype: Datatype,
         heads: Option<&Vec<ChangeHash>>,
         meta: &JsValue,
+        freeze: bool,
```
alexjg, line 373: If I understand correctly the purpose of this `freeze` argument is to track whether the object being patched (`nextDoc` in `progressDocument` should be `Object.freeze`ed after the patch is applied. Currently this only seems to be the case if the `Automerge` document was initialized with `freeze: true`, would it be possible then to just track the freeze state via a field on `automerge_wasm::Automerg` (e.g. `automerge_wasm::Automerge::freeze`)? That way we wouldn't need to feed the `freeze` argument through to every one of these functions.

wrappers/javascript/test/basic_test.ts
```diff
@@ -21,6 +21,50 @@ describe('Automerge', () => {
...
+            let doc1 = Automerge.from({ sub: { object: true } })
+            assert(Automerge.isAutomerge(doc1))
+            assert(!Automerge.isAutomerge(doc1.sub))
+            assert(!Automerge.isAutomerge("String"))
+            assert(!Automerge.isAutomerge({ sub: { object: true }}))
+            assert(!Automerge.isAutomerge(undefined))
+        })
+
+        it('it should recursively freeze the document if requested', () => {
+            let doc1 = Automerge.init({ freeze: true } )
```
alexjg, line 34: This feels like an API that will be surprising to javascript programmers, is it possible to hook into `Object.freeze` at all so folks can just do:

```
let doc1 = Automerge.init()
Object.freeze(doc1)
```

crates/automerge-wasm/src/interop.rs
orionz: Ok - good call - I added an `enableFreeze()` method which I don't love but it shadows the `enabledPatches()` which is also a stand-in for a better API later.  We can fix it up at a later date.

wrappers/javascript/test/basic_test.ts
orionz: Ok.  Fixed.

orionz:
These failed test seem to be coming from main and not this branch.  Going to merge but we need to get those errors fixed asap.

PR #448, opened 2022-10-27
Allow changing a cloned document

alexjg:
The logic for `clone` which was updated to support cloning a viewed document inadverantly left the heads of the cloned document state in place, which meant that cloned documents could not be `change`d. Set state.heads to undefined when cloning to allow changing them.


orionz:
Good catch.   All I ask is adding a 

`assert.deepEqual(fork, {key: "value"})`

After the clone and before the change to confirm it cloned at the right point.

alexjg:
Done!

PR #449, opened 2022-11-01
Update rust uuid dependency to v1.2.1

tosti007:
This PR updates the rust uuid dependency to the latest version `1.2.1`. [According to the docs](https://docs.rs/uuid/1.2.1/uuid/index.html#webassembly) WebAssembly is now enabled through the `js` feature rather than `wasm-bindgen`.

All rust crates seem to compile just fine and passes all tests (using `--workspace --all-features`) without any further changes, but maybe someone with a bit more knowledge should try it as well (also maybe for the js side?).


MarijnS95 (approved):

alexjg:
Thanks for this!

PR #454, opened 2022-11-07
Add AMobjObjType()

jkankiewicz:
@orionz, @alexjg and @ConradIrwin, I've done the following to ease writing language bindings based upon the C API:
* Exposed `autocommit::AutoCommit::object_type()` as `AMobjObjType()`.
* Replaced the phrase "Lamport timestamp" with "epoch timestamp" within the documentation.

### DIFF
--- a/rust/automerge-c/src/doc.rs
+++ b/rust/automerge-c/src/doc.rs
@@ -143,11 +143,11 @@ pub unsafe extern "C" fn AMcreate(actor_id: *const AMactorId) -> *mut AMresult {
 
 /// \memberof AMdoc
 /// \brief Commits the current operations on a document with an optional
-///        message and/or time override as seconds since the epoch.
+///        message and/or *nix timestamp (milliseconds).
 ///
 /// \param[in,out] doc A pointer to an `AMdoc` struct.
 /// \param[in] message A UTF-8 string or `NULL`.
-/// \param[in] time A pointer to a `time_t` value or `NULL`.
+/// \param[in] timestamp A pointer to a 64-bit integer or `NULL`.
 /// \return A pointer to an `AMresult` struct containing an `AMchangeHashes`
 ///         with one element.
 /// \pre \p doc `!= NULL`.
@@ -160,15 +160,15 @@ pub unsafe extern "C" fn AMcreate(actor_id: *const AMactorId) -> *mut AMresult {
 pub unsafe extern "C" fn AMcommit(
     doc: *mut AMdoc,
     message: *const c_char,
-    time: *const libc::time_t,
+    timestamp: *const i64,
 ) -> *mut AMresult {
     let doc = to_doc_mut!(doc);
     let mut options = CommitOptions::default();
     if !message.is_null() {
         options.set_message(to_str(message));
     }
-    if let Some(time) = time.as_ref() {
-        options.set_time(*time);
+    if let Some(timestamp) = timestamp.as_ref() {
+        options.set_time(*timestamp);
     }
     to_result(doc.commit_with(options))
 }
--- a/rust/automerge-c/src/doc.rs
+++ b/rust/automerge-c/src/doc.rs
@@ -5,7 +5,7 @@ use std::os::raw::c_char;
 
 use crate::actor_id::AMactorId;
 use crate::change_hashes::AMchangeHashes;
-use crate::obj::AMobjId;
+use crate::obj::{AMobjId, AMobjType};
 use crate::result::{to_result, AMresult, AMvalue};
 use crate::sync::{to_sync_message, AMsyncMessage, AMsyncState};
 
@@ -543,6 +543,31 @@ pub unsafe extern "C" fn AMobjSize(
         }
     } else {
         0
+    }
+}
+
+/// \memberof AMdoc
+/// \brief Gets the type of an object.
+///
+/// \param[in] doc A pointer to an `AMdoc` struct.
+/// \param[in] obj_id A pointer to an `AMobjId` struct or `AM_ROOT`.
+/// \return An `AMobjType`.
+/// \pre \p doc `!= NULL`.
+/// \internal
+///
+/// # Safety
+/// doc must be a valid pointer to an AMdoc
+/// obj_id must be a valid pointer to an AMobjId or std::ptr::null()
+#[no_mangle]
+pub unsafe extern "C" fn AMobjObjType(doc: *const AMdoc, obj_id: *const AMobjId) -> AMobjType {
+    if let Some(doc) = doc.as_ref() {
+        let obj_id = to_obj_id!(obj_id);
+        match doc.object_type(obj_id) {
+            None => AMobjType::Unknown,
+            Some(obj_type) => obj_type.into(),
+        }
+    } else {
+        AMobjType::Unknown
     }
 }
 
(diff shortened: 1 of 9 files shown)


rust/automerge-c/src/doc/list.rs
```diff
@@ -486,7 +486,7 @@ pub unsafe extern "C" fn AMlistPutStr(
 }
 
 /// \memberof AMdoc
-/// \brief Puts a Lamport timestamp as the value at an index in a list object.
+/// \brief Puts an epoch timestamp as the value at an index in a list object.
```
alexjg, line 489: We should probably specify "unix epoch" here.

rust/automerge-c/src/doc/map.rs
```diff
@@ -373,7 +373,7 @@ pub unsafe extern "C" fn AMmapPutStr(
 }
 
 /// \memberof AMdoc
-/// \brief Puts a Lamport timestamp as the value of a key in a map object.
+/// \brief Puts a epoch timestamp as the value of a key in a map object.
```
alexjg, line 376: As above, I think this should be "unix epoch"

rust/automerge-c/src/obj.rs
```diff
@@ -142,20 +142,35 @@ pub unsafe extern "C" fn AMobjIdIndex(obj_id: *const AMobjId) -> usize {
 /// \brief The type of an object value.
 #[repr(u8)]
 pub enum AMobjType {
+    /// An unknown type.
+    /// \note This tag is unalphabetized to evaluate as false.
+    Unknown = 0,
```
alexjg, line 147: As noted above, it might be better to remove this variant.

rust/automerge-c/src/obj.rs
```diff
@@ -142,20 +142,35 @@ pub unsafe extern "C" fn AMobjIdIndex(obj_id: *const AMobjId) -> usize {
...
     /// A list of Unicode graphemes.
     Text,
 }
 
+impl From<am::ObjType> for AMobjType {
+    fn from(o: am::ObjType) -> Self {
+        match o {
+            am::ObjType::Map => AMobjType::Map,
+            am::ObjType::List => AMobjType::List,
+            am::ObjType::Table => todo!(),
```
alexjg, line 161: You _could_ return `AMObjecType::Map` here, they're functionally the same and I think we should remove `am::ObjType::Table`.

rust/automerge-c/src/doc.rs
```diff
@@ -546,6 +546,31 @@ pub unsafe extern "C" fn AMobjSize(
...
+///
+/// # Safety
+/// doc must be a valid pointer to an AMdoc
+/// obj_id must be a valid pointer to an AMobjId or std::ptr::null()
+#[no_mangle]
+pub unsafe extern "C" fn AMobjObjType(doc: *const AMdoc, obj_id: *const AMobjId) -> AMobjType {
+    if let Some(doc) = doc.as_ref() {
+        let obj_id = to_obj_id!(obj_id);
+        match doc.object_type(obj_id) {
+            None => AMobjType::Unknown,
```
alexjg, line 566: I think it might make more sense to return an optional `AMObjType` (I'm not sure how one does that idiomatically in modern C, just return a `NULL` pointer?). `automerge::Automerge::object_type` only returns `None` if whatever `obj_id` points at is not an object (either the `OpId` it wraps is not in the document, or it is a scalar value). It's not really unknown in any sense.

rust/automerge-c/src/obj.rs
```diff
@@ -142,20 +142,35 @@ pub unsafe extern "C" fn AMobjIdIndex(obj_id: *const AMobjId) -> usize {
...
+    }
+}
+
 impl From<AMobjType> for am::ObjType {
     fn from(o: AMobjType) -> Self {
         match o {
             AMobjType::Map => am::ObjType::Map,
             AMobjType::List => am::ObjType::List,
             AMobjType::Text => am::ObjType::Text,
+            AMobjType::Unknown => todo!(),
```
alexjg, line 173: Removing the `AMObjType::Uknown` as mentioned earlier would also fix this todo :slightly_smiling_face:

rust/automerge-c/src/result.rs
```diff
@@ -85,7 +85,7 @@ use crate::sync::{AMsyncMessage, AMsyncState};
 /// The variant discriminator.
 ///
 /// \var AMvalue::timestamp
-/// A Lamport timestamp.
+/// An epoch timestamp.
```
alexjg, line 88: "unix epoch"

rust/automerge-c/src/result.rs
```diff
@@ -133,7 +133,7 @@ pub enum AMvalue<'a> {
     SyncMessage(&'a AMsyncMessage),
     /// A synchronization state variant.
     SyncState(&'a mut AMsyncState),
-    /// A Lamport timestamp variant.
+    /// An epoch timestamp variant.
```
alexjg, line 136: "unix epoch"

alexjg (commented):
Sorry for all the "unix epoch" comments, I just put them in there so it's easy to see which bits to fix if you do decide to change the nomenclature.

rust/automerge-c/src/doc/list.rs
ConradIrwin: and `in milliseconds` please.

```suggestion
/// \brief Puts number of milliseconds since the unix epoch as a timestamp at an index in a list object.
```

rust/automerge-c/src/doc.rs
ConradIrwin: I think a more typical style is what is done here (a sentinel value for unknown/invalid); pointers beget questions about ownership that don't apply in this case. maybe `AMobjType::Invalid` is the right phrasing?

ConradIrwin:
Thank you @jkankiewicz

rust/automerge-c/src/doc.rs
jkankiewicz: > I think it might make more sense to return an optional AMObjType (I'm not sure how one does that idiomatically in modern C, just return a NULL pointer?). 

@alexjg, yes, it would have to be NULL but then how would one free the storage for the `AMObjType` variant?
A variant doesn't complicate memory management like that.

alexjg: Yeah if it's idiomatic to return a sentinel value then maybe we just change the name? `Unknown` to me implies that there was an object there, we just didn't know what it was, when in fact we do know that there wasn't an object there.

rust/automerge-c/src/obj.rs
```diff
@@ -170,7 +169,7 @@ impl From<AMobjType> for am::ObjType {
             AMobjType::Map => am::ObjType::Map,
             AMobjType::List => am::ObjType::List,
             AMobjType::Text => am::ObjType::Text,
-            AMobjType::Unknown => todo!(),
+            AMobjType::Void => todo!(),
```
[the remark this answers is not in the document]
jkankiewicz: > This particular code path is only used when creating new objects, so it seems to me like we could remove the infallible From implementation and instead use a function which returns AMresult if the Void variant is specified, much like the to_obj_id and to_doc macros do.

@alexjg, you're right that no one should ever specify `AM_OBJ_TYPE_VOID` when calling `AM*PutObj()` and returning an `AMresult` containing an error to that effect is more certain than documenting that constraint via a precondition.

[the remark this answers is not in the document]
jkankiewicz: > [U]se a function which returns AMresult if the Void variant is specified, much like the to_obj_id and to_doc macros do.

@alexjg, I used a macro like the two that you mentioned.

Issue #455, opened 2022-11-08
automerge-c: uncaught panic if string contains a null byte


ConradIrwin:
I used the following typescript to create a doc containing a string with a null byte in it:

```typescript
import * as automerge from '@automerge/automerge' // 2.0.0-beta.4

let doc = automerge.init()

var newDoc = automerge.change(doc, (d: any) => {
  d.oops = 'o\x00ps'
})

let bytes = automerge.save(newDoc)
// => hW9KgzgYCpUAdAEQAlGmcMDRTlKAagbMI3V0owG3hG4vm1xsqt7I1lr4Yc0pMEkeiGUjKJAdUqx8qMyy5AYBAgMCEwIjAkACVgIIFQYhAiMCNAFCAlYCVwSAAQJ/AH8BfwF/AH8Afwd/BG9vcHN/AH8BAX8Bf0ZvAHBzfwAA
```

When I try to load this with automerge-C:

```
result = AMmapGet(doc, AM_ROOT, "oops", nil);
C.AMresultStatus(result) == AM_RESULT_OK;
C.AMerrorMessage(result) == "";
value = AMresultValue(result); // rust panic
```

The backtrace (with RUST_BACKTRACE=full):

```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: NulError(1, [111, 0, 112, 115])', automerge-c/src/result.rs:189:80
stack backtrace:
   0:        0x10059def4 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h1543c132bc4e188c
   1:        0x10056ccc4 - core::fmt::write::hda8e8eb84b49cbfc
   2:        0x1005a1ea8 - std::panicking::default_hook::{{closure}}::hdf06011cb093de6a
   3:        0x1005a271c - std::panicking::rust_panic_with_hook::h053d4067a63a6fcb
   4:        0x1005a2344 - std::panicking::begin_panic_handler::{{closure}}::hea9e6c546a23e8ff
   5:        0x1005a22bc - std::sys_common::backtrace::__rust_end_short_backtrace::hd64e012cf32134c6
   6:        0x1005a2288 - _rust_begin_unwind
   7:        0x1005ac598 - core::panicking::panic_fmt::hbfde5533e1c0592e
   8:        0x1005ac750 - core::result::unwrap_failed::h68832e989a8867c1
[...]

PR #457, opened 2022-11-10
Prevent panic when string contains a null character.

jkankiewicz:
@orionz and @ConradIrwin, I've changed `AMlistGet()` and `AMmapGet()` to return an `AMbyteSpan` instead of a `char const*` when a string value contains a null character in order to resolve issue #455.


rust/automerge-c/src/result.rs
```diff
@@ -184,13 +184,20 @@ impl From<(&am::Value<'_>, &RefCell<Option<CString>>)> for AMvalue<'_> {
...
-                    AMvalue::Str(match c_str.as_mut() {
+                    match c_str.as_mut() {
                         None => {
-                            let value_str = CString::new(smol_str.to_string()).unwrap();
-                            c_str.insert(value_str).as_ptr()
+                            let value_str = smol_str.as_str();
+                            if value_str.find('\0').is_some() {
+                                // A C string can't contain null characters
+                                // so we have to treat the string as bytes.
+                                AMvalue::Bytes(value_str.as_bytes().into())
```
alexjg, line 193: What happens in the following scenario:

- Bob creates an automerge document in JavaScript with a string property (let's call it `"badname"`) which contains a null byte
- Bob saves the document and sends it to Alice
- Alice loads the document using the C library, loads the `"badname"` property and adds a byte to the end
- Alice saves the document and sends it back to Bob
- Bob loads the document in Java-script again

From Bob's perspective the `"badname"` property has now changed from `String` to `Bytes` right?

jkankiewicz: @alexjg, from Alice's perspective, the "badname" value has effectively become read-only in the same sense that a value tagged as being `AM_VALUE_UNKNOWN` is read-only because it cannot be translated into an analogous C type.

As you rightly pointed out, the problem with this approach is that it doesn't communicate the value's translation into an alternate type and, even worse, Alice isn't prevented from editing it either.

It appears that a better approach would be to translate the value into an `AM_VALUE_UNKNOWN` instead because its bytes can still be read through the `bytes` member, its true data type can be communicated through the `type_code` member and the recipient will be unable to edit it.

@alexjg, what value should the `type_code` member be in this situation?

alexjg: Well, the type code would be `6` as specified [here](https://alexjg.github.io/automerge-storage-docs/#raw-value-columns).

However, it feels like the real problem is that we are using `CString` for the type of `AMresult::String` when in fact strings in automerge are UTF8 (which can contain null bytes). Should we consider using just a `Vec<u8>` there instead?

ConradIrwin: I like the idea of representing an AMresult::String using the same structure as an AMbyteSlice (and I share @alexjg's concern with having the type of the value change depending on its content).

If we want to support nulls in map keys too then that will require a major (if mechanical) change to the API to pass a pointer + length instead of a Cstring for `AMmapPut*` too.

The other APIs that ned fixing are `AMmapItemsValue`, `AMlistItemsValue`, `AMobjItemValue` (though I am not currently using these for automerge-go due to the friction around reading the AMvalue union using cgo).

jkankiewicz: > Well, the type code would be 6 as specified [here](https://alexjg.github.io/automerge-storage-docs/#raw-value-columns).

@alexjg, thank you!

> However, it feels like the real problem is that we are using CString for the type of AMresult::String when in fact strings in automerge are UTF8 (which can contain null bytes).

Reading that I'm reminded of Frank Zappa's quote about curing dandruff with decapitation.

I think it's fair to say that %99+ of strings in human-generated content aren't going to contain null bytes because the vast majority of physical and virtual keyboards do not furnish a NUL key; more than one step is required to insert a NUL character into a text document.

Isn't its continued use as a string-terminating sentinel in what would otherwise be just an array of bytes the chief reason for the inclusion of NUL (U+0000) in the Unicode character set? It is, after all, categorized as a control character.

It seems to me that if one wishes to store an array of bytes in an Automerge document then one should do so explicitly as a bytes value and not as a string. Translating an array of bytes that was conveyed as a string into an immutable array of bytes that's identified as a string provides a group of collaborators with an opportunity to negotiate how the value should be stored in future revisions of the document.

jkankiewicz: > The other functions that need fixing are `AMmapItemValue()`, `AMlistItemValue()`, `AMobjItemValue()`...

@ConradIrwin, I believe those functions are covered by the changes I've made already but I should add unit tests to confirm.

ConradIrwin:
> @ConradIrwin, I believe those functions are covered by the changes I've made already but I should add unit tests to confirm.

Sorry I didn't check! I thought they might not because those APIs do not return an `AMresult`.

> Reading that I'm reminded of Frank Zappa's quote about curing dandruff with decapitation.

Agreed! It's definitely not the idiomatic C way to do things.

I mostly want it not to panic, so I would also be happy with the options of:
* automerge errored when setting strings containing U+0000, so this situation can't happen.
* automerge itself works fine with U+0000, but automerge-c reports a sensible error.

As you say, this should be rare in practice.

It's possible we could split the difference: no U+0000 allowed in map keys; and either handle (or error) on string values with U+0000.

rust/automerge-c/src/result.rs
alexjg: I agree that putting a null byte in the middle of a UTF-8 string is a strange thing to do. However, we have specified that string values in automerge are UTF-8, so I think we should gracefully handle users doing such unnatural things. Using `Unknown` might work - we would need to add a lot of documentation to the effect that if you are using the library via FFI from an environment which does allow `U+0000` then you must handle both `Amresult::String` and `Amresult::Unknown`. This seems like it might be quite confusing.

It might well be better to just return an `Amresult::Error` here as @ConradIrwin suggests.

jkankiewicz: > It might well be better to just return an Amresult::Error here as @ConradIrwin suggests.

@alexjg, I suspect that in most cases receiving an `AM_VALUE_UNKNOWN` will manifest as an error anyway because the application's assertion that the `AMvalue.tag` member's value is `AM_VALUE_STR` will have been violated.

Returning a (usually) unexpected value instead of an error gives an organization the option of migrating legacy documents from a schema that forces a bytes field to be stored as a string to a schema where they're explicitly stored as bytes or to remove/substitute NUL characters in string values that were stored prior to the institution of a policy that forbids NUL characters within string values.

ConradIrwin: On the flip side errors have descriptive error messages... it would be quite confusing to get an `AMunknown` and not know why.

alexjg: I still think we should either handle UTF-8 strings throughout, or throw an error at this point. To me having a field which changes type depending on it's contents but which appears the same to the user seems quite confusing.

Note that I don't think we should assume that this is a case of someone attempting to store bytes in a string - it's equally likely that it happens as a bug in some text editor somewhere or as a weird escape sequence in source code or any number of other situations which produce some cursed utf8. In any situation where the user creating the document is working in a utf8 environment (e.g. source code of a program creating automerge documents, or a text editor, or a JavaScript console etc. etc.) the user would have no reason to believe there was anything wrong until the document causes someone loading it with the C library to get an `Unknown` value, this seems like it would be very hard to debug. On the flip side an error saying "this library doesn't handle utf8 with null bytes" is quite a lot clearer.

Imagine in a years time someone comes to us with a bug report from an FFI wrapper of `automerge-c` in Ruby saying "this document keeps telling me there's an unexpected Unknown type" vs "when I load this document it tells me it can't handle null byte utf8", which would be easier to help them debug? :slightly_smiling_face:

jkankiewicz: > On the flip side errors have descriptive error messages... it would be quite confusing to get an AMunknown and not know why.

@ConradIrwin, I think there's only one reason why an `AM_VALUE_UNKNOWN` was returned instead of the expected `AM_VALUE_STR`: there's more than one legitimate way to convert—or decline to convert—the underlying UTF-8 string into a `char const*` and the automerge-c library is refusing the temptation to guess.

As you've pointed out, some higher-level languages may not be able to access the members of the `AMvalue` union directly so automerge-c should provide optional functions for accessing those members as if `AMvalue` were an opaque type.

The `AMvalueStr()` function should take a callback function argument so that an application can specify on a case-by-case basis how a returned `AM_VALUE_UNKNOWN` should be handled either by conversion or error reporting. The function should also take a `void*` in/out argument so that state can be passed through to the callback function as well e.g. an error-reporting callback function can update an application-level error variable for languages that don't support exceptions.

For an organization that forbids NUL characters from being inserted into a string value, `AMvalue.str` can be accessed directly but for an organization that can't be as certain then they should call `AMvalueStr()` with an appropriate callback function pointer and state variable pointer instead.

alexjg: The reason for the `unknown` type code is to allow for us to add types in future and still allow older implementations to work with newer files containing the new types. This means that there is another reason that someone might get `AM_VALUE_UNKNOWN` instead of `AM_VALUE_STR`: if someone with a new version of automerge set the given property to a new type.  I suppose that could still be handled with the approach you're suggesting by switching on the `type_code`?

jkankiewicz: > This means that there is another reason that someone might get AM_VALUE_UNKNOWN instead of AM_VALUE_STR: if someone with a new version of automerge set the given property to a new type. I suppose that could still be handled with the approach you're suggesting by switching on the type_code?

@alexjg, yes, the callback function can use the `AMunknownValue.type_code` member to disambiguate in that circumstance.

All of the remaining `AMvalue*()` access functions should resemble the `AMvalueStr()` function that I've proposed so that they, too, can be flexibly configured to coerce unknown types into known ones as appropriate.

ConradIrwin: Back to where we started, should AMvalueStr() just always return an AMbyteSpan?

Almost all languages internally represent a string as a pointer + a length to avoid the obvious problems with the C representation (and it would be no harder for automerge-go to use this representation to convert to a Go string).

That way anyone who is using automerge-c directly will have to deal with it; but anyone who's just using it as a common translation layer will not have to worry. I suspect (though I have an obvious bias :D) there will be more users using automerge-c as FFI not directly.

jkankiewicz: > Back to where we started, should AMvalueStr() just always return an AMbyteSpan?

@ConradIrwin, after more consideration this is the approach towards which I'm leaning due to its simplicity.

For me, anything is preferable to the error approach because it would mean that applications using automerge-c wouldn't be able to process every Automerge document.

> I suspect (though I have an obvious bias :D) there will be more users using automerge-c as FFI not directly.

You're probably right about that.

alexjg: I like the string as `AMbyteSpan` approach. There is still one consideration which is that if you put non utf8 bytes in a string field then  it will not be loadable as the decoding code checks that the bytes are valid utf8. This won't result in unloadable documents because `automerge_c::doc::utils::to_str` would null out the string before putting it in the document, but it would be a little surprising. We would probably want to modify `AMmapPutStr` to return an error for non utf8 bytes in a string.

alexjg: Sorry that's slightly incorrect, `to_str` would currently call `Cstring::to_string_lossy`. The equivalent for rust strings would be `String::from_utf8_lossy`.

ConradIrwin: makes sense!

I think this is a great solution for values, and I like the utf8 check (similar to rust go's strings are utf-8 "by convention"). 

What do we want to do about map keys?
* Would it be reasonable to check for and disallow `U+0000` in map keys? If so, that seems the easiest for everyone.
* If not... should we make the same change so that map keys are represented as `AMbyteSpan` too; or just error?

jkankiewicz: > If not... should we make the same change so that map keys are represented as `AMbyteSpan` too; or just error?

@ConradIrwin, I think that string values should be handled consistently and that an application based on automerge-c should be able to process  an Automerge document with NUL characters in its map keys, too.

alexjg:
Can you rebase on `main`, that should fix the failing JS test.

jkankiewicz:
> Can you rebase on main, that should fix the failing JS test.

@alexjg, it has been done but the "js_tests" check is still failing but for a different, seemingly-unrelated problem this time.

The "windows" check failed because of insufficient disk space.

rust/automerge-c/src/result.rs
```diff
@@ -749,17 +737,17 @@ pub enum AMstatus {
...
-/// \return A UTF-8 string value or `NULL`.
+/// \return A UTF-8 string view as an `AMbyteSpan` struct.
 /// \pre \p result `!= NULL`.
 /// \internal
 ///
 /// # Safety
 /// result must be a valid pointer to an AMresult
 #[no_mangle]
-pub unsafe extern "C" fn AMerrorMessage(result: *const AMresult) -> *const c_char {
+pub unsafe extern "C" fn AMerrorMessage(result: *const AMresult) -> AMbyteSpan {
```
ConradIrwin, line 747: @jkankiewicz is it possible to leave this as a `*const c_char` - as the authors of automerge we can probably ensure that error messages don't contain null bytes; and that does make error handling a bit more awkward for C users, for no real benefit to wrapper languages.

PR #459, opened 2022-11-18
Added deno as a target for automerge-wasm

acurrieclark:
This adds a deno output for the `@automerge/automerge-wasm` package.

As Deno doesn't have a package manager, it imports packages directly from URLs. So the question remains as to how this Deno version might best be exposed and similarly how it will be best included in the (to be created) deno version of `@automerge/automerge`.

As a way to start, this should add the `deno` directory to the version published on npm, which should then be importable using one of the many CDNs. In an ideal world, it would be published to https://deno.land/x, but I am not entirely sure how that would fit in with your current publishing setup.

I have tested the updated scripts on my local machine, and the compiled output is produced as expected.


alexjg:
Two things:

1) Can you `rebase` on `main`, I believe the failing JS test is a typescript error I fixed earlier this week.
2) It's nice that it's so easy to get a deno target building. It would be nice to test that it actually imports and runs in `deno` somehow. Ideally we would run something like the `scripts/ci/js_tests` script but using Deno rather than node, but I think it would also be fine just to have a smoke test checking that we can import the package.

.github/workflows/advisory-cron.yaml
```diff
@@ -4,6 +4,7 @@ on:
     - cron: '0 18 * * *'
 jobs:
   cargo-deny:
+    if: ${{ false }}
```
alexjg, line 7: What's the reason for all these conditions? Is this just to check CI runs for deno and then you'll remove them before we merge?

acurrieclark: Exactly, just wanted to disable any other CI jobs temporarily. I did a test run already and found that the deno test was also being run by the `yarn test` command.

Will reverse these once I can see that the CI runs properly.

gavriguy:
@acurrieclark Thanks, I was looking for an automerge implementation for deno.

Can you, by any chance, share how one can add it to a deno file?

acurrieclark:
@gavriguy I am afraid for now I have just been manually porting the files to Deno. Essentially, I copied the deno wasm files from one of the [cdns](https://www.jsdelivr.com/package/npm/@automerge/automerge-wasm) and the typescript files from the [javascript directory](https://github.com/automerge/automerge-rs/tree/main/javascript/src) of this repo.

Then I made sure that the wasm imports in typescript were pointing to the wasm copy I had made, and added a bunch of:

```typescript
// @ts-nocheck
```

comments to the top of files to stop deno throwing typescript errors.

I know that there are plans for first party deno support, and I will be looking at how I can help achieve that from next week.

gavriguy:
Thank you very much for the answer. 
Unfortunately i don't have the knowledge to help, but I'd be happy to be a test user :)

acurrieclark:
@gavriguy there is a now a deno release for Automerge on deno.land. You can find it at https://deno.land/x/automerge

gavriguy:
Thank you. Looking forward to give it a try

PR #463, opened 2022-11-30
This test shows that the get_changes() function returns several empty changes for various transaction scenarios, currently believed to be in error.

issackelly:


issackelly:
I think I should mention that the behavior is the same if there is a commit message or if that is Null

jeffa5:
I believe that empty commits are an intended feature: see https://github.com/automerge/automerge-rs/blob/aaddb3c9ea84960924f6c68ca54f0af33a0b31dd/javascript/src/index.ts#L350-L361

alexjg:
It is valuable to be able to generate empty changes for cases like the docs mentioned above. However, I think this should probably always be an explicit user action. I'm not sure what a good API for this is. We can obviously just modify the return of `Transaction::commit` to return an `Option<ChangeHash>` - I'm having a play with that.

issackelly:
I was thinking a flag on CommitOpts might be appropriate.

alexjg:
I actually think this might be something a bit different to a flag on `CommitOptions`. I think the only scenario when you _want_ to create an empty commit is when you are creating a merge commit. In this case you _know_ that the `ChangeHash` will exist. This suggests to me that we want a separate method on `Automerge` and `Autocommit` to create an empty change and return the hash of the change. Then `commit` can just return an `Option<ChangeHash>` and always return `None` if there was nothing to do. This means the semantics of the return value of `commit` are easier to reason about because you don't have to also know what the `CommitOptions` were to understand what `Some` means.

PR #465, opened 2022-11-30
Improve docs for building automerge-c on a mac

ConradIrwin:
I've added a bunch of documentation on setting up the build environment (which was challenging to me as a rust newbie), and some more detailed documentation for automerge-c (to clarify expectations around concurrency, types, strings with null characters, etc.).

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,11 +40,11 @@ in that time.
 
 In general we try and respect semver.
 
-### JavaScript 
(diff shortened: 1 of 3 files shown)


alexjg:
This looks great to me, @jkankiewicz what do you think?

ConradIrwin:
Thanks @alexjg!

PR #467, opened 2022-11-30
Transaction empty

alexjg:
Transactions with no ops in them are generally undesirable. They take up space in the change log but do nothing else. They are not useless though, it may occasionally be necessary to create an empty change in order to list all the current heads of the document as dependents of the empty change.
    
The current API makes no distinction between empty changes and non-empty changes. If the user calls `Transaction::commit` a change is created regardless of whether there are ops to commit. To provide a more useful API modify `commit` so that if there is a no-op transaction then no changes are created, but provide explicit methods to create an empty change via `Transaction::empty_change`, `Automerge::empty_change` and `Autocommit::empty_change`. Also make these APIs available in Javascript and C.

@orionz  it would be great if you could check that the JS bindings make sense
@jkankiewicz I had to modify the C code a bit, please check I've not done anything silly

Closes #463

### DIFF
--- a/javascript/src/index.ts
+++ b/javascript/src/index.ts
@@ -301,7 +301,10 @@ export function change<T>(doc: Doc<T>, options: string | ChangeOptions<T> | Chan
     }
 }
 
-function progressDocument<T>(doc: Doc<T>, heads: Heads, callback?: PatchCallback<T>): Doc<T> {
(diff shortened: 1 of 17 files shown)


jkankiewicz (approved):
@alexjg, your changes look good, I just have a few suggestions.

PR #470, opened 2022-12-02
wasm: improve error messages

### LINKED ISSUE #469: Uncaught exception when applying a client-side change
When applying a change from another client that has been relayed over a node.js server through a web socket, I get an Uncaught exception from auto merge_wasm_bg.js:1102.

Here's a small node.js project that replicates the error.
```
npm install
./node_modules/webpack/bin/webpack.js
node.js index.js
```

Now open two browser tabs to localhost:8888, open the console, and click the 'Do Change' button.

[automerge-bug.zip](https://github.com/automerge/automerge-rs/files/10132384/automerge-bug.zip)

alexjg:
The error messages produced by various conversions in `automerge-wasm` were quite uninformative - often consisting of just returning the offending value with no description of the problem. The logic of these error messages was often hard to trace due to the use of `JsValue` to represent both error conditions and valid values - evidenced by most of the public functions of `automerge-wasm` having return types of `Result<JsValue, JsValue>`. Change these return types to mention specific errors, thus enlisting the compilers help in ensuring that specific error messages are emitted.

There's quite a bit of code here but it's all pretty mechanical really, just creating an error type for every place where we used to return a `JsValue` as an error.

Fixes #469


rust/automerge-wasm/src/value.rs
```diff
@@ -149,3 +147,19 @@ impl From<Datatype> for JsValue {
...
 }
+
+#[derive(Debug, thiserror::Error)]
+pub enum InvalidDatatype {
+    #[error("unknown datatype")]
+    Unknown(String),
+    #[error("datatype is not a string")]
+    NotString,
+    #[error("cannot handle unknown datatype")]
+    UnknownNotImmplemented,
```
jkankiewicz, line 158: @alexjg, typo.
```suggestion
    UnknownNotImplemented,
```

rust/automerge-wasm/src/value.rs
```diff
@@ -135,9 +132,10 @@ impl TryFrom<JsValue> for Datatype {
             "null" => Ok(Datatype::Null),
             d => {
                 if d.starts_with("unknown") {
-                    todo!() // handle "unknown{}",
+                    // TODO: handle "unknown{}",
+                    Err(InvalidDatatype::UnknownNotImmplemented)
```
jkankiewicz, line 136: @alexjg, typo.
```suggestion
                    Err(InvalidDatatype::UnknownNotImplemented)
```

Issue #451, opened 2022-11-02
automerge-c: Nice to haves for automerge-go


ConradIrwin:
I spent a bit of time trying to make automerge-go, using automerge-c as a basis. I was able to get reading/writing values, and loading/saving working (after #450), but there are a few things that would have made the task easier. When you're ready to focus on the C API, a few things "nice to haves" that came to mind.

* Go doesn’t support union types well, so if AMvalue was an opaque struct and had accessor methods for each type that would have been better (though I’m not sure how to handle someone calling the wrong accessor on the wrong variant; maybe just don’t do that :D). For go I ended up generating a bunch of stub methods.
* When iterating over a map or a list, you can either get a Value or an ObjId. It’d be tidier if the objId was also returned by AMmapItemValue instead of having a separate accessor there (how this is handled in AMmapGet).
* When iterating over a map, if you get an item with an object id, you need to retain the AMresult that contains all the mapItems just to keep the object id pointer valid. It’d be nice if objectid was returned by value instead of reference (but not sure the performance impact of that). 
* (This is also true for methods like AMmapGet that do combine the Value and ObjId into one struct: because all other values can be converted to Go types immediately, the only time that I need to retain the AMresult from AMmapGet is if it returned an object id, which makes calling AMfree at the right time a bit more complicated than I’d like.)
* If you write an int > 2**52, the javascript version crashes on load; it’d be good to either return an error, or fix the js implementation.
* (not sure if a bug or a feature?) AMactorIdBytes() allows creating a 0-length actor id.
* It took me a while to get the build working to generate libautomerge.a and automerge.h (running on Mac, the scripts had a few syntax errors according to my local copy of bash; and I am missing a lot of implicit dependencies). It'd be nice if there was a place I could just download the built versions.
* (not sure if a real problem?) Currently automerge.h contains some structs that say they depend on architecture; it'd be nice (but not essential) if there was one header file that worked across multiple architectures. This may already be the case in practice, but I've only tested on darwin/arm64.

I'm not very familiar with rust, (nor that familiar with C), so there may be reasons for the above choices that I don't understand; and as I said, I was able to build what I wanted so none of this should be considered a bug.

ConradIrwin:
A few more things here as I've made more progress:

[...]

PR #485, opened 2022-12-22
Return an error from automerge-c on empty commit

ConradIrwin:
Ref #451

Before this change it was unclear why commit would sometimes return void.


ConradIrwin:
CC @jkankiewicz

ConradIrwin:
Closing this for now. On reconsideration, it may be better to do something else (like return an empty ChangeHashes), or just leave as is

PR #495, opened 2023-01-12
Prepare js code for deno release

acurrieclark:
This is the start of a PR to get the JS codebase ready for an eventual deno release to match the new `wasm` release.

Would love some feedback on how you would prefer to manage the scripts in `package.json`. eg. would you like a single `test` script which runs both the main js tests and the deno smoke test?

Most of the changes here are to adapt the code to meet Deno's typescript standards. For example, types are explicitly declared as such on imports and exports.

There are a few changes to handle cases of implicit any. The largest job here would be to tackle the `proxies.ts` file, so as a stopgap the deno build currently prepends this file with a `// @ts-nocheck` for the deno output.

The UUID implementation for deno is slightly different, so the `uuid.deno.ts` file accounts for this with a deno specific output.


javascript/tsconfig.json
```diff
@@ -15,5 +15,5 @@
     "outDir": "./dist"
   },
   "include": ["src/**/*", "test/**/*"],
-  "exclude": ["./dist/**/*", "./node_modules"]
+  "exclude": ["./dist/**/*", "./node_modules", "./src/**/*.deno.ts"]
```
alexjg, line 18: Rather than putting this in here and then having an `exclude` in all the `config/*` files could we have another `deno` specific config file like `config/deno.config.ts`?

alexjg: Oh, hmm, maybe I've misunderstood things, this is actually fine :grimacing:

javascript/src/uuid.deno.ts
```diff
@@ -0,0 +1,24 @@
+import * as v4 from "https://deno.land/x/uuid@v0.1.2/mod.ts"
```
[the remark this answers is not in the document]
acurrieclark: Absolutely. Will add now

PR #498, opened 2023-01-15
Add Fuzz Testing

ConradIrwin:
Adds fuzz testing and fixes the issues found

One open question is whether or not to commit the corpus/ directory
(https://github.com/rust-fuzz/cargo-fuzz/issues/194)

I have committed the artifacts/ directory to use as regression tests for the
fixes I added.


alexjg:
The recommendation in the linked thread is to _not_ store these things in the repository, although that recommendation is admittedly not a strongly felt one. In this case we are using the corpus primarily for reproducing test failures so I wonder if it might  be better to explicitly write out the failing testcases rather than using the corpus - this might be easier to debug when those test cases fail as well. Thoughts?

We also have a bunch of development related crates now (this and `edit-trace` ), I think it might be time to add a `HACKING.md` or similar file describing how to run benchmarks and fuzz tests (given that we are not running them in CI). That can probably happen later though.

ConradIrwin:
The corpus itself is not included in this PR (it's in a separate directory with another ~2000 files, and total ~500kb at the moment); just the crashers.

I'm happy to run a script so that the crashers are "whole" and in a separate directory instead of re-using the fuzz output if that would be better (definitely agree it'd be easier to debug if the files contain the headers).

alexjg:
Ah that makes sense, yeah I reckon moving the crashers into their own testcases would be worth doing in this PR.

ConradIrwin:
Done!

Sent via Superhuman ***@***.***>

rust/automerge/src/automerge.rs
```diff
@@ -805,23 +805,27 @@ impl Automerge {
...
         }
         Ok(())
     }
 
-    fn apply_change<Obs: OpObserver>(&mut self, change: Change, observer: &mut Option<&mut Obs>) {
+    fn apply_change<Obs: OpObserver>(
+        &mut self,
+        change: Change,
+        observer: &mut Option<&mut Obs>,
+    ) -> Result<(), AutomergeError> {
```
alexjg, line 826: I don't think this function should return a `Result`. `apply_change` is a private function which is only ever intended to be called with a change which is "causally ready". You can see this in e.g. `apply_change_with`, where we use `is_causally_ready` or `pop_next_causally_ready` to make sure that the change dependencies are available before applying it and if the dependencies are not ready we put the change into `self.queue`.

As you've discovered, these checks were _not_ being made in `load`. I think the solution would be to add these checks there.

ConradIrwin: :+1: I'll remove this fix from this branch to unblock the rest, and resubmit in a few days

PR #499, opened 2023-01-16
Update patch types

acurrieclark:
Added new `InsertPatch` type, which now applies to inserts on `Text` and `List`/`Sequence` elements.
Added new `SpliceTextPatch` type which is produced when text is added to a `TextV2` string element.
Removed `SplicePatch` type, which no longer exists.

### DIFF
--- a/javascript/src/stable.ts
+++ b/javascript/src/stable.ts
@@ -41,7 +41,8 @@ import type {
 export type {
   PutPatch,
   DelPatch,
-  SplicePatch,
+  SpliceTextPatch,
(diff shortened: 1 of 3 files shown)


acurrieclark:
The `deno_tests` ci action fails right now, but should pass once the action has been fixed.

alexjg:
Could you rebase this on `main` to see if the CI failures go away?

acurrieclark:
Done! This obviously introduces a BC by removing the `SplicePatch` type.

PR #507, opened 2023-01-24
bail after 100 patches, im:: clocks

orionz:
Currently I bail on patches when 100 have queued and then just reconstruct the document from scratch.  For sure we need a better heuristic than that but this is a first draft and solves a serious issue. 

Pre-computing a clock for every change when clocks have hundreds of actors was spending a lot of time copying memory around.  I switched to immutable HashMap to share common data and got a very large speedup.   There are a bunch of other low hanging optimizations but I think we need to see if we can figure out how to be performant without pre-computed clocks.

### DIFF
--- a/rust/automerge-wasm/src/lib.rs
+++ b/rust/automerge-wasm/src/lib.rs
@@ -574,6 +574,12 @@ impl Automerge {
         let mut object = object
             .dyn_into::<Object>()
             .map_err(|_| error::ApplyPatch::NotObjectd)?;
+
+        if self.doc.observer().has_overflowed() {
+            self.doc.observer().enable(true);
+            return Ok(self.export_object(&am::ROOT, Datatype::Map, None, &meta)?);
+        }
+
         let patches = self.doc.observer().take_patches();
         let callback = callback.dyn_into::<Function>().ok();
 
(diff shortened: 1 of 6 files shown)


alexjg:
This is nice and gets us _much_  closer to acceptable sync speeds. My main issue is that we have a number of people (e.g. @acurrieclark ) depending on  the patch callbacks to implement various state synchronization mechanisms. I think this change will currently break any client which is depending on the patch callback to notify them of every change that is made. I can think of two things which would fix that:

1. Allow clients to opt out of the overflow behaviour 
2. Add a mechanism to the patch callback which notifies the client that we skipped a bunch of changes. Some clients might be able to use that information to reset state from the source of truth which is the automerge document

acurrieclark:
I am afraid that my understanding of rust and my knowledge of the depths of automerge aren't sufficient to offer comment on the PR directly, but I can confirm that I am using the patch callback to manage Undo/Redo in an application right now. I haven't encountered an example in my usage so far where more than 15-20 patches have been generated by a single change though.

That said, I haven't explored how patches are produced when large numbers of changes are synced yet.

alexjg:
This is addressed in #517

PR #516, opened 2023-02-02
Use our leb128 parser for values

ConradIrwin:
This ensures that values in automerge documents are encoded correctly,
and that no extra data is smuggled in any LEB fields.

I am not sure this is the right approach to error handling, as it seems like
it should be simpler. Two things that got in the way were the type of ParseResult
requiring specific kinds of Error, and trying to figure out how nested the final
value should be.


alexjg:
Could you rebase this on `main` to fix the `cargo deny` CI failure?

rust/automerge/tests/test.rs
```diff
@@ -1412,6 +1412,31 @@ fn fuzz_crashers() {
...
+    // the value metadata says "2", but the LEB is only 1-byte long and there's an extra 0
+    let input = vec![
+        133, 111, 74, 131, 101, 189, 58, 107, 0, 110, 1, 16, 189, 164, 138, 64, 231, 24, 77, 199,
+        169, 158, 106, 229, 216, 158, 46, 212, 1, 202, 20, 126, 211, 151, 175, 73, 116, 227, 106,
+        198, 46, 28, 245, 111, 85, 96, 236, 50, 60, 255, 184, 1, 237, 26, 202, 247, 135, 165, 111,
+        117, 39, 6, 1, 2, 3, 2, 19, 2, 35, 2, 64, 2, 86, 2, 8, 21, 3, 33, 2, 35, 2, 52, 1, 66, 2,
+        86, 2, 87, 2, 128, 1, 2, 127, 0, 127, 1, 127, 1, 127, 0, 127, 0, 127, 7, 127, 1, 97, 127,
+        0, 127, 1, 1, 127, 1, 127, 40, /* */ 28, 0, /* */ 127, 0,
+    ];
+    assert!(Automerge::load(&input).is_err());
```
ConradIrwin, line 1426: I'd love to match the error message rather than just saying "is an error", what's the best way to do that; my editor wasn't suggesting any kind of ".string()" method on the returned error value.

rust/automerge/tests/test.rs
```diff
@@ -1412,6 +1412,31 @@ fn fuzz_crashers() {
...
+fn overlong_leb() {
+    // the value metadata says "2", but the LEB is only 1-byte long and there's an extra 0
+    let input = vec![
+        133, 111, 74, 131, 101, 189, 58, 107, 0, 110, 1, 16, 189, 164, 138, 64, 231, 24, 77, 199,
+        169, 158, 106, 229, 216, 158, 46, 212, 1, 202, 20, 126, 211, 151, 175, 73, 116, 227, 106,
+        198, 46, 28, 245, 111, 85, 96, 236, 50, 60, 255, 184, 1, 237, 26, 202, 247, 135, 165, 111,
+        117, 39, 6, 1, 2, 3, 2, 19, 2, 35, 2, 64, 2, 86, 2, 8, 21, 3, 33, 2, 35, 2, 52, 1, 66, 2,
+        86, 2, 87, 2, 128, 1, 2, 127, 0, 127, 1, 127, 1, 127, 0, 127, 0, 127, 7, 127, 1, 97, 127,
+        0, 127, 1, 1, 127, 1, 127, 40, /* */ 28, 0, /* */ 127, 0,
+    ];
```
ConradIrwin, line 1425: I'm also happy to put these fixtures into files instead of inline if that's better.

rust/automerge/tests/test.rs
alexjg: Yeah you're running into the problem of our `AuomergeError` enum being a giant grab bag of errors. On my infinitely long list of things to do is to split it into a set of more specific errors for different interactions.

There is another matter which is that probably the details of the specific thing that has gone wrong (the LEB encoding being overlong) don't really belong in the public API, it feels very much like an implementation detail. On the other hand, we do want to ensure that loading here is failing for the reason we expect it to. I suspect the best way to do that would be via a "metamorphic test", i.e. generate an input which _doesn't_ have the overlong LEB and check that it successfully loads, then modify the LEB encoding to be overlong and check that it fails to load.

ConradIrwin: Makes sense! My test initially passed because the checkup mismatched which is why i wanted to be more specific.

do you want a different approach to error handling in parse_input?

alexjg: Yeah I think the error handling is fine. I need to simplify the parse error stuff, it's pretty clunky due to a me not getting around to making a trait for `ParseResult` (which would allow combinators which are used in multiple contexts to be generic over any `ParseResult`, rather than having to do some weird stuff to feed error kinds out). It's on my list

rust/automerge/src/columnar/encoding/col_error.rs
```diff
@@ -6,6 +6,14 @@ pub(crate) struct DecodeColumnError {
 
 impl std::error::Error for DecodeColumnError {}
 
+impl From<crate::storage::parse::leb128::Error> for DecodeColErrorKind {
```
alexjg, line 9: I don't think we need this `From` implementation right? Removing this also allows us to remove the `pub(crate)` on `DecodeColErrorKind` as well.

ConradIrwin: Thanks! I was doing a lot of trial-and-error to get it working.

Pushed a new version now I've figured out why lifetimes were not working out for me before.

PR #523, opened 2023-02-08
Updated `unstable.splice` function doc type to `Extend`.

### LINKED ISSUE #522: Does new `Extend` type apply to `unstable.splice`?
I see that there is a new `Extend` type which has been introduced for the `ChangeFn`. It seems to me that as the `unstable.splice` function should always be called within a `change` function, it should have the new `Extend` type.

If someone can confirm, will submit a PR. Many thanks.

acurrieclark:
While this certainly works, I am not entirely sure this is the most sensible way forward. 

The `_isProxy`, `_state` and `_obj` functions all quite rightly expect a `Doc<T>` input. Is there an alternative?

Closes #522

### DIFF
--- a/javascript/src/unstable.ts
+++ b/javascript/src/unstable.ts
@@ -48,7 +48,7 @@ export {
   type ScalarValue,
 } from "./unstable_types"
 
-import type { PatchCallback } from "./stable"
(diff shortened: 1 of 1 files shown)


acurrieclark:
Closing this as it’s likely to be unnecessary shortly

PR #524, opened 2023-02-08
Port Marks to Main

orionz:
Porting the marks code to main.  Currently mark() and spans() work.  Most tests fail tho because they need to be able to serialize and deserialize the ops.


rust/automerge-wasm/src/lib.rs
```diff
@@ -786,6 +787,212 @@ impl Automerge {
...
+    }
+
+    pub fn unmark(&mut self, obj: JsValue, mark: JsValue) -> Result<(), JsValue> {
+        let (obj, _) = self.import(obj)?;
+        let mark = self.import_obj(mark)?;
+        self.doc.unmark(&obj, &mark).map_err(to_js_err)?;
+        Ok(())
+    }
+
+    pub fn spans(&mut self, obj: JsValue) -> Result<JsValue, JsValue> {
```
philschatz, line 830: :question: I am a little confused about the difference between a span and a mark.

As an example, would bolding a sentence and then unbolding a word in the middle of the sentence require something like the following?

```ts
let list = doc.putObject("_root", "list", "");

// Initially, bold the whole sentence
doc.splice(list, 0, 0, "aaabbbccc");
doc.mark(list, "[0..9]", "bold" , true);

// Then, to unbold the 'bbb' we need to find the correct span in raw_spans, ...
const raw_spans = doc.raw_spans(list);
const the_span = raw_spans[0];
// ... unmark the span/mark ...
doc.unmark(list, the_span.id);
// and then add 2 marks
doc.mark(list, "[0..3]", "bold" , true)
doc.mark(list, "[6..9]", "bold" , true)
```

Rather than alternating terminology, would it be helpful to name the methods to be more mark (or span)-specific? For example:

```ts
let list = doc.putObject("_root", "list", "");

// Initially, bold the whole sentence
doc.splice(list, 0, 0, "aaabbbccc");
const mark = doc.mark(list, "[0..9]", "bold" , true);

// Then, to unbold the 'bbb' we call unmark
doc.unmark(span.id);

// Or, if the client uses the markup() method to render it could be something like this:
const markup = doc.markup(list);
// [
//   [ [ '39@aaabbbccc', 'bold', 'boolean', true ] ],
//   'aaabbbccc',
//   [],
// ]
doc.unmark(markup[0][0][0]);

// and then add 2 marks
doc.mark(list, "[0..3]", "bold" , true);
doc.mark(list, "[6..9]", "bold" , true);
```

rust/automerge/src/query/spans.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    pub(crate) fn into_spans(self, m: &OpSetMetadata) -> Vec<Span<'a>> {
+        self.spans.into_iter().map(|s| s.into_span(m)).collect()
+    }
+
+    pub(crate) fn check_marks(&mut self) {
+        let mut new_marks = HashMap::new();
+        for op in &self.ops {
+            if let OpType::MarkBegin(m) = &op.action {
+                new_marks.insert(m.name, &m.value);
```
philschatz, line 72: :question: It seems like hashing on the mark name prevents multiple comments on text; is that intentional?

rust/automerge/src/query/spans.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    pub(crate) fn into_spans(self, m: &OpSetMetadata) -> Vec<Span<'a>> {
+        self.spans.into_iter().map(|s| s.into_span(m)).collect()
+    }
+
+    pub(crate) fn check_marks(&mut self) {
+        let mut new_marks = HashMap::new();
+        for op in &self.ops {
+            if let OpType::MarkBegin(m) = &op.action {
```
philschatz, line 71: :question: Could the ExId be included in `self.marks` as well? For users that use the data from `doc.spans()` to render in an editor, it seems that including the mark ExId would be useful, rather than having to look it up later in `.raw_spans()`.

philschatz (commented):
:wave: Hello! I was reading through the Peritext paper/code and was excited to help with getting marked ranges added to Automerge and heard work was already underway here.

I tried it out to get a feel for how it would work and noticed a few surprising behaviors. If they are intentional then please ignore these comments; if not I'm happy to submit a PR. It's been fun to read the diff!

rust/automerge/src/query/spans.rs
orionz: the names of the marks are unique per name - for something like bold use 

`doc.mark(list, "[0..9]", "bold" , true);`

or

`doc.mark(list, "[0..9]", "font-weight" , "bold");`

this way different markings with the same name will merge together or overwrite each other if the values are different.  For comments - each of which are distinct - i would recommend using unique values for name.  Something like

`doc.mark(list, "[0..9]", `comment:${hash}` , "This is a comment");`

I'm realizing the tests have just been to validate that the code is working and not to demonstrate how to use mark in the real world.  ;)  I'll make sure to fix that before these features go live!

rust/automerge-wasm/src/lib.rs
orionz: You are correct - I got marks and spans mixed up when naming those functions.  With the last push I removed `spans()` and added `marks()` and renamed `raw_spans()` to `rawMarks()`.  `rawMarks` is intended for debugging and will likely be removed before this goes live.  Also unsure if `unmark()` will be in the final api.

As for how to unbold a section - you should do it like this.

```
  doc.mark(list, "[0..9]", "bold" , true);
  doc.mark(list, "[3..6]", "bold" , false);
```

This way it'll play nicely with other concurrent marks being made by other users.

rust/automerge/src/query/spans.rs
orionz: I've made some major changes to the api.  Now the only api calls are 

```typescript
  mark(obj: ObjID, key: string, range: string, value: Value, datatype?: Datatype): void;
  unmark(obj: ObjID, key: string, start: number, end: numbr): void;
  marks(obj: ObjID): Mark[];
```

Unmark now takes a range.  Marks do not have Id's.  The terminology should be consistent.

rust/automerge-wasm/index.d.ts
```diff
@@ -103,6 +103,12 @@ export type PutPatch = {
   conflict: boolean
 }
 
+export type MarkPatch = {
+  action: 'mark'
+  path: Prop[],
+  marks: Mark[]
+}
+
```
philschatz, line 111: The matching `unmark` patch seems to be missing:

```suggestion

export type UnmarkPatch = {
  action: 'unmark'
  path: Prop[],
  start: number
  end: number
  key: string
}
```

philschatz (commented):
Thanks, these changes helped.

With them I was able to update the Peritext repo to use Automerge again. You can try it [here](https://philschatz.com/peritext/dist/) and see the diff [here](https://github.com/inkandswitch/peritext/compare/main...philschatz:peritext:updates#diff-67aec1adcf4cadd88a9983a3111f5b54f2c98f8d8242a36814d36578b0b76ace)

rust/automerge/src/automerge.rs
```diff
@@ -1309,6 +1345,64 @@ impl ReadDoc for Automerge {
...
     }
 
+    fn get_marks<O: AsRef<ExId>>(&self, obj: O) -> Result<Vec<Mark<'_>>, AutomergeError> {
+        let (obj, obj_type) = self.exid_to_obj(obj.as_ref())?;
+        let encoding = ListEncoding::new(obj_type, self.text_encoding);
+        let ops_by_key = self.ops().iter_ops(&obj).group_by(|o| o.elemid_or_key());
+        let mut pos = 0;
+        let mut marks = MarkStateMachine::default();
+
+        Ok(ops_by_key
```
alexjg, line 1355: I think this could be done without the side-effecting `pos` stuff by using a `fold::<(usize, Vec<Mark>), ..>((0, Vec::new(), ..)` rather than than a `filter_map` followed by `collect`

rust/automerge/src/marks.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    fn mark_below<'b>(
+        state: &'b mut [(OpId, Mark<'a>)],
+        index: usize,
+        mark: &Mark<'a>,
+    ) -> Option<&'b mut Mark<'a>> {
+        Some(
+            &mut state[0..index]
+                .iter_mut()
+                .filter(|(_, m)| m.data.key == mark.data.key)
```
alexjg, line 152: Any reason we're using `data.key` here and `key()` in `mark_above`?

rust/automerge/src/op_observer.rs
```diff
@@ -111,6 +119,15 @@ pub trait OpObserver {
...
     fn delete_seq<R: ReadDoc>(&mut self, doc: &R, objid: ExId, index: usize, num: usize);
 
+    fn mark<'a, R: ReadDoc, M: Iterator<Item = Mark<'a>>>(
+        &mut self,
+        doc: &'a R,
+        objid: ExId,
+        mark: M,
+    );
+
+    fn unmark<R: ReadDoc>(&mut self, doc: &R, objid: ExId, key: &str, start: usize, end: usize);
```
alexjg, line 129: I think we should rename `key` to `mark_name` here.

rust/automerge/src/op_observer/vec_observer.rs
```diff
@@ -0,0 +1,561 @@
...
+
+use crate::{marks::Mark, ObjId, OpObserver, Prop, ReadDoc, ScalarValue, Value};
+
+use crate::sequence_tree::SequenceTree;
+
+use crate::op_observer::BranchableObserver;
+use crate::op_observer::{Patch, PatchAction};
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum TextRepresentation {
```
alexjg, line 13: It's upsetting that we need to make this `pub` because it's really a backwards compatibility artifact specifically for `automerge-wasm`. I can't think of a neat way to avoid it though. On that basis I think we should do two things:

1. Add some documentation to the `Array` variant explaining that new implementations should not use this representation.
2. Given that we have this `TextRepresentation` enum in our public API we can at least replace `OpObserver::test_as_seq` with `OpObserver::text_representation() -> TextRepresentation` which will be clearer

rust/automerge/src/op_observer/vec_observer.rs
```diff
@@ -0,0 +1,561 @@
...
+struct VecOpObserverInner<T: TextIndex> {
+    pub(crate) patches: Vec<Patch<T::Item>>,
+    pub(crate) text_rep: TextRepresentation,
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct VecOpObserver(VecOpObserverInner<Utf8TextIndex>);
+
+#[derive(Debug, Clone, Default)]
+pub struct VecOpObserver16(VecOpObserverInner<Utf16TextIndex>);
```
alexjg, line 51: We should document that this is the same as `VecOpObserver` but text indices will use utf-16 offsets rather than `utf-8`.

rust/automerge/src/query/insert.rs
```diff
@@ -110,6 +110,12 @@ impl<'a> TreeQuery<'a> for InsertNth {
             self.last_seen = None;
             self.last_insert = element.elemid();
         }
+        /*-------------------*/
```
alexjg, line 113: What's the purpose of the comment bars?

rust/automerge/src/query/seek_mark.rs
```diff
@@ -0,0 +1,124 @@
...
+
+impl<'a> SeekMark<'a> {
+    pub(crate) fn new(id: OpId, end: usize, encoding: ListEncoding) -> Self {
+        SeekMark {
+            id,
+            encoding,
+            end,
+            found: false,
+            next_mark: None,
+            mark_key: "".into(),
```
alexjg, line 33: Should `mark_key` instead be `Option<SmolStr>` to capture this logic?

rust/automerge/src/read.rs
```diff
@@ -129,6 +129,16 @@ pub trait ReadDoc {
     /// Get the type of this object, if it is an object.
     fn object_type<O: AsRef<ExId>>(&self, obj: O) -> Result<ObjType, AutomergeError>;
 
+    /// Get all marks on a current sequence
+    fn get_marks<O: AsRef<ExId>>(&self, obj: O) -> Result<Vec<Mark<'_>>, AutomergeError>;
```
alexjg, line 133: We've used `x()` rather than `get_x()` for getters in this trait (which is common in Rust). Suggest we do that here as well.

rust/automerge/src/transaction/inner.rs
```diff
@@ -630,6 +631,7 @@ impl TransactionInner {
             // handle the observer
             if let Some(obs) = op_observer.as_mut() {
                 match splice_type {
+                    //SpliceType::Text(text, _) => { //if !obs.text_as_seq() => {
```
alexjg, line 634: nit: remove comment

rust/automerge/src/transaction/transactable.rs
```diff
@@ -88,6 +89,21 @@ pub trait Transactable: ReadDoc {
         text: &str,
     ) -> Result<(), AutomergeError>;
 
+    fn mark<O: AsRef<ExId>>(
+        &mut self,
+        obj: O,
+        mark: Mark<'_>,
+        expand: (bool, bool),
```
alexjg, line 96: I think that rather than having these two booleans, which are hard to read for the caller, we should use an enum like:

```rust
enum ExpandMark  {
    Left,
    Right,
    Both,
    None,
}

impl ExpandMark {
    fn left(&self) -> bool {
        matches!(self, Self::Left | Self:Both)
    }
    fn right(&self) -> bool {
        matches!(self, Self::Right | Self::Both)
    }
}
```

rust/automerge/tests/test.rs
```diff
@@ -1283,6 +1283,7 @@ fn test_change_encoding_expanded_change_round_trip() {
     let change = automerge::Change::try_from(&change_bytes[..]).unwrap();
     assert_eq!(change.raw_bytes(), change_bytes);
     let expanded = automerge::ExpandedChange::from(&change);
+    println!("{:?}", expanded);
```
alexjg, line 1286: nit: remove print

alexjg (commented):
I've left a whole bunch of comments :slightly_smiling_face: 

When we initially talked about this design we talked about handling marks which expand to before or after the document text (the equivalent of Peritext's `startOfText` and `endOfText` attributes) by omitting either the begin or end mark. I.e. if you find a begin mark with no end mark then it is the equivalent of `type: "endOfText"` and if you find an end mark with no begin mark then it is the equivalent of `type: "startOfText"`. As far as I can see that is not implemented here, do we end up not needing it or is that something we would implement in future?

One other naming convention. I suggest that we refer to what is currently called the mark "key" as the "mark name". This is because "key" is already an overloaded term, appearing in maps, so I think it would be clearer to use a different term.

rust/automerge-wasm/index.d.ts
```diff
@@ -236,3 +269,18 @@ export class SyncState {
   sentHashes: Heads;
   readonly sharedHeads: Heads;
 }
+
+export type ChangeSetDeletion = {
```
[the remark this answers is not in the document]
orionz: Nothing anymore!

javascript/src/unstable.ts
```diff
@@ -233,6 +237,66 @@ export function splice<T>(
   }
 }
 
+export function mark<T>(
+  doc: Doc<T>,
+  prop: stable.Prop,
+  key: string,
+  range: string,
```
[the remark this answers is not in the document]
orionz: @pvh - wanted your input - this is the one piece of feedback I didn't know  what to do with - the current API where expand left / right and start and end is all packed into a string `"(3..30)"` or `"[10..20)"` is obviously not great - what do you think of alex's proposal here to make it `Automerge.mark(doc.foo.bar, "baz", "mark_name", "mark_value", { sticky: "both", start: 10, end: 20 })`

rust/automerge-wasm/index.d.ts
orionz: Wonderful news - I've added the unmark patch.  Good catch.

PR #529, opened 2023-02-10
A few setup fixes

philschatz:
:wave: Thanks for this great project! While getting getting set up locally I ran into a few items.

Also, the following line seems to be duplicated in `automerge.rs`:

```rust
self.history_index.insert(change.hash(), history_index);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,9 @@ brew install cmake node cmocka
 # install yarn
 npm install --global yarn
 
+# install javascript dependencies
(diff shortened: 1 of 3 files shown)


alexjg:
This is all nice stuff, thanks! Any idea why the lint is failing?

philschatz:
Hmm, I'm not sure about the clippy error and was unable to reproduce it so I removed it. Can you try running the Actions again?

PR #540, opened 2023-03-02
Add default export to wasm package.json

echarles:
While trying to update https://github.com/dmonad/crdt-benchmarks to latest @automerge/automerge-wasm, I have hit `ERR_PACKAGE_PATH_NOT_EXPORTED` error while importing with `import Automerge from '@automerge/automerge-wasm'``.

This PR adds an `default` entry to the exports listed in package.json.

### DIFF
--- a/rust/automerge-wasm/package.json
+++ b/rust/automerge-wasm/package.json
@@ -55,6 +55,7 @@
   },
   "exports": {
     "browser": "./bundler/automerge_wasm.js",
+    "default": "./nodejs/automerge_wasm.js",
     "require": "./nodejs/automerge_wasm.js"
   }
 }


alexjg:
Can you point me to some docs on the `default` key? I've not seen it before. I also didn't run into this problem when I attempted such a port myself (which I have not completed), what environment are you running in?

echarles:
These are the changes needed to update the crdt-benchmarks repo to latest automerge release https://github.com/datalayer-externals/crdt-benchmarks/commit/9701a905a7d327c3e0795e5238e2a01f25387762

I don't have any doc to share and added the default key by trial-and-error. If you have time, you can check out that branch and see how it works for you.

orionz:
Sorry - we had all kinds of issues with default exports.  Apparently they spec on them is unclear and after getting a bunch of bug reports we decided the best thing to do was remove it.  I think the correct way to fix crdt benchmarks would be a PR for them to not use the default exports anymore.

PR #542, opened 2023-03-03
Suppress clippy warning in parse.rs + bump toolchain

chrstnb:
This clippy warning was causing issues when running `./scripts/ci/run`; one option would be to rename the getter, or to switch up the naming of the fields, but I figured suppressing the clippy warning would be reasonable.

```
error: getter function appears to return the wrong field
   --> automerge/src/storage/parse.rs:311:5
    |
311 | /     pub(crate) fn bytes(&self) -> &'a [u8] {
312 | |         self.original
    | |         ------------- help: consider using: `self.bytes`
313 | |     }
    | |_____^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#misnamed_getters
    = note: `-D clippy::misnamed-getters` implied by `-D warnings`

error: could not compile `automerge` due to previous error
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: 1.66.0
+          toolchain: 1.67.0
           default: true
           components: rustfmt
       - uses: Swatinem/rust-cache@v1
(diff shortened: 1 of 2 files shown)


alexjg:
I very much appreciate someone else doing the work of making the monthly tribute to our oppressive robotic overlord! Unfortunately CI is failing because the CI scripts are still on the `1.66` toolchain, which doesn't know about this lint. If you could update the toolchain in `.github/workflows/ci.yaml` to `1.67` it should all work.

PR #544, opened 2023-03-07
Public clock API

jeffa5:
For one of my projects I've found a need to work with vector clocks from automerge heads primarily because the heads can't be compared to indicate which is newer. This PR adds a conversion from heads to a clock and vice-versa, as well as some methods on the public vector clock that enable users to interact with it and create their own for testing.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "flake-utils": {
       "locked": {
-        "lastModified": 1667395993,
-        "narHash": "sha256-nuEHfE/LcWyuSWnS8t12N1wc105Qtau+/OdUAjtQ0rA=",
+        "lastModified": 1676283394,
+        "narHash": "sha256-XX2f9c3iySLCw54rJ/CZs+ZK6IQy7GXNY4nSOyu2QG4=",
         "owner": "numtide",
         "repo": "flake-utils",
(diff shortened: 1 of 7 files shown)


alexjg:
Is there a reason that the details of the clocks need to be exposed or would it suffice to expose an API for determining the ancestry relationships of two change hashes? The reason I ask is that I very much want to get rid of Actor IDs entirely and so it would be nice if we could do this without introducing an API which ties us to a specific model of causality.

jeffa5:
My main thought is that with a vector clock clients can jump to a point in history that they haven't necessarily observed before, I don't think the ancestry relationship would be enough for this because they'd need to know the historical hashes first. I'll make this a draft and have a think around this.

I'd be interested to hear more about your thoughts on removing the actor ids, is there an issue for it with more detail?

alexjg:
I created a discussion about removing actor IDs #546

orionz:
I'm ok with having an external clock as long as you have an actual use for it.

1. I personally would like it if it was a BTreeMap internally so the iterator has predictable behavior (and order of actorId's matters in a lot of cases). 
2. I like how ExId has a versioned binary format to make it easy to serialize.  Seems like a worthwhile addition to this as well.

jeffa5:
I think I've managed to avoid need the exposed clocks for now so closing this.

PR #545, opened 2023-03-08
smaller automerge c

ConradIrwin:
Generate significantly smaller automerge-c builds

This cuts the size of libautomerge_core.a from 25Mb to 1.6Mb on macOS
and 53Mb to 2.7Mb on Linux.

As a side-effect of setting codegen-units = 1 for all release builds the
optimized wasm files are also 100kb smaller.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -10,13 +10,8 @@ members = [
 resolver = "2"
 
 [profile.release]
-debug = true
 lto = true
-opt-level = 3
+codegen-units = 1
 
 [profile.bench]
 debug = true
-
-[profile.release.package.automerge-wasm]
-debug = false
-opt-level = 3
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,10 +2,10 @@ name: CI
 on:
   push:
     branches:
-    - main
+      - main
   pull_request:
     branches:
-    - main
+      - main
 jobs:
   fmt:
     runs-on: ubuntu-latest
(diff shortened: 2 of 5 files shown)


alexjg:
LGTM, I'm happy to merge if @jkankiewicz is happy with the cmake stuff

jkankiewicz:
> LGTM, I'm happy to merge if @jkankiewicz is happy with the cmake stuff

@ConradIrwin, invoking `bash` is unnecessary because the `execute_process()` function already pipes the output of one `COMMAND` into the input of the next e.g.
```
execute_process (
    COMMAND "rustc -vV"
    COMMAND "sed -n 's/host: //p'"
    OUTPUT_VARIABLE CARGO_TARGET
    OUTPUT_STRIP_TRAILING_WHITESPACE
)
```
Unfortunately, this still isn't sufficiently cross-platform because, like `bash`, the `sed` command isn't available on Windows either but a *.cmake script can be written to parse the output from `rustc` instead (CMake is Turing-complete).

ConradIrwin:
Thanks @jkankiewicz, updated!

rust/Cargo.toml
```diff
@@ -10,13 +10,8 @@ members = [
...
-opt-level = 3
+codegen-units = 1
 
 [profile.bench]
-debug = true
-
-[profile.release.package.automerge-wasm]
-debug = false
-opt-level = 3
+debug = true
```
jkankiewicz, line 17: GitHub prefers for every file to be terminated with a newline.

jkankiewicz (approved):

PR #550, opened 2023-03-10
Add hash_for_opid function

jeffa5:
This function enables users to get the hash that corresponds to the change that created the operation id given.
This hash can then be used to view historical versions of the document at that point, useful if multiple things change in a single transaction and you want to look back in time at those.


jeffa5:
This doesn't enable use inside transactions as I think that could be misleading (can't use ids created within the transaction) but I'd be happy to change it if it seems useful (can also change down the line).

rust/automerge/src/automerge.rs
```diff
@@ -1432,6 +1432,34 @@ impl ReadDoc for Automerge {
             .get(hash)
             .and_then(|index| self.history.get(*index))
     }
+
+    fn hash_for_opid(&self, opid: &ExId) -> Option<ChangeHash> {
+        match opid {
```
orionz, line 1437: You need to convert the ExId to an ObjId via `let obj = self.exid_to_obj(obj)?;`  The reason being is that if the ExId was created by parsing a string the actor_id will be correct but the actor_index will not be.  The `exid_to_obj()` method sorts this out.

rust/automerge/src/read.rs
```diff
@@ -196,4 +196,7 @@ pub trait ReadDoc {
 
     /// Get a change by its hash.
     fn get_change_by_hash(&self, hash: &ChangeHash) -> Option<&Change>;
+
+    /// Get the hash of the change that contains the given opid.
+    fn hash_for_opid(&self, opid: &ExId) -> Option<ChangeHash>;
```
orionz, line 201: I'd like to NOT add new functions to ReadDoc if it can be avoided.  In fact I'd like to remove `get_change_by_hash` and `get_missing_deps` from the trait.   Having it be a public function on `Automerge` and `Autocommit` should be sufficient.

orionz (changes requested):
Looks good - If you can address those two issues I'll merge it in.

rust/automerge/src/automerge.rs
jeffa5: I've added a `exid_to_opid`, which is what that `exid_to_obj` did internally that gets us what we need for this function. Otherwise we get errors for querying the hashes of non-object ops

rust/automerge/src/transaction/manual_transaction.rs
```diff
@@ -39,6 +39,10 @@ impl<'a, Obs: observation::Observation> Transaction<'a, Obs> {
             observation: Some(obs),
         }
     }
+
+    pub fn hash_for_opid(&self, opid: &ExId) -> Option<ChangeHash> {
```
alexjg, line 43: As you say in the comments, this feels like it might be confusing. Could we add some docs to the effect that if you lookup an opid you created in this transaction you'll get `None`?

I suppose we could even detect that it's an opid created in this transaction and have a return type that reflects that. Something like:

```rust
enum HashInTx {
    ThisTx,
    OlderChange(ChangeHash)
}
```

jeffa5: I've added the docs. Given the differences when calling it on a Automerge / Transaction vs an AutoCommit I think keeping the types the same is easiest for now.

orionz (approved):

PR #552, opened 2023-03-10
First Draft of a Diff API Api

orionz:
Uses an arbitrary observer to create a set of patches to get from any previous point in the graph to a later point.

Basic format is

```js
   let patches = doc.diff(heads1, heads2);
```

The algorithm probably needs a deep look from @ept as it came as a result of trial and error but currently I have not been able to write a test that breaks the current implementation.

Known bugs:  

1. ~~Needs a fix when a put is applied to a text string changing the char length of that position - currently not possible because text is only accessible via slice but could exist in the data and I want to make sure its covered.~~

2. The doc.path() api call used by the observer is generating a path for the end state and not point in time of the diff - need a way to fix this.  I have an idea for a general purpose `fn at(&self, Heads) -> AutomergeAt<'_>` api call that will allow all of the `get_at` and `keys_at` api's to be removed.

3. Currently marks are ignored


jeffa5:
I think it would be interesting to have some benchmarks around this to see where the time is being spent. My concern is when the diff will be small but there are lots of other bits to compare (lots of objects for instance).

The main way I could see this being different would be to get the relevant changes and work off the ops in those, but maybe this isn't viable?

ept:
I would like to give this algorithm some careful thought, but I don't have the bandwidth to review the code right now. Maybe we could talk this through in a month when I'm back from parental leave?

orionz:
Also I'd imagine there's a good heuristic where if we see apply changes or receive sync message contains a number of ops larger than some percent of the total doc ops - we should just call the unobserved function and then call diff after the fact - similar to how when we do a load on an empty doc now we call `observer_document` at the end which is kind of like `diff(&[].&doc.get_heads())`

jeffa5:
Hmm, having a quick play around, it seems like `At<'a>` should be able to work if we add lifetime bounds, but I'm not sure where to add them exactly yet.

orionz:
This work was finished and will be a part of PR #572

PR #560, opened 2023-03-15
Fix SeekOpWithPatch

rroohhh:
As documented by the FIXMEs, it was handling last_seen wrong. This uses the same logic as Nth.


orionz:
Thanks for this.  This code is being rewritten in PR #572

PR #562, opened 2023-03-17
`Automerge::repair` - fixing mismatched heads where possible

alexjg:
This is an incomplete implementation of an `Automerge::repair` tool which might be able to repair documents with mismatched heads. 

The basic idea is this. In practice the mismatched heads errors I have seen are due to an issue where the predecessors of an operation are not correctly added to the `OpSet` (which is possible without causing weird behaviour at runtime because `pred` is not used in the `OpSet` at all, just `succ`) but _are_ added to the encoded changes in `TransactionInner::commit`. This means that the change hashes we encode are different to the change hashes we calculate on decompression. To fix this then, we should be able to scan through the opset looking for ops which _should_ have a `pred` - because there have been subsequent `set` operations which we know were not concurrent due to the change graph - and add the `preds` to those ops, then rebuild the change graph.

Unfortunately I haven't managed to fix anything with this approach. I think that's partly because my logic for detecting mismatching predecessors is wrong - but it's also probably that I am basing this logic on errors I know about, and documents corrupted due to other bugs would need other logic to fix.

I'm pushing this because parts of the logic may be useful to someone who has better ideas about how to check for corruption errors (in particular the change log rebuilding might be useful and the additional  traversals on the `ChangeGraph`). However, I think an alternative approach is probably needed where instead of trying to repair the document, we focus on trying to extract the usable history from it - maybe some kind of rebase operation or a "extract current state" operation. We could also do with more checking when _encoding_ that the heads match. Mismatching heads errors would be much less pernicious if we threw them _before_ persisting or sending data.


rust/automerge/src/doctor.rs
```diff
@@ -0,0 +1,123 @@
...
+    // change containing the operation but which does not have a `pred` is suspicious. There's no
+    // way that such an operation would be generated by any well behaved library.
+    //
+    // scan through every op, building a list of operations which _should_ precede each op
+    let clocks = doc.build_clocks();
+
+    let mut keyops = Vec::new();
+    let mut mismatches = Vec::new();
+    for (obj, _typ, ops) in doc.ops().iter_objs() {
+        let ops_by_key = ops.group_by(|o| o.key);
```
orionz, line 33: This will work for maps but not for lists - should be `group_by(|o| o.elemid_or_key())` since the first op in a list that has the insert flag has a different key than the rest and could get grouped with the previous elemid

PR #563, opened 2023-03-19
fix: `Nth` query sometimes miscounting

pajowu:
This PR fixes a bug in the `Nth` query which led to it not counting the visible element in the parent after skipping a child node.

### DIFF
--- a/rust/automerge/src/lib.rs
+++ b/rust/automerge/src/lib.rs
@@ -261,7 +261,7 @@ mod map_range_at;
 pub mod marks;
 pub mod op_observer;
 mod op_set;
-mod op_tree;
+pub mod op_tree;
 mod parents;
 mod query;
 mod read;
--- a/rust/automerge/src/lib.rs
+++ b/rust/automerge/src/lib.rs
@@ -295,6 +295,7 @@ pub use op_observer::{
 };
 pub use parents::{Parent, Parents};
 pub use read::ReadDoc;
+pub use sequence_tree::SequenceTree;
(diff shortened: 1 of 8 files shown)


alexjg:
Thanks for this, I know how difficult these bugs can be to track down. This looks good to me but I would like to wait for a review from @orionz as he spends a lot of time in this part of the codebase.

pajowu:
Great, thank you.

pajowu:
I added the same logic in the nth-query (which includes the fix from #560) and the insert query and added regression tests for both. I also rebased the to the current main and fixed the test for the changes in the `Patch` api

orionz:
Hey there - thanks for the PR.  Good find.

I think you've introduced a new bug into `seek_op_with_patch.rs` - while we're tracking last_seen but we're not using it. You removed the part where we do a `num_vis -= 1`

If one node ends with [ .., key1 ] and the next begins with [ key1, ...] - a conflicting value we don't want to double count key1 ergo the `-1` since that key has already been counted ... although reading this code I'm realizing we need to be able to handle the case where the widths of the keys are different - realizing the -1 is wrong tho because the width of the last element may be something other than 1.  Perhaps we need to change `last_seen` from `Option<Key>` to `Option<(Key,usize)>` and record the width so we know how much to back up by.

Alternatively I'm wondering if it would be better to store `last_visible_element` in the index for each node so we can make each of these queries simpler

pajowu:
I updated the commit with the missing clause, I must have made a mistake copy-pasted the code.

orionz:
Thank you both for this PR!  Sorry I was slow to get it merged

PR #564, opened 2023-03-20
Document release processes

alexjg:


teohhanhui:
This seems to be missing any mention of the `automerge-test` crate, which would be a problem when the `automerge` crate gets bumped to the next semver backwards-incompatible version, e.g. `0.3.0` -> `0.4.0`.

PR #572, opened 2023-04-13
A rewrite of the internal op_set query and iterator code

orionz:
This rewrite has been a long time coming.  The goal was to move all of the fussy `op_set` query code into a smaller number of functions to make the code easier to audit and test and reduce the overall surface area of of bugs.  This patch (as of now) has about 2 lines deleted for each line added so I'm happy to say its overall much less code.

The primary goal here was to have as few simple functions as possible to accomplish these goals and only optimize if absolutely necessary.   Some things - such as NOT writing a normal and `at` version of every query had me concerned there would be a performance regression but overall everything is running 10-25% faster.  Some things, such as `doc.text()` seems to be 3x faster at least.

The op_set walking code now lives  in

- Keeping track of the logical list `index` vs op set `pos`: `src/query/list_state.rs`
- Searching the `op_set` for a specific `opid`: `src/query/opid.rs`
- Finding the ops overwritten in an op insert the `seek_op` function in `src/op_tree.rs`
- Walking the top visible ops on an object is in `op_set.rs` function `top_ops()`

The dozen complex queries reproducing a lot of the `op_tree` walking logic have been removed and now use these 4 tools to do their job.

There was a 20% performance regression on `apply_changes()` so I did end up having a 20% performance regression so now there is a expensive `seek_op_with_obsrever()` and `seek_op_simple()` method with two similar code paths.

Issues:

1. ~~The signatures of many public methods have changed.  They now are all `Box<dyn Iterator<Item = T>>`.   This is bad b/c it prevents the compiler from inlining iterator code.  I suspect someone with better Rust skills than me can help me make the correct type declarations.~~
2.  Some of the old iterators supported `DoubleEndedIterator`.  Since everything is now based on the `OpTreeIter` which does not support this trait that feature has been dropped.  I don't think it's especially needed but there are several tests that rely on it that had to be disabled.
3. Given the size of this rewrite it may be prudent to get a rust fuzz tester added before accepting this PR, or it could be pushed to a later PR.
4. ~~The `seek_mark` query still needs to be ported over to using `list_state`.~~
5. The C api needs to be updated to some small API changes

Further work:

1.~~I built this on top of the `diff` branch so it includes an incomplete `diff()` implementation.  I don't think that should halt the merge but it does mean diff should be finished as soon as possible so as not to have half working methods in the system.~~
2.  I was hoping to merge `diff` `current_state` and `seek_op_with_observer` into a single code path.  That work has not been done and I think is best saved for a later PR.
6. I noticed the `can_shortcut_search` only works on `splice()` - it could use a rewrite to work more generally now that the query code is so much simpler.


rust/automerge/src/op_tree.rs
```diff
@@ -71,78 +190,200 @@ impl OpTreeInternal {
...
+
+            if e.visible() {
+                return Some((op, index, false));
+            }
         }
+        Some((op, index, op.visible()))
     }
 
-    pub(crate) fn map_range_at<'a, R: RangeBounds<String>>(
+    pub(crate) fn seek_op_with_observer<'a>(
```
alexjg, line 306: I would find this easier to read if it were `seek_to_op_with_observer` or `find_op_with_observer`

rust/automerge/src/op_tree.rs
```diff
@@ -71,78 +190,200 @@ impl OpTreeInternal {
...
         }
     }
 
-    pub(crate) fn list_range<R: RangeBounds<usize>>(
-        &self,
-        range: R,
-    ) -> Option<query::ListRange<'_, R>> {
-        if self.root_node.is_some() {
-            Some(query::ListRange::new(range, self))
+    pub(crate) fn seek_op_simple<'a>(
```
alexjg, line 323: As per other comments I think this would be better named `find_op_without_observer`

rust/automerge/src/op_tree.rs
```diff
@@ -71,78 +190,200 @@ impl OpTreeInternal {
...
-    pub(crate) fn keys(&self) -> Option<query::Keys<'_>> {
-        if self.root_node.is_some() {
-            Some(query::Keys::new(self))
-        } else {
-            None
+    pub(crate) fn top_ops(&self, clock: Option<Clock>) -> TopOps<'_> {
+        TopOps::new(OpTreeIter::new(self), clock)
+    }
+
+    pub(crate) fn seek_update_for_op_simple<'a>(
```
alexjg, line 197: How about `found_op`?

rust/automerge/src/op_tree.rs
```diff
@@ -71,78 +190,200 @@ impl OpTreeInternal {
...
+
+        SeekOpFound::simple(pos, succ)
     }
 
-    pub(crate) fn keys_at(&self, clock: Clock) -> Option<query::KeysAt<'_>> {
-        if self.root_node.is_some() {
-            Some(query::KeysAt::new(self, clock))
-        } else {
-            None
+    pub(crate) fn seek_update_for_op_for_observer<'a>(
```
alexjg, line 226: How about `found_op_for_observer`?

rust/automerge/src/op_tree.rs
```diff
@@ -278,46 +519,43 @@ impl PartialEq for OpTreeInternal {
...
+        mut iter: T,
+        key: Key,
+        clock: Option<&Clock>,
+    ) -> Self {
+        let mut found = Self {
+            end_pos: start_pos,
+            ops: vec![],
+            ops_pos: vec![],
+        };
+        let mut op = iter.nth(start_pos);
```
alexjg, line 541: Rather then using `nth` and then `loop` can we use `Iterator::skip(start_pos - 1)`?

rust/automerge/src/op_tree.rs
```diff
@@ -71,78 +190,200 @@ impl OpTreeInternal {
...
+        index: usize,
+    ) -> SeekOpFound<'a> {
+        let mut iter = self.iter();
+        let mut found = None;
+        let mut before = None;
+        let mut num_before = 0;
+        let mut overwritten = None;
+        let mut after = None;
+        let mut succ = vec![];
+        let mut next = iter.nth(pos);
```
alexjg, line 240: Can we use `Iterator::skip(pos - 1)` here?

rust/automerge/src/op_tree.rs
```diff
@@ -71,78 +190,200 @@ impl OpTreeInternal {
...
+
+    pub(crate) fn seek_update_for_op_simple<'a>(
+        &'a self,
+        meta: &OpSetMetadata,
+        op: &'a Op,
+        mut pos: usize,
+    ) -> SeekOpFound<'a> {
+        let mut iter = self.iter();
+        let mut succ = vec![];
+        let mut next = iter.nth(pos);
```
alexjg, line 205: Can we use `Iterator::skip(pos - 1)` here followed by a normal `for` loop over `iter`?

rust/automerge/src/query/list_state.rs
```diff
@@ -0,0 +1,142 @@
+use crate::op_tree::{LastInsert, OpTreeNode};
+use crate::query::QueryResult;
+use crate::types::{Key, ListEncoding, Op};
+
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) struct ListState {
+    pub(crate) encoding: ListEncoding,
```
alexjg, line 7: Can we make most of these fields private and put the ones which are needed to be read crate wide (`last_seen`, `target`, `index`, and `pos`) behind accessors?

rust/automerge/src/query/list_state.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub(crate) last_seen: Option<Key>,
+    pub(crate) last_width: usize,
+    pub(crate) target: usize,
+    pub(crate) index: usize,
+    pub(crate) pos: usize,
+    pub(crate) clean: bool,
```
alexjg, line 13: As per comments on the `query::Index::clean` field, I think this might be easier to read if it was named `has_seen_insert` or similar.

rust/automerge/src/query/nth.rs
```diff
@@ -45,12 +35,16 @@ impl<'a> Nth<'a> {
         if let Some(e) = self.ops.first().and_then(|op| op.elemid()) {
             Ok(Key::Seq(e))
         } else {
-            Err(AutomergeError::InvalidIndex(self.target))
+            Err(AutomergeError::InvalidIndex(self.idx.target - 1))
```
alexjg, line 38: Maybe we should use `saturating_sub` here? It's technically not necessary because we initialize `ListState` with `target + 1` so I think `target` should always be `> 0`, but we're in an error branch here already so it's probably worth programming defensively to avoid the panic.

rust/automerge/src/read.rs
```diff
@@ -27,6 +30,11 @@ pub trait ReadDoc {
     /// This function may in future be changed to allow getting the parents from the id of a scalar
     /// value.
     fn parents<O: AsRef<ExId>>(&self, obj: O) -> Result<Parents<'_>, AutomergeError>;
+    fn parents_at<O: AsRef<ExId>>(
```
alexjg, line 33: Needs a short docstring explaining that this is the same as `parents` but at a given heads.

rust/automerge/src/read.rs
```diff
@@ -38,28 +46,34 @@ pub trait ReadDoc {
     /// * If the object ID `obj` is not in the document
     fn path_to_object<O: AsRef<ExId>>(&self, obj: O) -> Result<Vec<(ExId, Prop)>, AutomergeError>;
 
+    fn path_to_object_at<O: AsRef<ExId>>(
```
alexjg, line 49: Needs docs. In particular we should clarify what the behavior is here if the object is in the document but was not visible at `heads`

rust/automerge/src/types.rs
```diff
@@ -158,6 +159,12 @@ pub enum ObjType {
     Text,
 }
 
+impl Default for ObjType {
```
alexjg, line 162: This seems surprising to me. Do we need a `Default` implementation for `ObjType`? I'm very suspicious of `Default` implementations which are not isomorphic to some kind of "zero" value, which `ObjType` doesn't have (to my mind anyway).

alexjg (commented):
Whole bunch of comments, mostly about naming. Overall I really like this, it dramatically simplifies things.

rust/automerge/src/automerge/diff.rs
```diff
@@ -0,0 +1,1399 @@
...
+
+impl<'a> Deref for Winner<'a> {
+    type Target = Op;
+
+    fn deref(&self) -> &'a Self::Target {
+        self.op
+    }
+}
+
+struct OpState {}
```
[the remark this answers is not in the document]
orionz: Lol

rust/automerge/src/automerge/diff.rs
```diff
@@ -0,0 +1,1399 @@
...
+                    conflict,
+                },
+                PatchAction::SpliceText { index, value } => ObservedPatch {
+                    action: "splice".into(),
+                    path: ex_path_and(path, index),
+                    value: value.into_iter().collect::<String>().into(),
+                    conflict: false,
+                },
+                PatchAction::Mark { .. } => {
+                    todo!()
```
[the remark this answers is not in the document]
orionz: I actually wrote a bunch in `automerge-wasm/test/diff.ts` but we could always use more

rust/automerge/src/autocommit.rs
```diff
@@ -317,6 +315,22 @@ impl<Obs: Observation> AutoCommitWithObs<Obs> {
...
     }
 
+    pub fn diff(
+        &mut self,
+        start: &[ChangeHash],
+        end: &[ChangeHash],
+    ) -> Result<Obs::Obs, AutomergeError> {
+        self.ensure_transaction_closed();
+
+        if let Some(observer) = self.observation.observer() {
```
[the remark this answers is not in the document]
orionz: clever!

rust/automerge/src/automerge/diff.rs
```diff
@@ -0,0 +1,1399 @@
...
+                path: "/key".into(),
+                action: "put_map".into(),
+                value: ScalarValue::Counter(12.into()).into(),
+                conflict: false,
+            })
+            .as_ref()
+        );
+    }
+
+    // test text update that changes length
```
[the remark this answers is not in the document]
orionz: And the tests revealed a bug! 🤪

rust/automerge/src/op_observer.rs
```diff
@@ -148,6 +148,14 @@ pub trait BranchableObserver {
     /// thrown away on `rollback()`
     fn branch(&self) -> Self;
 
+    /// Branch used by diff to communicate that patches are explicitly being asked for
+    fn explicit_branch(&self) -> Self
```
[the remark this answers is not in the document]
orionz: Tells toggle observer to turn itself on if its off.  Without this calling `doc.diff( a,b)` without first calling `doc.enablePatches()` would return nothing.  Open to alternatives.

rust/automerge/src/op_tree.rs
orionz: Because then we have to call `Iterator::next()` one more time than we need to in the absolute hottest function in the entire codebase? 😄

orionz: Also start_pos could be zero

rust/automerge/src/query/seek_mark.rs
```diff
@@ -1,56 +1,45 @@
...
-    seen: usize,
-    last_seen: Option<Key>,
     super_marks: HashMap<OpId, smol_str::SmolStr>,
     pub(crate) marks: Vec<Mark<'a>>,
 }
 
 impl<'a> SeekMark<'a> {
     pub(crate) fn new(id: OpId, end: usize, encoding: ListEncoding) -> Self {
         SeekMark {
+            idx: ListState::new(encoding, usize::MAX),
```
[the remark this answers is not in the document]
orionz: Yeah . These are the few functions where adding an `if` statement literally costs us 1% on a benchmark.  I'll take a usize::MAX over an Option<usize> just to avoid having to check if before <.  This gives the same result with fewer ops.

rust/automerge/src/transaction/inner.rs
```diff
@@ -657,9 +674,10 @@ impl TransactionInner {
         mark: Mark<'_>,
         expand: ExpandMark,
     ) -> Result<(), AutomergeError> {
-        let (obj, _obj_type) = doc.exid_to_obj(ex_obj)?;
+        let (obj, _obj_type, _encoding) = doc.exid_to_obj(ex_obj)?;
         if let Some(obs) = op_observer {
             let action = OpType::MarkBegin(expand.before(), mark.data.clone().into_owned());
+            // FIXME - encoding not considered
```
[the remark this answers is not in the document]
orionz: I think its old

rust/automerge/src/types.rs
orionz: Nope.  Whatever I needed that for was removed.

alexjg:
Almost everything looks good to me now, I put in a PR to this PR with a few small changes, take a look at that. There are only two things I'm still unsure about

1. Do we really need `BranchableObserver::explicit_branch`? It feels like requiring the user to call `enablePatches` before calling `diff` is reasonable. But maybe that API is hard to use from `automerge` JS?
2. Do we need the `OpObserver::unmark` call? It looks like that is never called and instead we represent "unmarking" something with a new `mark` that has `ScalarValue::null` as its value?

PR #576, opened 2023-04-20
Renumber marks columns in change chunks

### LINKED ISSUE #575: internal: the column id for marks is different between doc and change chunks
Currently there are two definitions of MARK_NAME_COL_ID and EXPAND_COL_ID:

```
src/storage/change/change_op_columns.rs-const PRED_COL_ID: ColumnId = ColumnId::new(7);
src/storage/change/change_op_columns.rs-const EXPAND_COL_ID: ColumnId = ColumnId::new(8);
src/storage/change/change_op_columns.rs:const MARK_NAME_COL_ID: ColumnId = ColumnId::new(9);
--
src/storage/document/doc_op_columns.rs-const SUCC_COL_ID: ColumnId = ColumnId::new(8);
src/storage/document/doc_op_columns.rs-const EXPAND_COL_ID: ColumnId = ColumnId::new(9);
src/storage/document/doc_op_columns.rs:const MARK_NAME_COL_ID: ColumnId = ColumnId::new(10);
```

This doesn't cause a problem in practice (as the two parsers are unrelated) but it is a bit untidy. We should probably change to use 9 and 10 in both change chunks and doc chunks before these appear in the wild.

/cc @alexjg @orionz

ConradIrwin:
They now match the ids in doc chunks

Fixes #575

### DIFF
--- a/rust/automerge/src/storage/change/change_op_columns.rs
+++ b/rust/automerge/src/storage/change/change_op_columns.rs
@@ -32,8 +32,8 @@ const INSERT_COL_ID: ColumnId = ColumnId::new(3);
 const ACTION_COL_ID: ColumnId = ColumnId::new(4);
 const VAL_COL_ID: ColumnId = ColumnId::new(5);
 const PRED_COL_ID: ColumnId = ColumnId::new(7);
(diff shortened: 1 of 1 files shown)


alexjg:
LGTM @orionz?

PR #578, opened 2023-04-20
Allow custom expand on unmark

### LINKED ISSUE #577: marks: unmark does not clear mark from beginning of doc
Consider the following test:

```
fn marks() {
    let mut tx = doc.transaction();

    let text_id = tx.put_object(&ROOT, "text", ObjType::Text).unwrap();

    tx.splice_text(&text_id, 0, 0, "hello world").unwrap();

    let mark = Mark::new("bold".to_string(), true, 0, "hello".len());
    tx.mark(&text_id, mark, ExpandMark::Both).unwrap();

    // add " cool"
    tx.splice_text(&text_id, "hello".len(), 0, " cool").unwrap();

    // unbold "hello"
    tx.unmark(&text_id, "bold", 0, "hello".len()).unwrap();

    // insert "why " before hello.
    tx.splice_text(&text_id,  0, 0, "why ").unwrap();
}
```

The document goes through the following versions:
* hello world
* <b>hello</b> world
* <b>hello cool</b> world
* <b></b>hello<b> cool</b> world
* <b>why </b>hello <b> cool</b> world

I would expect the final version to not have "why " be bolded.

This only happens with ExpandMark::Before or ExpandMark::Both; and can be "fixed" by using ExpandMark::Both instead of ExpandMark::None in unmark: https://github.com/automerge/automerge/blob/02c51aaef92357112255951206b1464de3051b7c/rust/automerge/src/transaction/inner.rs#L694 

From reading here: https://www.inkandswitch.com/peritext/ we should let the user provide the appropriate ExpandMark for `unmark` too, as it seems like different use-cases may want different behaviour.

A less pragmatic (but maybe cooler) solution would be try and calculate the behavious for unmark such that there are no empty marks and that each remaining mark has the same expand behaviour as the original:

* If the `unmark` lines up with the start or end of a mark with the same name, copy the expand behaviour for that side so it exactly undoes it.
* If one side of the `unmark` is contained within an existing mark, then use the opposite of the other end's behaviour for that mark (so if your unmark ends between the start/end of a mark with ExpandMark::Before, you would not set ExpandMark::After).
* If one side of the `unmark` is not contained within an existing mark, then maybe default to ExpandMark::None.

I'm not sure if the necessary information to calculate this is available to the `unmark` method though; particularly in the case of concurrent edits, the necessary information may not be up-to-date at the time `unmark` is called, so it may require doing this at the time the op tree is being parsed.

ConradIrwin:
This lets users have more control over how marks are removed.

Along the way add a bit of documentation to the marks feature

Fixes #577

### DIFF
--- a/javascript/src/unstable.ts
+++ b/javascript/src/unstable.ts
@@ -268,9 +268,8 @@ export function mark<T>(
 export function unmark<T>(
   doc: Doc<T>,
   prop: stable.Prop,
-  name: string,
-  start: number,
-  end: number
+  range: MarkRange,
+  name: string
 ) {
   if (!_is_proxy(doc)) {
     throw new RangeError("object cannot be modified outside of a change block")
@@ -282,7 +281,7 @@ export function unmark<T>(
   }
   const obj = `${objectId}/${prop}`
   try {
-    return state.handle.unmark(obj, name, start, end)
+    return state.handle.unmark(obj, range, name)
   } catch (e) {
     throw new RangeError(`Cannot unmark: ${e}`)
   }
--- a/javascript/test/marks.ts
+++ b/javascript/test/marks.ts
@@ -24,7 +24,7 @@ describe("Automerge", () => {
       })
 
       doc1 = Automerge.change(doc1, d => {
-        Automerge.unmark(d, "x", "font-weight", 7, 9)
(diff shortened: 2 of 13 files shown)


orionz:
Thanks for doing this!

PR #581, opened 2023-04-21
conflict ordering bug in `doc.get*()`

orionz:
Found a bug where `doc.get[_all]()` would return results in the wrong order when a conflict existed.  This would lead to the wrong conflict being materialized on document load in the javascript client.


orionz:
Included a version bump to the packages with this

Issue #535, opened 2023-02-20
API to enable indexing text document and cursor tracking


munhitsu:
An ask for API.
As per @ept recommendation migrating issue from: https://github.com/automerge/automerge-swift/issues/55

To use Automerge for Text editor one needs to maintain cursor position while remote changes are arriving.
One way of doing it is to expose 2 functions:
- translate from cursor's Int position to immutable Character ID
- translate from immutable Character ID to an int position.

Such 2 APIs will also enable us to index chunks of a live documents. E.g. Tracking chapters.

### discussion
The other way is to manually process remote changes and apply them to the cursor position, but it's duplicating the rust library code and is not solving the text indexing dilemma.

PR #585, opened 2023-04-22
Adding Text field API to translate between usize position and Cursor

munhitsu:
It's implementing #535.

But it's O(N)


alexjg:
I think this is something which would be useful to add. There are two things I can think of that we might want to think about:

1. The most practical concern: https://github.com/automerge/automerge/pull/572 just landed and that may require changes in this code.
2. I don't think we should use the `ExId` for the "address". Instead i think we should introduce an `ExElemId` type (for "external element ID"). Internally we use `crate::types::ElemId` to refer to elements in a sequence. This is different to an object ID in that the "zero" element always refers to the start of the sequence whereas the zero of `ExId` is the Root object. It's also a bit clearer to have a different type for elements of a sequence vs object IDs.

orionz:
Thanks for this.  I wanted to add a lot so you should find a PR for your PR! ;)

munhitsu:
@orionz thanks for awesome contribution. Now it's a nice cheap O(log(n)). Can't wait to use it.
@alexjg comments addressed

alexjg:
This looks great. The only remaining thing is that we need some docs on the new `ReadDoc` methods and on the `Cursor` type describing what this is for.

munhitsu:
I've just added some light touch docs

PR #590, opened 2023-04-26
Add Automerge::diff_utf8 and Automerge::diff_utf16

alexjg:
Instead of `OpObserver::explicit_branch` we add some methods on Automerge and on Autocommit which forward to `Automerge`. This allows us to directly create patches in `automerge_wasm::Automerge::diff` rather than relying on the `Autocommit` to branch the observer and create patches. Users who need more flexible patch generation can still use `Automerge::observe_diff`


orionz:
we decided to go a different route

PR #594, opened 2023-05-05
wasm/js: Run patchCallback in JS, not in webassembly

### LINKED ISSUE #593: Calling `getHeads` on a `PatchInfo.after` document error
The `getHeads` call in the following code throws a `recursive use of an object detected which would lead to unsafe aliasing in rust` error. This is with Automerge `2.0.3`.

```ts
const doc = init<{ count: number }>();

const newDoc = change(
  doc,
  {
    patchCallback: (_, patchInfo) => {
      const heads = getHeads(patchInfo.after); // => error: recursive use of an object detected which would lead to unsafe aliasing in rust 
    },
  },
  (doc) => {
    doc.count = 1;
  },
);

getHeads(newDoc); // <= this works without issue
```

I am aware that `PatchInfo` also contains a `to` property which contains the heads, but I would expect the `after` document to be a fully usable document, which could be passed outside the callback.

alexjg:
This fixes #593. Passing the pointer to the current document into the callback from the webassmebly side is basically never going to be safe because we end up with multiple borrows of the doc. Even if we were to not pass the doc pointer into the callback, the user could do something unsafe by calling another method on a reference to the document retained elsewhere. Instead, we run the callback in JS, so we don't have to worry about re-entrant borrows of the document.

### DIFF
--- a/rust/automerge-wasm/src/lib.rs
+++ b/rust/automerge-wasm/src/lib.rs
@@ -577,18 +577,41 @@ impl Automerge {
     }
 
     #[wasm_bindgen(js_name = applyPatches)]
-    pub fn apply_patches(
+    pub fn apply_patches(&mut self, object: JsValue, meta: JsValue) -> Result<JsValue, JsValue> {
+        let (value, _patches) = self.apply_patches_impl(object, meta)?;
+        Ok(value)
+    }
+
+    #[wasm_bindgen(js_name = applyAndReturnPatches)]
+    pub fn apply_and_return_patches(
         &mut self,
         object: JsValue,
         meta: JsValue,
-        callback: JsValue,
     ) -> Result<JsValue, JsValue> {
+        let (value, patches) = self.apply_patches_impl(object, meta)?;
+
+        let patches: Array = patches
+            .into_iter()
+            .map(interop::JsPatch)
+            .map(JsValue::try_from)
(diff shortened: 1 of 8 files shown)


rust/automerge-wasm/src/lib.rs
```diff
@@ -608,23 +609,27 @@ impl Automerge {
...
-        }
+        let patches = if !return_patches {
+            JsValue::null()
+        } else {
+            let patches: Array = patches
+                .into_iter()
+                .map(interop::JsPatch)
+                .map(JsValue::try_from)
+                .collect::<Result<_, _>>()?;
+            let info = Object::new();
```
orionz, line 620: Info isn't used.  Shouldn't this be removed?

rust/automerge-wasm/src/lib.rs
```diff
@@ -580,15 +580,16 @@ impl Automerge {
     pub fn apply_patches(
         &mut self,
         object: JsValue,
+        return_patches: JsValue,
```
orionz, line 583: I would have made a new method `getAndApplyPatches()` that returns the object value and removed the callback from the other method instead of adding a Boolean argument and complicating the return value for the simple case.

orionz (approved):
Left some comments!  Looks great overall!

PR #599, opened 2023-05-12
Save orphaned changes when saving the whole document

### LINKED ISSUE #595: Automerge.save does not persist orphan changes, potentially leading to a loss of convergence
Hey folks!

First of all, Automerge is amazing, thank you for all your hard work and the commitment to making it open source!

I've been playing with it for an upcoming application I'm working on, and I've encountered a non-intuitive behavior, which might be considered a bug:

> It seems like "orphan" changes are not persisted with `Automerge.save`, is it an expected behavior?

I've put together a small example (also available on [Stackblitz](https://stackblitz.com/edit/stackblitz-webcontainer-api-starter-btubxh?file=index.js) for easy testing, just run `node index.js` in the terminal):

```js
const Automerge = require("@automerge/automerge");
const { createHash } = require("node:crypto");

function sha256(content) {
  return createHash("sha256").update(content).digest("hex");
}

let doc1 = Automerge.init();

doc1 = Automerge.change(doc1, "first change", (doc) => {
  doc.field = "one";
});

let doc2 = Automerge.init();
doc2 = Automerge.merge(doc2, doc1);

doc2 = Automerge.change(doc2, "second change", (doc) => {
  doc.field = "two";
});

const secondChange = Automerge.getLastLocalChange(doc2);

doc2 = Automerge.change(doc2, "third change", (doc) => {
  doc.field = "three";
});

const thirdChange = Automerge.getLastLocalChange(doc2);

console.log("doc1", doc1, sha256(Automerge.save(doc1))); // doc1 { field: 'one' } 614f96017f9cbe826d893d5a2c41251697e5e5bb9b63d2cea3c75ecdf84dbb42
console.log("doc2", doc2, sha256(Automerge.save(doc2))); // doc2 { field: 'three' } f0ac3f0046407885f4984d2b98da735668031f0d3fbaf9a247633061fd0b0ec2

console.log("now we apply the third change");
[doc1, patch] = Automerge.applyChanges(doc1, [thirdChange]);

// Here I would expect the saved Hash to be different, as we added the third change
console.log("doc1", doc1, sha256(Automerge.save(doc1))); // doc1 { field: 'one' } 614f96017f9cbe826d893d5a2c41251697e5e5bb9b63d2cea3c75ecdf84dbb42

// This causes a non-intuitive difference in behavior if we load the documents from persistence

let persistedDoc1 = Automerge.load(Automerge.save(doc1));

console.log("now we apply the second change");
[doc1, patch] = Automerge.applyChanges(doc1, [secondChange]);
[persistedDoc1, patch] = Automerge.applyChanges(persistedDoc1, [secondChange]);

// In theory, I would expect the two to be equal, as both persistedDoc1 and doc1 have been applied the same changes
console.log("doc1", doc1, sha256(Automerge.save(doc1)));  // doc1 { field: 'three' } f0ac3f0046407885f4984d2b98da735668031f0d3fbaf9a247633061fd0b0ec2
console.log(
  "persistedDoc1",
  persistedDoc1,
  sha256(Automerge.save(persistedDoc1))
);  // persistedDoc1 { field: 'two' } 24ff50608b01fcd094118c0a3b3ceabf701b34c0fcf3a1213b0eb71747973679
```

In the example above, I would expect both `doc1` and `persistedDoc1` to have the same content, as they've both encountered the same changes (third and second), despite being out of order.

I might be missing something, but I think this should be considered a bug as it invalidates the convergence guarantee from the [docs](https://automerge.org/docs/cookbook/real-time/#changes-interface) when dealing with persisted/saved documents:

> The counterpart, Automerge.applyChanges(oldDoc, changes) applies the list of changes to the given document, and returns a new document with those changes applied. Automerge guarantees that whenever any two documents have applied the same set of changes — even if the changes were applied in a different order — then those two documents are equal. That property is called convergence, and it is the essence of what Automerge is all about.

Thoughts? 

I'm running the latest released Automerge version: 2.0.3

Again, thank you for all your hard work! And if I can help further in some way, just let me know

alexjg:
Orphaned changes were being omitted when saving the full document. This meant that you could end up with a situation where two documents which have received the same set of messages have different states because one of them had been saved and reloaded whilst containing an orphaned change.

Modify Automerge::save() so that it encodes the orphaned changes as change chunks following the saved document chunk. To allow users to control this behavior also add `Automerge::save_with_options` and `AutoCommit::save_with_options` which provide a way to specify that you wish to discard the orphan changes.

Fixes #595 

@ConradIrwin I had to delete a few of the fuzz crashers to get this to run. Based on their names I suspect they were testing for exactly the thing we are now allowing (the code path which would have previously resulted in a `MissingDeps` error) is that correct?


ConradIrwin:
Yup! Good to delete those now that it doesn’t error

alexjg:
Based on [this discussion in slack](https://automerge.slack.com/archives/C61RJCM9S/p1684239212209099) (with some valuable feedback from @acurrieclark ) I think this needs a little more work. As the PR currently stands it would be possible to load a single change (e.g. the output of `save_incremental`) and not get  an error, instead the document would just be empty. This seems like it would be surprising in quite a lot of scenarios and in general it's probably something that would be considered an application error.

It seems likely to me that the only situation where by default you want to allow missing dependencies is when loading orphaned changes which have been encoded after the initial document chunk. On that basis I think I will modify this to do the following:

1. If the initial chunk is a document chunk allow missing dependencies
2. Otherwise throw an error, but provide a flag to allow loading missing dependencies if the user really wants to

PR #608, opened 2023-05-24
Remove Expose

orionz:
Add logic deal with exposed objects inside the history struct - observers no longer need to deal with it.

Found and fixed two small logic bugs.  When running diff() the path would sometimes error when computing if it was visible or not.  Also the TopOp iterator was reporting conflicts sometimes when they were not there

I added a conflict flag to MapRange and ListRange as I needed them for walking the exposed objects.  The tuple is getting pretty large - it might be time to create a MapRangeValue and ListRangeValue struct.


alexjg (commented):
Some tiny nits and also I think we _should_ add `MapRangeValue` and `ListRangeValue`

PR #610, opened 2023-05-26
Expose marks in the C API

ConradIrwin:
I've tried to copy the patterns from the rest of the C API.

@jkankiewicz it'd be good to check that I'm not leaking memory, as I'm not
really sure how the 'static lifetime thing works.


jkankiewicz:
> I've tried to copy the patterns from the rest of the C API.
@ConradIrwin, thanks for doing that.

>  [I]t'd be good to check that I'm not leaking memory, as I'm not really sure how the 'static lifetime thing works.
I'm looking at it.

jkankiewicz:
> I've tried to copy the patterns from the rest of the C API.
@ConradIrwin, thanks for doing that.

>  [I]t'd be good to check that I'm not leaking memory, as I'm not really sure how the 'static lifetime thing works.
I'm looking at it.

rust/automerge-c/src/doc/mark.rs
```diff
@@ -0,0 +1,294 @@
...
+    After = 3,
+    /// Include text inserted at the start offset.
+    Before = 2,
+    /// Include text inserted at either offset.
+    Both = 4,
+    /// The default tag, not a mark expansion mode signifier.
+    Default = 0,
+    /// Exclude text inserted at either offset.
+    None = 1,
+}
```
jkankiewicz, line 51: @ConradIrwin, if the tags' ordinals aren't meaningful then this declaration can be simplified a bit e.g.
```suggestion
pub enum AMmarkExpand {
    /// Include text inserted at the end offset.
    After = 1,
    /// Include text inserted at the start offset.
    Before,
    /// Include text inserted at either offset.
    Both,
    /// The default tag, not a mark expansion mode signifier.
    Default = 0,
    /// Exclude text inserted at either offset.
    None = 4,
}
```

rust/automerge-c/src/doc/mark.rs
```diff
@@ -0,0 +1,294 @@
...
+    }
+    Default::default()
+}
+
+/// \memberof AMmark
+/// \brief Gets the value of a mark.
+///
+/// \param[in] mark A pointer to an `AMmark` struct.
+/// \return A pointer to an `AMresult` struct with an `AMitem` struct
+///          containing the value.
```
jkankiewicz, line 119: @ConradIrwin, every other function that returns an unpredictable type of value is documented as
```suggestion
/// \return A pointer to an `AMresult` struct with an `AMitem` struct.
```

rust/automerge-c/src/item.rs
```diff
@@ -1133,7 +1218,7 @@ impl TryFrom<&AMitem> for (am::Value<'static>, am::ObjId) {
 /// \enum AMvalType
 /// \installed_headerfile
 /// \brief The type of an item's value.
-#[derive(PartialEq, Eq)]
+#[derive(PartialEq, Eq, Debug)]
```
jkankiewicz, line 1221: @ConradIrwin, is the `Debug` trait necessary now?

ConradIrwin:
@jkankiewicz Thanks for the second pass, I've gone back to your numbering of the item.rs enum. I think this is good to go!

jkankiewicz (approved):
@ConradIrwin, thanks for exposing the marks API to C!

PR #612, opened 2023-06-01
js: Add automerge.diff

alexjg:
Frequently it is useful to get a set of patches which take you from one set of heads in a document to another. Add automerge.diff to achieve this.

### DIFF
--- a/javascript/src/stable.ts
+++ b/javascript/src/stable.ts
@@ -9,6 +9,7 @@ import {
   Counter,
   type Doc,
   type PatchCallback,
+  type Patch,
 } from "./types"
 export {
   type AutomergeValue,
@@ -785,6 +786,16 @@ export function getHistory<T>(doc: Doc<T>): State<T>[] {
       return <T>state
     },
   }))
+}
+
+/**
+ * Create a set of patches representing the change from one set of heads to another
+ *
+ * If either of the heads are missing from the document the returned set of patches will be empty
+ */
+export function diff(doc: Doc<any>, before: Heads, after: Heads): Patch[] {
+  const state = _state(doc)
+  return state.handle.diff(before, after)
 }
 
 /** @hidden */
--- a/javascript/src/unstable.ts
+++ b/javascript/src/unstable.ts
@@ -98,6 +98,7 @@ export {
   toJS,
   isAutomerge,
   getObjectId,
+  diff,
 } from "./stable"
 
 export type InitOptions<T> = {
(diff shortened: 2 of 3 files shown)


annabelle:
@alexjg, This is a great feature and I'm glad it was implemented! I'm wondering if it would be possible to use the `Patch` result given from `diff` and apply it unto a document? 

My end goal here is to effectively "revert" to an older version. My plan was to find the old version using `Automerge.getHistory` then use `Automerge.diff` to create a reversion changeset. Is that the right way to achieve this or is there an alternative way?

If this PR is not the best location to discuss this, I will happily open an issue.

alexjg:
We can definitely will expose an API for applying patches soon yes. What are you trying to achieve though? Often when people want this kind of revert logic they often actually need something like https://github.com/automerge/automerge/pull/617, which adds a `changeAt(doc, headsToChangeAt, changeFn)` which allows you to apply a change to the document as it was at some set of heads. Would that serve you?

annabelle:
Am guessing `changeAt` is analogous to a `git branch` using an earlier version. Would that be a correct assessment? What I'm looking for is an analog to `git revert` of a range, which creates an mirror diff commit to head containing all anti-changes between two versions. I hope the git analogy works here. 

If this helps, here is the code that I whipped up quickly to illustrate what I'm after:
```js
const doc = ...
const history = Automerge.getHistory(doc);

const revertTo = history[_SOME_INDEX_].snapshot;
const patch = Automerge.diff(merged, Automerge.getHeads(doc), Automerge.getHeads(revertTo));

// Ideally, this should be something like `const reverted = Automerge.applyPatch(doc, patch);`
const reverted = Automerge.change(merged, {time: Date.now()}, d => {
  patch.forEach(change => {
    const [attr, idx] = change.path; // quick and dirty path driller
    switch(change.action) {
      case "del":
        d[attr].deleteAt(idx, change.length);
        break;
      case "insert":
        d[attr].insertAt(idx, ...change.values);
        break;
      case ...
    }
  });
});
```

annabelle:
@alexjg, I've noticed that #617 was merged and a [new release tag was cut](https://github.com/automerge/automerge/releases/tag/js%2Fautomerge-2.1.0-alpha.3), however it seems that the version at NPM is missing `changeAt`. I wanted to test if it fits my use case today so we could continue the discussion if needed, but am not able to find `changeAt` in `2.1.0-alpha.3`.

alexjg:
Yeah I'll be cutting a pre-release with `changeAt` this evening. Re. your `revert` usecase, that is an API we want to provide but it's not in yet so tracking and inverting the patches would probably be the best option for now.

PR #616, opened 2023-06-02
[bug] Incorrect `before` documents in PatchCallback when deleting from a list or object

acurrieclark:
When removing an item from a list or deleting an object property, the `before` document in the PatchCallback does not contain the deleted entry.

This PR contains 2 failing tests illustrating the issue


orionz:
Merged this into PR #620 as the fix I already made for that issue overlapped with this issue.  Thanks for the PR!

PR #617, opened 2023-06-02
isolate/integrate/changeAt

orionz:
Ok.  I added some new functionality to Automerge and it comes in three flavors.  First:

```rust
    ...
    let tx = doc.transaction_at(PatchLog::null(), heads);
```
This creates a transaction object that instead of operating at `HEAD` it operates at given heads.  The ReadDoc interface will return state that began at head.   The change will have deps equal to the heads passed in.  Etc... 

Making this work required several changes.

The big one is creating new ActorId's for each level of concurrency.  This is handled in the `fn isolate_actor()`. I did this by prepending some bytes to the primary actor id based on the level of actor id concurrency.  This is obviously inefficient, and could be incorporated in a future version of the binary format, but as for right now will usually mean 1 or 2 extra actor id's in the save file.

Next, I created a new internal read-doc like api that always takes a `clock: Option<Clock>`.  This was needed because now ManualTransaction and AutoCommit might or might not be passing a clock with each call.  As the base and `_at` versions of all the functions were basically the same since the op_set rewrite this was very easy.

Lastly. I modified the `InsertNth` query and `inner_splice` function to also operate correctly if not working on HEAD.   They were written with the assumption that they were they had to change.

Once this was done I went after `AutoCommit`.  It gets two new functions.  `doc.isolate(heads)` and `doc.integrate()`.  One locks autocommit into a chain of transactions `at(heads)` that when complete chain into a new transaction based on the previous one.  This means that non local changes will no longer be seen.  When `integrate` is called the document reverts to HEAD.  The biggest change needed for this was having the PatchLog follow the scope of the document.  This required adding some code to each place that takes in non-local changes, but also to `patch_to()` the new heads when either an `isolate()` or `integrate()` is called.

This really made me feel the power and flexibility of the diff code as all I needed to add was 

```rust
        // we may be isolated so we dont use self.doc.get_heads()
        let before = self.get_heads();
        if before.as_slice() != after {
            let before_clock = self.doc.clock_at(&before);
            let after_clock = self.doc.clock_at(after);
            diff::log_diff(&self.doc, &before_clock, &after_clock, &mut self.patch_log);
        }
    }
```

Finally I added `doc = Automerge.changeAt(doc, heads, (d) => { ... }` to the JavaScript library.  All that was needed was called `handle.isolate(heads)` before running the change function and running `handle.integrate()` afterwards.

This is brand new and does some crazy new things.  Need good feedback for sure.

### DIFF
--- a/javascript/src/constants.ts
+++ b/javascript/src/constants.ts
@@ -4,6 +4,7 @@ export const STATE = Symbol.for("_am_meta") // symbol used to hide application m
 export const TRACE = Symbol.for("_am_trace") // used for debugging
 export const OBJECT_ID = Symbol.for("_am_objectId") // symbol used to hide the object id on automerge objects
 export const IS_PROXY = Symbol.for("_am_isProxy") // symbol used to test if the document is a proxy object
+export const CLEAR_CACHE = Symbol.for("_am_clearCache") // symbol used to tell a proxy object to clear its cache
 
 export const UINT = Symbol.for("_am_uint")
 export const INT = Symbol.for("_am_int")
--- a/javascript/src/internal_state.ts
+++ b/javascript/src/internal_state.ts
@@ -1,6 +1,6 @@
 import { type ObjID, type Heads, Automerge } from "@automerge/automerge-wasm"
 
-import { STATE, OBJECT_ID, TRACE, IS_PROXY } from "./constants"
+import { STATE, OBJECT_ID, CLEAR_CACHE, TRACE, IS_PROXY } from "./constants"
 
 import type { Doc, PatchCallback } from "./types"
 
@@ -25,6 +25,10 @@ export function _state<T>(doc: Doc<T>, checkroot = true): InternalState<T> {
     throw new RangeError("must be the document root")
   }
   return state
+}
+
+export function _clear_cache<T>(doc: Doc<T>): void {
+  Reflect.set(doc, CLEAR_CACHE, true)
 }
 
 export function _trace<T>(doc: Doc<T>): string | undefined {
--- a/javascript/src/proxies.ts
+++ b/javascript/src/proxies.ts
@@ -19,6 +19,7 @@ import {
   TRACE,
   IS_PROXY,
   OBJECT_ID,
+  CLEAR_CACHE,
   COUNTER,
   INT,
   UINT,
@@ -241,6 +242,9 @@ const MapHandler = {
     }
     if (key === TRACE) {
       target.trace = val
+      return true
+    }
+    if (key === CLEAR_CACHE) {
       return true
     }
     const [value, datatype] = import_value(val, textV2)
--- a/javascript/src/stable.ts
+++ b/javascript/src/stable.ts
@@ -242,6 +242,7 @@ export function clone<T>(
   const heads = state.heads
   const opts = importOpts(_opts)
   const handle = state.handle.fork(opts.actor, heads)
+  handle.updateDiffCursor()
 
   // `change` uses the presence of state.heads to determine if we are in a view
   // set it to undefined to indicate that this is a full fat document
@@ -344,6 +345,24 @@ export function change<T>(
   }
 }
 
+export function changeAt<T>(
+  doc: Doc<T>,
+  scope: Heads,
+  options: string | ChangeOptions<T> | ChangeFn<T>,
+  callback?: ChangeFn<T>
+): Doc<T> {
+  if (typeof options === "function") {
+    return _change(doc, {}, options, scope)
+  } else if (typeof callback === "function") {
+    if (typeof options === "string") {
+      options = { message: options }
+    }
+    return _change(doc, options, callback, scope)
+  } else {
+    throw RangeError("Invalid args for changeAt")
+  }
+}
+
 function progressDocument<T>(
   doc: Doc<T>,
   heads: Heads | null,
@@ -373,7 +392,8 @@ function progressDocument<T>(
 function _change<T>(
   doc: Doc<T>,
   options: ChangeOptions<T>,
-  callback: ChangeFn<T>
+  callback: ChangeFn<T>,
+  scope?: Heads
 ): Doc<T> {
   if (typeof callback !== "function") {
     throw new RangeError("invalid change function")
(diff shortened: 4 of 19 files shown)


ept:
Awesome! Just a question, since I'm not in the details…

> I modified the InsertNth query and inner_splice function to also operate correctly if not working on HEAD.

How do you count list indexes when not working on HEAD? Since different list elements will be visible on different branches, it seems to me like the optree will need to have some per-branch metadata about the number of visible elements on each tree node. Otherwise Nth queries would require a linear scan over all the list elements. Is that right?

orionz:
> How do you count list indexes when not working on HEAD?

When not working on HEAD I cant use the index information.  I have to walk every op in the list and say "is this visible at this clock".   Basically ops are `O(n)` when isolated vs `O(log n)` when not.  A slightly evolved version of this with explicit long lived branched - it would make sense to build and maintain an index  per branch.

alexjg (commented):
It is gratifying how neatly this has fallen out of the diff work. I left a few tiny comments but I have a bigger thought about the API.

This implementation is effectively anonymous branches. The branch is represented
implicitly by the current state of the `AutoCommit`. I.e. when you call
`AutoCommit::isolate` the `ReadDoc` and `Transactable` methods on `AutoCommit`
refer to the anonymous branch until `AutoCommit::integrate` is called. This
makes things like `receive_sync_message` a little awkward and also makes it
hard to implement things like chaining changes in an isolate on top of each
other in JS a bit awkward to express.

What if we went a bit further down the branches route? Instead of a single 
`Isolate` on the `AutoCommit` we instead have a `HashMap<BranchId, Isolate>`.
The `ReadDoc` and `Transactable` implementations on `AutoCommit` continue to
operate as they always have done and to access branch specific stuff you call
`AutoCommit::on_branch(BranchId)`. Every branch manages it's own diff cursor.

I'm wondering if an API something like this would work:

```rust
let some_old_heads: Vec<ChangeHash> = ...
let mut doc: AutoCommit = ...
let my_branch = doc.create_anon_branch(&some_old_heads)?;

let text_id: ObjId = ...
// insert some text on the branch
doc.on_branch(&my_branch).splice(&text_id, 0, 0, "hello")?;
// read the text back
println!("The text is {}", doc.on_branch(&my_branch).text(&text_id))
// generate diffs for the inserted text. Note that this requires each branch
// to have it's own diff cursor and may also imply we want a trait for the
// `diff_*` methods currently on `AutoCommit`
doc.on_branch(&my_branch).diff_incremental()

// Finally integrate the branch
doc.integrate_branch(&my_branch)
```

I think this allows us to have anonymous branches in JS as well,
`automerge.anonymousBranch(heads)` just saves some additional state in the
javascript frontend to represent the branch ID and then we check for a branch
ID in `change` etc

orionz:
Feedback on the API ideas.

```rust
    doc.on_branch(&my_branch).splice(&text_id, 0, 0, "hello")?;
```

You're going to have one change/transaction per operation here almost no matter what.  This seems almost identical  to `doc.transaction_at(heads)` except maybe these branches autocommit when dropped and update the heads associated with a branch id.

It isn't clear to me now what `integrate_branch()` does.  Isn't it just `drop_branch()`

What would the code in `changeAt()` look like?   Would you need to manually do the `patch_to()` and from?

What about:

```rust
     doc.create_temp_branch("hello123", heads);
     doc.checkout_temp_branch("hello123")     // calls patch_to under the hood
     let b in doc.branches() {
        log!(" branch {:?}", b);
     }
     doc.splice(obj, "text", 10, 0, "hello");
     if is_ok() {
         doc.merge_branch("hello123");  // calls patch_to
     } else {
         // I think it would be easy to add code to make changes in a branch NOT appear in 
         // getHeads() and therefore be droppable b/c nothing will ever depend on them
         doc.drop_branch("hello123")
     }

```

alexjg:
What I was thinking when sketching the API above was that maybe there is  a way to support multiple branches at once without either a) cloning the doc or b) have to constantly patch around between different frontend states. It seems like it would be extremely useful to be able to have two different patches live at the same time.

On that basis I was thinking that a branch could be like this:

```rust
pub struct AutoCommit {
    ...
    branches: HashMap<BranchID, BranchState>,
}

struct BranchState {
    patch_log: PatchLog,
    diff_cursor: Vec<ChangeHash>,
    heads: Vec<ChangeHash>,
    tx: TransactionInner,
}

struct Branch<'a> {
    doc: &'a Automerge, 
    state: &'a mut BranchState,
}

impl<'a> Transactable<'a> for Branch<'a> {
    // implement in terms of TransactionInner and generally behave like autocommit
}

impl<'a> ReadDoc for Branch<'a> {
    //implement in terms of `Automerge` 
}

impl<'a> Branch<'a> {
    fn diff_incremental(&mut self) -> Option<Patch> {
        // same as AutoCommit::diff_incremental but using the patch log and 
        // diff cursor from self.state
    }
}

impl AutoCommit {
    fn on_branch<'a>(&'a mut self, branch: BranchId) -> Option<Branch<'a>> {
        self.branches.get_mut(&branch).map(|state| Branch { doc: &self.doc, state })
    }

    fn branch(&mut self) -> BranchId {
        ...
    }

    fn integrate_branch(&mut self, branch: BranchId) {
        // commit the transaction in self.state
    }

}
```

This would then mean that every function in the `automerge_wasm` interface could take an optional branch ID argument, in which case rather than calling `AutoCommit::put` (for example) it would call `AutoCommit::on_branch(branch_id)::put` and rather than calling `AutoCommit::diff` it would call `AutoCommit::on_branch::diff`.

I think this means you could then store the branch ID in the meta state on the JS side, so the proxies can grab the branch ID out of the meta state and pass it to WASM when making changes. Which in turn means you could  have multiple branches live at once.

```typescript
const doc = // some automere doc

// This creates a new branch using `AutoCommit::branch`, copies the materialized state of the doc and saves the 
// branch ID in a hidden key on that state
let branch1 = automerge.anonymousBranch(doc) // creates a new branch 

// This passes the hidden branch ID to `automerge_wasm::Automerge::splice_text`
branch1 = automerge.change(branch, d => {d.foo = "barbar"})

let branch2 = automerge.anonymouseBranch(doc)

// Now branch1 and branch2 are both available to examine and compare.
```

But having written all this out it a) does seem like an awful lot of work and maybe we should leave it for another day and b) definitely needs more thought than I have given it.

Issue #619, opened 2023-06-03
automerge-js: text v1 duplicates emojis on load from 2.0.1 onwards


ConradIrwin:
I am not yet sure why this is the case, but the following test case fails with @automerge/automerge master, 2.0.3, 2.0.2, 2.0.1. It works correctly on 2.0.0.

```javascript
  it("should support unicode when creating text", () => {
    s1 = Automerge.from({
      text: new Automerge.Text("🐦"),
    })
    assert.strictEqual(s1.text.get(0), "🐦")

    s2 = Automerge.load(Automerge.save(s1))
    assert.strictEqual(s2.text.toString(), "🐦") // => FAIL, got "🐦🐦"
  })
```

I have narrowed it down to the loading side of things (loading the saved document in rust/go is fine; and conversely loading a document created by rust/go into javascript is also broken).

It will duplicate all emojis at any position in the document (this is just a minimal test case).

PR #620, opened 2023-06-03
Failing test for #619

ConradIrwin:


orionz:
`export_list()` needs to generate a list of values but is using `length()` and `get(obj,index)` to build this list - this doesn't work for utf16 text b/c there is not a 1-to-1 mapping between indexes and internal objects.  The best way with the current api to walk the values correctly with either list or text is with `list_range()` - perhaps we should disallow `get(obj,index)` for text() as what it returns is not obvious.   Perhaps instead provide a `text_slice(obj, range)` for grabbing spans of text - as this will make more sense for text objects and wont cause the confusion that get() does.

orionz:
Also note - this could arise on utf8 if somehow a Scalar::String() gets inserted into a text object.  This can't happen using the `splice()` api but the file format does not disallow it.

ConradIrwin:
Thanks for the fix!

PR #622, opened 2023-06-05
[Proposal] New Change API

acurrieclark:
While I think this might actually be usable as is, this is more of a thought exercise as to how the `Extend` type might be phased out of the `change` function.

### The Problem
As it stands, the change function takes an `Extend<T>` type, which exists purely to decorate any array elements in the Automerge document tree as `List`s. The `List` type includes the `insertAt` and `deleteAt` methods, which do not exists on the `Array` type.

The `Extend` type wreaks enormous typescript havoc, most noticeably when trying to infer document types when using helper functions or when trying to initialise a new array within a `change` callback.

### The Proposal
This seems the most straightforward path to weaning ourselves off `List`'s additional functionality. Here I have:

- Added 2 utility functions which replace the `List` methods, but maintain a similar API. These are to be used inside a change callback
  - insertAt<T>(list: T[], index: number, ...values: T[])
  - deleteAt<T>(list: T[], index: number, numDelete?: number)
- Added a `change` function to `unstable`, with an updated `ChangeFn` type. This uses the document type `T` directly.

Only small changes were required to the existing `change` function to achieve this, so this has involved a level of duplication. In particular, I had to completely copy the `progressDocument` function, but this could easily be extracted to avoid this.

All existing tests pass and I have added a few additional ones to cover the new behaviour. No breaking changes that I am aware of; even swapping `change` out for `unstable.change` would work directly as the `List` methods are still present on the proxy. Essentially, typescript would complain, but the code would run.

I know this will be of use to me (especially as I am working in the `unstable` hinterlands already) so I am sure it will be of use to others.


alexjg:
I think we should actually modify the stable `ChangeFn` to use `T` directly. This will continue to work just fine for JS users, for typescript users they would need to remove the `Extend` type annotation anywhere they are annotating the type of the argument to the change function. This is technically a breaking change, but it doesn't break the runtime semantics (because `insertAt` still exists) and I think almost everyone who is using typescript will have banged their head against `Extend` and be very happy to see it gone.

acurrieclark:
Done. Provided the tests all pass, the `stable` change function API should now be updated to take an argument of type <T>.

I have also moved the new `insertAt` and `deleteAt` utility functions into `stable`

Issue #601, opened 2023-05-13
Large performance regression on maps


jeffa5:
Trying out the latest Automerge pieces in my project (which uses lots of maps) I was surprised to find a large performance regression, it seems to fundamentally come down to:

https://github.com/automerge/automerge/blob/696a1278bd67d4085f0e0d318caa0763c06d808b/rust/automerge/src/op_tree.rs#L350

When we do a `get` we used to have a query that would intelligently skip over nodes that had no visible elements, this now visits every operation.

This is particularly bad in my case where we have lots of operations to a single key as it has to do a linear scan over all of them each time, rather than quickly skipping to the end.

Is there a chance of bringing back this performance? It should also have been visible in the map benchmarks, though they don't run in CI, https://github.com/automerge/automerge/blob/696a1278bd67d4085f0e0d318caa0763c06d808b/rust/automerge/benches/map.rs#L51

jeffa5:
https://github.com/automerge/automerge/commit/a9612371e01b6e3cd4cb13bfcd4b9f25b63958ab#diff-6d231a8ae883941c8db1c7a991f754a90277721095749fa61339b697fdae2271 seems to have squashed the optimisations before the prop file was removed.

alexjg:
Hmm, well it's upsetting that we didn't spot this and it points to a need to get the benchmarks running in CI somehow. 

It would be nice to avoid using the `Query` infrastructure if we can. The reason to move to the iterator based approach was that we were duplicating bugs all over the place due to the low level API `Query` uses being pretty hard to abstract. Would it suffice here to have an iterator which only visits visible ops to pass to `OpsFound::new`? I think such an iterator would be generally useful anyway (we have `TopOps` but that still vsits every op, it just filters some out).

orionz:
> However, since the Prop query was removed we'd need to re-add it, perhaps this could be done?

I don't think this is it...

Old code was:

```rust
        let start = binary_search_by(child, ops, |op| m.key_cmp(&op.key, &self.key));
        self.pos = start;
        QueryResult::Skip(start)
```

New code is:

```rust
                let pos = self.binary_search_by(|op| meta.key_cmp(&op.key, &key));
                Some(OpsFound::new(pos, self.iter(), key, clock))
           // then inside the new(func) it calls
                let mut next = iter.nth(start_pos);
```

which should do the exact same thing as `skip`

Will dig in further

jeffa5:
[...]

PR #623, opened 2023-06-05
Fixes performance regression on maps (#601)

orionz:
The old code was optimized to be fast if there were many props with few updates but not few props with many updates.  This is optimized for both.


orionz:
```
   map/repeated put/10000  time:   [20.958 ms 21.070 ms 21.185 ms]
                        thrpt:  [472.04 Kelem/s 474.61 Kelem/s 477.14 Kelem/s]
                 change:
                        time:   [-92.308% -92.247% -92.188%] (p = 0.00 < 0.05)
                        thrpt:  [+1180.1% +1189.8% +1200.1%]
                        Performance has improved.
```

Issue #603, opened 2023-05-17
proposal: Cursor API for text editing


ConradIrwin:
As observed in #602, it's surprisingly difficult to get indexing right in unicode strings in some languages (in this case javascript, but go has similar issues). It's also hard to use indexes to edit a document in the face of concurrent changes being synced (you have to continually recalculate what your index is).

Under the hood automerge inserts do not actually use the index under the hood, they convert it into a "after" relationship with an element that already exists in the list. We could probably make both problems less obnoxious by exposing an API like a cursor that represents a position in a list.

The cursor should "float" so that if new content is inserted (or old content deleted) at the position of the cursor, it moves around, so that the same cursor can be re-used if the user continues to edit at the same point.

```
doc.content = new automerge.Text()
doc.content.insertAt(0, ...'hello world')
let cursor = doc.content.cursor(/* todo */)
doc.content.splice(cursor, 0, ' cool') // hello cool| world
// *or*
doc.content.splice(cursor, 5, ' moon') // hello moon|
// *or*
doc.content.splice(cursor, -5, 'goodbye') // goodbye| world
```

There are a whole bunch of ways it might be useful to create a cursor, but the important ones are:
* From an automerge index `doc.content.cursor({index: 5})`
* From a utf-16/utf-8 offset `doc.content.cursor({utf8length: 5} // {utf16length: 5})`

It might also be helpful to have helpers:
* `doc.content.cursor({from: cursor, index: 5})` to jump 5 codepoints forward
* `doc.content.cursor({from: cursor, utf8length: -5})` to jump backwards a number of codepoints that makes up 5 utf-8 bytes.

Another idea would be to try and avoid passing integers and allow users to pass the string so that automerge can do the calculation:
* `doc.content.cursor({after: 'hello'})` (though this assumes that the passed argument is a prefix of the document).

The advantage of using a cursor for something like this is that you only need to recalculate the position once (and when the cursor is moved), which should reduce the impact of all the code needed to figure out which index is which.

ept:
This seems like a good idea! For a collaborative text editor that wants to render other users' cursor positions, it would also be good to be able to serialize a cursor into a string that can be shared with collaborators, and turned back into an index for rendering at the recipients.

orionz:
I exposed the low level cursor api via the JS interface.  Any chance this will fit your needs?

ConradIrwin:
Which branch is that on? (grepping for `cursor` in the javascript or automerge-wasm folders returns nothing on main).

[...]

PR #624, opened 2023-06-06
add getCursor/getCursorPosition allow use in js::splice()

orionz:
We added a cursor api to the rust a while back - exposing it to wasm and js.  Currently only the unstable `Automerge::splice()` can take the cursor as an index, however the  `Automerge::getCursorPosition()` call can be used anywhere else.  Added this because of #603

### DIFF
--- a/javascript/src/unstable.ts
+++ b/javascript/src/unstable.ts
@@ -51,12 +51,13 @@ export {
   type ScalarValue,
 } from "./unstable_types"
 
-import type { Mark, MarkRange, MarkValue } from "./unstable_types"
+import type { Cursor, Mark, MarkRange, MarkValue } from "./unstable_types"
 
 import type { ScalarValue, PatchCallback } from "./stable"
 
 import { type UnstableConflicts as Conflicts } from "./conflicts"
 import { unstableConflictAt } from "./conflicts"
+import type { InternalState } from "./internal_state"
 
 export type {
   PutPatch,
@@ -66,6 +67,7 @@ export type {
   IncPatch,
   SyncMessage,
   Heads,
+  Cursor,
 } from "@automerge/automerge-wasm"
 
 export type { ChangeOptions, ApplyOptions, ChangeFn } from "./stable"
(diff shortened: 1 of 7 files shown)


rust/automerge-wasm/src/lib.rs
```diff
@@ -804,6 +804,43 @@ impl Automerge {
...
+        heads: Option<Array>,
+    ) -> Result<String, error::Cursor> {
+        let (obj, obj_type) = self.import(obj).unwrap_or((ROOT, am::ObjType::Map));
+        if obj_type != am::ObjType::Text {
+            return Err(error::Cursor::InvalidObjType(obj_type));
+        }
+        let index = index as usize;
+        let heads = get_heads(heads)?;
+        let cursor = self.doc.get_cursor(obj, index, heads.as_deref())?;
+        Ok(cursor.to_string())
```
alexjg, line 821: Rather than using the `Display` implementation here should we consider using the base64 encoding of the serialization format we defined in `automerge/src/cursor.rs`? The advantage of that approach is that we have a version tag in the format so if we want to change how cursors are represented (for ex. to include a branch ID) we can do so transparently.

javascript/src/unstable.ts
```diff
@@ -237,13 +238,52 @@ export function splice<T>(
     throw new RangeError("invalid object for splice")
   }
   const value = `${objectId}/${prop}`
+  if (typeof index == "string") {
```
alexjg, line 241: This feels like it could produce extremely confusing error messages if you were to accidentally pass a string as the `index` but were not intending to pass a cursor. Could we instead use a wrapper class for `Cursor` so we can match using `instanceof` and so that users can't accidentally pass in strings?

rust/automerge-wasm/src/lib.rs
orionz: At the end of the day it's on opid.  Being able to see a string on the js side and having it look like "10@aabbcc", seeing the actor and what not, feels consistent clear and transparent and consistent with how we're handling object ids.

javascript/src/unstable.ts
orionz: We could wrap it in a class.  It would make handling it harder - you can't just save it to a database or to disk - we'd need a `toString()` and ` new Cursor(string)` and some documentation.    Maybe instead I ad a regex to make sure the string is in `ctr@actor` form on the branch?

PR #626, opened 2023-06-06
Handle splitting string into code points (Fixes #602)

### LINKED ISSUE #602: Javascript insertAt API lets you get document into a confusing state
I'm building a VSCode extension that keeps an automerge document up to date, I ended up doing this:

```
    this.doc = automerge.change(this.doc, d => {
      for (let c of changes) {
          d.content.deleteAt(c.rangeOffset, c.rangeLength)
          d.content.insertAt(c.rangeOffset, c.text))
      }
    })
```

This works fine for single-character changes, but there's a bug. For `insertAt` I need to actually do something like: `d.content.insertAt(c.rangeOffset, ...c.text.split('')))` because I need to pass each character one at a time to automerge otherwise the offsets into the document get messed up.

A more reduced example is this:

```ts
import * as automerge from '@automerge/automerge'
let doc = automerge.init<{content: automerge.Text}>()
doc = automerge.change(doc, d => {
  d.content = new automerge.Text()
  d.content.insertAt(0, 'ten')
})

console.log(doc.content.toString()) // => "ten"
console.log(doc.content.length) // => 1 !!!!
```

I would expect the previous code to act as though I had done this instead:
```
import * as automerge from '@automerge/automerge'
let doc = automerge.init<{content: automerge.Text}>()
doc = automerge.change(doc, d => {
  d.content = new automerge.Text()
  d.content.insertAt(0, 't', 'e', 'n')
})

console.log(doc.content.toString()) // => "ten"
console.log(doc.content.length) // => 3
```

As I understand it the goal is to have one index per unicode codepoint (I wrote up my understanding [here](https://pkg.go.dev/github.com/automerge/automerge-go#Text)). It turns out this isn't very intuitive in javascript!

What I'd like to make work is:
```
d.content.insertAt(0, '🧟‍♀️')
```

This should insert four unicode codepoints (U+1FD9F (Zombie), U+200D (zero-width joiner), U+2640 (female), U+FE0F (variation selector)):

This took me a while to find
* `'🧟‍♀️'.split()` gives a list of length one: '🧟‍♀️'; 
* `'🧟‍♀️'.split('')` gives a list of length five:  (U+D83E (first surrogate), U+DDDF (second surrogate), U+200D (zero-width joiner), U+2640 (female), U+FE0F (variation selector)).
* [...'🧟‍♀️'] gives the list I need (though I would not have thought to try that without [this](https://stackoverflow.com/questions/54369513/how-to-count-the-correct-length-of-a-string-with-emojis-in-javascript)).

In summary I think we should change the API of the text .insertAt() so that `d.content.insertAt(0, '🧟‍♀️')` acts the same as `d.content.insertAt(0, ...'🧟‍♀️')`, and correctly ensures that the document has one index per unicode codepoint.

At the same time I think we should change the API to throw an error if you do `d.content.insertAt(0, ...'🧟‍♀️'.split(''))` because this ends up inserting two surrogate pairs into the document, which come back out as unicode replacement characters instead of a zombie.

Even if we do fix this, there's still an issue for users to figure out the correct index to use, but that seems solvable separately.

ConradIrwin:
In the next major release of automerge, we'll expose the new text
API; but in the meantime, let's fix this footgun in the existing
implementation.

I decided against adding yet another API (.insertTextAt) because
it seems likely that it'd be deprecated in the next release anyway.

### DIFF
--- a/javascript/test/text_v1.ts
+++ b/javascript/test/text_v1.ts
@@ -68,7 +68,7 @@ describe("Automerge.Text", () => {
   it("should allow modification before an object is assigned to a document", () => {
     s1 = Automerge.change(Automerge.init(), doc => {
       const text = new Automerge.Text()
-      text.insertAt(0, "a", "b", "c", "d")
+      text.insertAt(0, "abcd")
       text.deleteAt(2)
       doc.text = text
       assert.strictEqual(doc.text.toString(), "abd")
@@ -293,5 +293,31 @@ describe("Automerge.Text", () => {
     let s4 = Automerge.from({ some: "value" })
     s4 = Automerge.merge(s4, s2)
     assert.strictEqual(s4.text.toString(), "🐦")
+  })
+
+  it("should let you insert strings", () => {
+    s1 = Automerge.from({
+      text: new Automerge.Text(""),
+    })
+
+    s1 = Automerge.change(s1, d => {
+      d.text.insertAt(0, "four")
+    })
+
+    assert.strictEqual(s1.text.length, 4)
+  })
+
+  it("should index by unicode code points", () => {
+    s1 = Automerge.from({
+      text: new Automerge.Text(""),
+    })
+
+    s1 = Automerge.change(s1, d => {
+      d.text.insertAt(0, "🇬🇧")
+      d.text.insertAt(2, "four")
+    })
+
+    assert.strictEqual(s1.text.length, 6)
+    assert.strictEqual(s1.text.toString(), "🇬🇧four")
   })
 })
--- a/javascript/src/proxies.ts
+++ b/javascript/src/proxies.ts
@@ -995,6 +995,13 @@ function textMethods(target: Target) {
       const text = context.text(objectId)
       return text.indexOf(o, start)
     },
+    insertAt(index: number, ...values: any[]) {
+      if (values.every(v => typeof v === "string")) {
+        context.splice(objectId, index, 0, values.join(""))
+      } else {
+        listMethods(target).insertAt(index, ...values)
+      }
+    },
   }
   return methods
 }
--- a/javascript/src/text.ts
+++ b/javascript/src/text.ts
@@ -128,7 +128,11 @@ export class Text {
         "object cannot be modified outside of a change block"
       )
     }
-    this.elems.splice(index, 0, ...values)
+    if (values.every(v => typeof v === "string")) {
+      this.elems.splice(index, 0, ...values.join(""))
+    } else {
+      this.elems.splice(index, 0, ...values)
+    }
   }
 
   /**


orionz:
I really want to stop changing text objects with any API call other than splice(string).  I've considered making it throw an error and will likely depricate it soon.  Perhaps you could try this

```js
    insertAt(index: number, ...values: any[]) {
      context.splice(objectId, index, 0, values.join(""))
    },
```

and this

```js
  insertAt(index: number, ...values: Array<string>) {
     if (this[STATE]) {
       throw new RangeError(
         "object cannot be modified outside of a change block"
       )
     }
    this.elems.splice(index, 0, values.join(""))
   }
```

instead.  This way the process of breaking up the string always happens inside the engine and non-strings will never end up inside a text object.

orionz:
Thanks for the work - The rollout plan you suggest looks pretty close to what I was imagining.

javascript/test/text_v1.ts
```diff
@@ -294,4 +294,30 @@ describe("Automerge.Text", () => {
...
+  })
+
+  it("should index by unicode code points", () => {
+    s1 = Automerge.from({
+      text: new Automerge.Text(""),
+    })
+
+    s1 = Automerge.change(s1, d => {
+      d.text.insertAt(0, "🇬🇧")
+      d.text.insertAt(2, "four")
```
annabelle, line 317: I have a question which may be my misunderstanding but I would have imagined the insertion here would be under index 4 as `"🇬🇧".length` is 4. 

What is the 2 here representing? If this was in fact insertion mid-codepoints, I would've expected `🇬four🇧` as the emoji `🇬🇧` is composed of those two regional indicators.

Can you please explain this test? mostly from the standpoint of a user of the library, as I have different expectations of what should happen here.

ConradIrwin: It comes down to the way that Unicode handles flags, and the way that JavaScript handles Unicode.

The flag emoji is rendered as one glyph, but under the hood is represented by two unicode codepoints, (see https://emojipedia.org/flag-united-kingdom/), U+1F1EC and U+1F1E7.

automerge uses list positions based on codepoints and so 2 represents the position after the flag.

JavaScript was created early on in the history of Unicode when it was believed that there would be a maximum of 2**16 characters; and so it represented strings using a legacy encoding known as UCS-2 which uses two bytes per code point.

When Unicode realized that they needed more characters the UTF-16 encoding was created to be backward compatible with UCS-2. UTF-16 like UCS-2 uses two bytes to represent each “code unit”. For many characters there is an exact equivalence where one code unit represents the code point with the same value; but to represent more than 2**16 code points, some Unicode “code points” (including both the code points in the flag) are represented as a pair of two code units (see https://learn.microsoft.com/en-us/globalization/encoding/surrogate-pairs).

So the flag emoji is 4 code units long, and two code points long, but only conceptually one user perceivable glyph.

ConradIrwin: I should also have mentioned that there is a plan to ship a new way to interact with automerge text that will let you use JavaScript string indices, but I’m not exactly sure when that will ship.

In the meantime, I’m using https://www.npmjs.com/package/code-point-mapping.

Issue #609, opened 2023-05-26
Empty marks are returned by marks()


ConradIrwin:
This may be related to #604; currently if you create a mark and then clear half of it then marks will report an empty mark at the start of the string.

This may be useful in the case where inserting text at that location will inherit the mark, but (as in https://github.com/automerge/automerge/commit/f6cefc235e1dcec138ee447a07bf6f690e3ae20b#diff-d822dff8eea94f7201e80ed6a88e0bf89b5f10970198f586cb16ccae63b7f1ceR1714-R1719) it's not useful when newly added text is not affected.

As the expand property of a mark is not introspectable in the current API, it'd be nice to only return marks that would expand if text was inserted there.

orionz:
This gets fixed by #629.  Will be merged soon

Issue #604, opened 2023-05-17
Automerge may attach deleted mark to new inserted text


zxch3n:
Steps to reproduce:

```rust
#[test]
fn mark() {
    let mut doc = Automerge::new();
    let mut tx = doc.transaction();
    let text_id = tx.put_object(&ROOT, "text", ObjType::Text).unwrap();
    tx.splice_text(&text_id, 0, 0, "hello world").unwrap();
    let mark = Mark::new("bold".to_string(), true, 2, 8);
    tx.mark(&text_id, mark, ExpandMark::After).unwrap();
    let mark = Mark::new("link".to_string(), true, 3, 6);
    tx.mark(&text_id, mark, ExpandMark::None).unwrap();

    tx.splice_text(&text_id, 1, 10, "").unwrap(); // 'h'
    dbg!(tx.text(&text_id).unwrap(), tx.marks(&text_id).unwrap());
    tx.splice_text(&text_id, 0, 0, "a").unwrap(); // 'ah'
    dbg!(tx.text(&text_id).unwrap(), tx.marks(&text_id).unwrap());
    tx.splice_text(&text_id, 2, 0, "a").unwrap(); // 'ah<bold>a</bold>'
    dbg!(tx.text(&text_id).unwrap(), tx.marks(&text_id).unwrap());
}
```

Before the last insertion, the text is `ah` with no mark. So the expected marks after the last insertion should be empty as well. But actually the inserted 'a' has a bold mark.

It's caused by the Peritext algorithm. I describe the suggested fix here https://github.com/inkandswitch/peritext/issues/32

orionz:
Found a good fix for this.  Basically when scanning tombstones ignore all mark begin/end pairs or marks between begin/end pairs

alexjg:
Fixed by #629

Issue #539, opened 2023-03-01
Identifying conflicts after merge


mpareja:
Is there a way to detect conflicts encountered during an `Automerge.merge` operation without having to walk the entire document tree calling `Automerge.getConflicts`?

I tried hooking into `patchCallbacks` but it does not include patches when the `local` doc's changes are chosen based on lamport timestamp.

alexjg:
There's no API for it built into the document yet, though I think we should add one. Can you post an example of the problem with the `patchCallback` not being called on a conflict? It _should_ be I think so this might be a bug.

mpareja:
Below is a sample that can easily be turned into a test.

```js
let doc1 = Automerge.from({ str: 'first' }, 'b'); // actorId is later than 'a'
let doc2 = Automerge.clone(doc1, 'a');
doc1 = Automerge.change(doc1, (d) => d.str = 'second');
doc2 = Automerge.change(doc2, (d) => d.str = 'third');

const [merged, patches] = mergeAndGetPatch(doc1, doc2);

console.log('CHOSEN', merged.str); // 'second' - doc1's value is chosen because 'b' actor id is later than 'a'
console.log('PATCHES', patches); // [] - no patches recorded since doc1's value is chosen
console.log('CONFLICTS', Automerge.getConflicts(merged, 'str')); // { '2@a1': 'third', '2@a2': 'second' }
```

Regarding your comment about `patchCallback`, I suppose I'm not entire sure how others are using it. I can imagine an argument for the way it's implemented since _nothing actually changed in the document_.

For our use case, it would be convenient if `patchCallback` included the entry. Further to that, it would be useful if I could provide `MergeOptions` like so:

```js
Automerge.merge(local, remote, { patchCallback })`
```

mpareja:
Whoops, missed a function in that test code:
```js
export function mergeAndGetPatch<TData>(
  localDoc: Automerge.Doc<TData>,
  remoteDoc: Automerge.Doc<TData>
): [Automerge.Doc<TData>, ModelPatch[]] {
  // it's recommended we clone this or else the original is not mutable...
  const remoteClone = Automerge.clone(remoteDoc);

  let allPatches = [] as ModelPatch[];
  const patchCallback = (patches: Automerge.Patch[]) => {
    allPatches = allPatches.concat(patches);
  };

  // create a clone with tracking (don't know how else to listen for merges)
  const localWithTracking = Automerge.merge(Automerge.init<TData>({ patchCallback }), localDoc);
[...]

PR #629, opened 2023-06-08
rework marks & patches

orionz:
Done:

* rework how patches and marks interact
* inserts & splice patches now contain all current marks
* wrote a [PATCH.md](https://github.com/automerge/automerge/blob/patch_marks/rust/automerge-wasm/PATCH.md)
* patch marks are not more predictable (sort by `name`, `start`)
* expose conflict information in JS patches (#539)
* several bugs around marks (empty marks) are gone (#609)
* fixed an issue where deleted marks would reappear (#604)

Todo:

* ~~LOTS of code has changed - need more tests to make sure all new code paths are covered~~
* ~~performance tests - this should have negligible impact on mark-less code but need to confirm~~
* ~~several FIXME's need to be resolved~~

### DIFF
--- a/rust/automerge-wasm/PATCH.md
+++ b/rust/automerge-wasm/PATCH.md
@@ -1,0 +1,189 @@
+
+## Automerge Patchs
+
+### Path
+
+All patches have a path.  This path describes the set of properties that need to be traversed to reach the intended target.
+
+For instances.  If a document looked like this
+
+```js
+{
+  animals: [
+      {  name: "Lion",
+         genus: "Panthera",
+      },
+      {  name: "Chimpanzee",
+         genus: "Panthera",
+      }
+  ]
+}
+```
+
+And a patch came to change the genus of Chimpanzee to "Pan" it would look like this.
+
+```js
+{
+  action: "put",
+  path: [ "animals", 1, "genus" ],
+  value: "Pan",
+}
+```
+
+### Values
+
+Patch values can be
+
+```ts
+type PatchValue = string | number | boolean | null | Date | Uint8Array | {} | []
+```
+
+These stand in for the different scalar types in Automerge plus an empty object or an empty list.
+
+### Put
+
+Changing a single key in a map or index in a list.  This never changes the length of a list.
+
+```ts
+type PutPatch = {
+  action: 'put'
+  path: Prop[],
+  value: PatchValue,
+  conflict?: boolean
+}
+
+let patch : PutPatch = {
+  action: "put",
+  path: [ "config", "enabled" ],
+  value: true,
+  conflict: true, // optional
+}
+```
+
+The conflict field indicates that another conflict value is present which can be queried via an api call.
+If the field is missing, it's assumed to be false.
+
+### Insert
+
+Inserts one or more values into a list.  This increases the length of the list by the given amount.
+
+```ts
+type InsertPatch = {
+  action: 'insert'
+  path: Prop[],
+  values: PatchValue[],
+  conflicts?: boolean[]
+}
+
+let patch : InsertPatch = {
+  action: "insert",
+  path: [ "emoji", 3 ],
+  values: [ "😍", "🔥", "🎆" ],
+  conflicts: [ false, true, false ] // optional
+}
+```
+
+If the `conflicts` field is missing, it's assumed to be all false.
+
+### Delete
+
+Delete patches remove a key from a map or one or more elements from a list or text object.
+
+```js
+
+type DelPatch = {
+  action: 'del'
+  path: Prop[],
+  length?: number,
+}
+
+let patch : DelPatch = {
+  action: "del",
+  path: [ "items", 3 ],
+  length: 10, // optional
+}
+```
+
+The length field is present only on sequences and when there is a run of consecutive deletes.
+
+### Inc
+
+Increment a number by 'value`
+
+```ts
+type IncPatch = {
+  action: 'inc'
+  path: Prop[],
+  value: number
+}
+
+let patch : IncPatch = {
+  action: "inc",
+  path: [ "config", "logins" ],
+  value: 1
+}
+```
+
+### Splice
+
+Splice a string into a text object
+
+```ts
+type SpliceTextPatch = {
+  action: 'splice'
+  path: Prop[],
+  value: string,
+}
+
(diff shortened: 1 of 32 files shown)


rust/automerge/src/automerge/current_state.rs
```diff
@@ -95,31 +138,30 @@ fn log_list_patches<'a, I: Iterator<Item = &'a Op>>(
     obj: &ObjId,
     ops: I,
 ) {
-    //let exid = doc.id_to_exid(obj.0);
     let mut marks = MarkStateMachine::default();
     let ops_by_key = ops.group_by(|o| o.elemid_or_key());
     let mut len = 0;
-    let mut finished = Vec::new();
+    //let mut finished = Vec::new();
```
alexjg, line 144: Commented code

rust/automerge/src/automerge/current_state.rs
```diff
@@ -132,11 +174,11 @@ fn log_list_patches<'a, I: Iterator<Item = &'a Op>>(
...
                 })
         })
-        .for_each(|(index, (val_enum, (value, opid)))| {
+        .for_each(|(index, (val_enum, (value, opid, marks)))| {
             let conflict = val_enum > 0;
-            patch_log.insert(*obj, index, value.clone().into(), opid, conflict);
+            patch_log.insert(*obj, index, value.clone().into(), opid, conflict, marks);
         });
-    patch_log.mark(*obj, &finished);
+    //patch_log.mark(*obj, &finished);
```
alexjg, line 181: Commented code

rust/automerge/src/automerge/diff.rs
```diff
@@ -274,171 +273,48 @@ fn get_inc(before: &Winner<'_>, after: &Winner<'_>) -> Option<i64> {
...
-        }
-        f_vec.push(Mark {
-            start,
-            end,
-            data: Cow::Owned(MarkData {
-                name: from.data.name.clone(),
-                value: ScalarValue::Null,
-            }),
-        });
+    fn process(&mut self, op: &'a Op) -> Option<Patch<'static>> {
```
alexjg, line 304: I believe you can remove the comment above about this being O(n^2) as it is now O(n)?

rust/automerge/src/marks.rs
```diff
@@ -21,6 +23,129 @@ pub struct Mark<'a> {
...
+                len,
+                value: value.clone(),
+            })
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Default)]
+pub struct MarkSet {
+    marks: Arc<BTreeMap<SmolStr, ScalarValue>>,
```
alexjg, line 90: Why does this need to be `Arc` rather than `Rc`?

rust/automerge/src/marks.rs
```diff
@@ -21,6 +23,129 @@ pub struct Mark<'a> {
...
+impl<'a> Iterator for MarkSetIterator<'a> {
+    type Item = (&'a str, &'a ScalarValue);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iter.next().map(|(name, value)| (name.as_str(), value))
+    }
+}
+
+impl MarkSet {
+    pub fn iter(&self) -> MarkSetIterator<'_> {
```
alexjg, line 107: Can we just return `impl Iterator<..>` here and avoid the manually implemented `MarkSetIterator`

rust/automerge/src/patches/patch.rs
```diff
@@ -49,14 +52,15 @@ pub enum PatchAction {
         index: usize,
         /// The values that were inserted, in order that they appear. As with [`Self::PutMap`] and
         /// [`Self::PutSeq`] the object ID is only meaningful for `Value::Obj` values
-        values: SequenceTree<(Value<'static>, ObjId)>,
-        conflict: bool,
+        values: SequenceTree<(Value<'static>, ObjId, bool)>,
+        marks: Option<MarkSet>,
```
alexjg, line 56: Missing docs

rust/automerge/src/query/list_state.rs
```diff
@@ -1,6 +1,43 @@
...
+use fxhash::FxBuildHasher;
+use std::collections::HashMap;
+use std::ops::{Deref, DerefMut};
+
+#[derive(Debug, Clone, Default, PartialEq)]
+pub(crate) struct MarkMap<'a> {
+    map: HashMap<OpId, &'a MarkData, FxBuildHasher>,
+}
+
+impl<'a> Deref for MarkMap<'a> {
```
alexjg, line 14: Implementing `DeRef` for things which aren't smart pointers is generally a bad idea. Per [the docs](https://doc.rust-lang.org/std/ops/trait.Deref.html) on `Deref`:

> Deref should only be implemented for smart pointers to avoid confusion

Can we just wrap the relevant methodos from `HashMap`?

alexjg (commented):
Looks great! I've left a bunch of small comments and questions.

rust/automerge-wasm/PATCH.md
```diff
@@ -0,0 +1,189 @@
...
+  path: Prop[],
+  values: PatchValue[],
+  conflicts?: boolean[]
+}
+
+let patch : InsertPatch = {
+  action: "insert",
+  path: [ "emoji", 3 ],
+  values: [ "😍", "🔥", "🎆" ],
+  conflicts: [ false, true, false ] // optional
```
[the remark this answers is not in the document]
acurrieclark: Hi @orionz, I was wondering how marks work in this context. This presumably means that all of the values inserted have the marks applied?

If there were multiple inserts but only one had marks, would they appear as separate patches?

If there are marks on part of one of the strings, would the marks appear in a separate patch?

Can there be a conflict in a value, but different marks in each conflict?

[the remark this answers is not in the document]
alexjg: This implementation splits inserts and splices into ranges with the same marks. I.e. yes, if there are multiple inserts with different marks they will appear as separate patches.

Marks are independent of conflicts, they apply to ranges of elements in a sequence. If the elements themselves are conflicted that doesn't affect the marks.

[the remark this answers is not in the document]
acurrieclark: Understood, thank you @alexjg

Issue #632, opened 2023-06-08
Marks have `begin` and `end` - maybe `begin` and `length` would be better


orionz:
@jkankiewicz pointed out that `end < begin` is an un needed error case - if we we instead passed in an unsigned length we'd have one less error case to handle.  Is making an API change worth it? Thoughts?

heckj:
For what it's worth, length would be a smidge easier to compute than an end index when working with the Swift language overlay. Either is workable, but length seems more consistent with the rest of the API here.

alexjg:
Begin and length makes sense to me.

PR #635, opened 2023-06-08
Suggested changes for PR #610

jkankiewicz:
@ConradIrwin, these are my suggested changes to PR #610 in order to bring it in line with the existing C API.

They mostly involve:
* Disambiguating a default-initialized `AMmarkExpand` value from an `AM_MARK_EXPAND_NONE` value.
* Normalizing the documentation.

I was going to make it return an error when an `end` argument's value is less than its corresponding `start` argument's value but @orionz will probably address that issue by changing the Rust API to take `begin` and `length` arguments instead (Issue #632).


rust/automerge-c/src/doc/mark.rs
```diff
@@ -80,14 +108,14 @@ pub unsafe extern "C" fn AMmarkName(mark: *const AMmark) -> AMbyteSpan {
...
+    Default::default()
 }
 
 /// \memberof AMmark
-/// \brief Gets the value of the mark
+/// \brief Gets the value of a mark.
 ///
 /// \param[in] mark A pointer to an `AMmark` struct.
-/// \return A pointer to an `AMresult` struct with an `AMitem` struct.
+/// \return A pointer to an `AMresult` struct with an `AM_VAL_TYPE_MARK` item.
```
ConradIrwin, line 118: @jkankiewicz The value can be of any type (string/number/etc.) so I think the previous doc was better.

rust/automerge-c/src/item.rs
```diff
@@ -1241,26 +1241,26 @@ pub enum AMvalType {
     F64 = 1 << 8,
     /// A 64-bit signed integer value.
     Int = 1 << 9,
+    /// A mark
+    Mark = 1 << 10,
     /// A null value.
```
ConradIrwin, line 1246: @jkankiewicz It's good practice when adding to an enum to keep the old values the same. Although dependents shouldn't depend on the actual value, it makes life easier if they could.

rust/automerge-c/src/doc/mark.rs
jkankiewicz: > The value can be of any type (string/number/etc.) so I think the previous doc was better.

@ConradIrwin, you're right, it's a mistake to identify it as an `AM_VAL_TYPE_MARK` item.

rust/automerge-c/src/item.rs
jkankiewicz: > It's good practice when adding to an enum to keep the old values the same.

@ConradIrwin, it's good practice when there's a legacy unversioned serialization format because there's no way to ensure that a C enum's tag will be deserialized from its previously- corresponding integer value otherwise. In this case, however, a C enum's tag is translated into its corresponding Rust enum's variant prior to serialization so its corresponding integer value is irrelevant.

Furthermore, I try to keep a C enum's tags sorted in alphabetical order because, unlike with Rust's `match` expression, the C compiler will not report an error when a `case` for one or more tags in an enum is missing from a `switch` statement on a value of that enum. It's easier for one to notice that a tag has been neglected when they're in alphabetical order.
While I could retain the subsequent tags' previous bit assignments despite the insertion of a new tag in alphabetical order, it's helpful when debugging for the adjacency relationships between bits in a bitmask to mirror the adjacency relationships between their corresponding tags.

> Although dependents shouldn't depend on the actual value, it makes life easier if they could.

I agree that they shouldn't and I would add that dependents should be strongly encouraged to prefer named constants over magic numbers.

jkankiewicz:
> Let me know if you feel strongly about the enum stuff, and if so, I'm happy to defer to your re-numbering approach.

@ConradIrwin, I feel strongly that a C enum's tags should be declared in alphabetical order and that if two bits are adjacent within a bitmask that their corresponding tags should also be adjacent within their encompassing enum's declaration.

PR #639, opened 2023-06-11
Allow negative `del` for splice (Fixes #603)

### LINKED ISSUE #603: proposal: Cursor API for text editing

ConradIrwin:
This allows cursors to be used for common text editing operations
without having to recalculate indices.

Add some documentation and tests while I'm at it.

I think this (coupled with the new API using javascript-native string indices) fully addresses the idea in #603.

### DIFF
--- a/javascript/src/unstable.ts
+++ b/javascript/src/unstable.ts
@@ -279,11 +279,6 @@ export function splice<T>(
 
   index = cursorToIndex(state, value, index)
 
-  if (del < 0) {
-    index += del
-    del = 0 - del
-  }
-
   try {
     return state.handle.splice(value, index, del, newText)
   } catch (e) {
--- a/javascript/src/unstable.ts
+++ b/javascript/src/unstable.ts
@@ -242,6 +242,21 @@ function cursorToIndex<T>(
   }
 }
 
+/**
+ * Modify a string
+ *
+ * @typeParam T - The type of the value contained in the document
+ * @param doc - The document to modify
+ * @param path - The path to the string to modify
+ * @param index - The position (as a {@link Cursor} or index) to edit.
+ *   If a cursor is used then the edit happens such that the cursor will
+ *   now point to the end of the newText, so you can continue to reuse
+ *   the same cursor for multiple calls to splice.
+ * @param del - The number of code units to delete, a positive number
+ *   deletes N characters after the cursor, a negative number deletes
+ *   N characters before the cursor.
+ * @param newText - The string to insert (if any).
+ */
 export function splice<T>(
   doc: Doc<T>,
   path: stable.Prop[],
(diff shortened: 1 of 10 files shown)


orionz:
I like this behavior.  Feels like it should be implemented in `rust/automerge/src/transaciton/inner.ts` instead of in the javascript library.  @alexjg - agree?

orionz:
If Alex gives a thumbs up to this change in the rust are you ok making it (it would the same 3 lines and changing all the `del`'s from unsigned to signed.   If not I'm happy to do it.

alexjg:
Yeah I agree, this seems like a sensible change to make in the `Transactable::splice_text` API

ConradIrwin:
Happy to make the change to rust @alexjg. Based on your comment I'll only make the change for `splice_text`, but I'm also happy to extend it to `splice`. (On the one hand it doesn't seem as useful for lists, on the other it's annoying to have two very similar APIs that work in slightly different ways).

alexjg:
I don't see any harm in extending this to `splice` and it would be nice to have a consistent API

PR #640, opened 2023-06-12
from(doc) makes a shallow copy

orionz:
Realized that `Automerge.from(doc)` could not be used to make a shallow copy.

Made a change to the proxy to allow this.  Also noticed that `readme` and `frozen` attrs were still in the proxy and are no longer needed since we moved to patches and incremental updates so I removed them.


javascript/src/stable.ts
```diff
@@ -78,6 +78,8 @@ export type ChangeOptions<T> = {
   time?: number
   /** A callback which will be called to notify the caller of any changes to the document */
   patchCallback?: PatchCallback<T>
+  /** copy doc references **/
+  allowDocRef?: boolean
```
alexjg, line 82: This should basically never be used externally right? Is there a way to hide it from the API? Maybe using the `@internal` annotation?

orionz: I dont see any reason it needs to be internal only.  If someone wants to copy a chunk of a document from one place to another or from one document to another there is now a way.  It's not the default because it's probably not the behavior people are expecting most of the time but I'm happy its possible now.

PR #641, opened 2023-06-12
Use UTF-8 indices for automerge-c

ConradIrwin:
Currently the C API expects strings to be encoded in UTF-8 (and provides
them that way), but expects programmers to provide indices in UTF-32.

This code changes that so that (by default) indexes are assumed to be
UTF-8, but you can build a version that uses UTF-32 indexing with a
compile time flag.


ConradIrwin:
@alexjg it might also be worth you looking at the rust side of this. In particular I'm not sure the implications of adding a cargo "feature" to toggle this, or whether there's a better way.

alexjg:
I don't love the feature flag. The problem is that feature flags in cargo are additive, which means that if any dependency in a build activates the `utf8-indexing` feature then that feature is active for the every crate in the build. This is a pretty short road to dependency hell.

However, fixing this is a bit more work. We need to reintroduce a generic parameter for the text indexes. I want to do this but I think it requires a bit more deliberate design to make the complexity manageable and worth it (I think this can be bundled up with more sophisticated user provided indexes on each optree for example). It seems to me that the ability to use utf-8 indexes is going to be pretty important for FFI usecases and so we should keep the feature flag in, but maybe we can add some documentation around how you definitely shouldn't use it if anyone is going to be depending on your code?

jkankiewicz:
@ConradIrwin, I've opened PR #643 for my suggested changes to PR #641.

jkankiewicz (approved):

PR #642, opened 2023-06-12
Unicode aware offsets for mark/unmark

ConradIrwin:
This makes the indices passed to mark() match those passed to splice().


alexjg:
LGTM, just needs a little formatting. @orionz all look good to you?

PR #643, opened 2023-06-15
Suggested changes for PR #641

jkankiewicz:
@ConradIrwin, this is the sum of the changes I'm suggesting for PR #641; most of them are to the documentation.

I did not attempt to fix the failure in the `test_AMmark_unicode_indexing()` unit test case.


ConradIrwin:
All makes sense to me, thanks for the cleanup! (Also nice trick with doxygen to get both to show in the docs).

ConradIrwin:
The AMmark_unicode_indexing test is fixed by #642

PR #645, opened 2023-06-17
Scaffolding for generic move operations

LiangrunDa:
A generic move operation is capable of moving any element, including objects and scalar value, from one location to another location. In this context, a "location" refers to either a list object or a map object.

To implement a generic move operation, one or two fields need to be added to an operation, which will cause backward incompatibility. But the current commit focuses on creating the necessary scaffolding code without affecting the compatibility of the current version. The `operation` struct will continue to be modified in the future updates. 

From the high-level point of view, the generic move operation will maintain a struct named MoveManager, which is responsible for maintaining global information on all the move operations. For example, when two move operations move the same object, the MoveManager will decide which one is the winner based on the Lamport clock. Besides, MoveManager will check whether a move operation introduces a loop, and if so, the move operation will not take effect. The MoveManager is a memory structure that is reconstructed alongside the `Document` and is updated with each newly generated operation.

In addition, a new OpType needs to be introduced: `move`. A move operation will be inserted at the destination location each time a move is made. Subsequently, to make the removed element in the source location invisible, the move operation is added to the successor of the operation of the source location.


orionz:
Hello there.  Thank you for this PR.  I am excited to see move ops finally make their way into Automerge!

A few thoughts on how this can be integrated...   I suspect the MapManager object should live on `OpSetMetadata` struct.  This is a place we store actorid and prop information to allow us to do lamport compares without having to have a copy of the actor on every opid in the system.   The metadata should be onhand any time we need it. 

The critical thing for performance and easy of use is that the move op's need to be able to correctly resolve the `Op::visible_at()` method and the `Op::value_at()` method.  I suspect the visible_at() and value_at() methods will need to take `OpSetMetadata` as an argument because the methods need to be able to return the correct state of the op at any point in the op history.  Has the move happened yet?   Has it been conflicted?  Will it be conflicted later or moved again later but hasn't yet?  This also means the MoveOp will need to have on it enough information on it to resolve a value locally - a copy of the scalar value it moved or the opid and objtype of the object moved.  I think this means we should NOT support moving counters because we move the counter and not the increments.  (let me know if I'm wrong).

There is also a `Op::visible()` and `Op::value()` method that does NOT take a clock and I would imagine also not take the metadata as an argument.  For these to work I would imagine that you'd need to store on the op itself weather or not it's been moved, or voided by a cycle etc.  This will make the `insert_op()` for moves a little more complicated b/c you may need to insert the move op and then invalidate another move op in another object all in one go.   This will be critical b/c the length indexes need to be updated to take into account an invalidated move within a list they track.

We'll also need to make sure the op serialization / deserialization works with the binary format.

I see you have a data-structure that manages the shape of the tree.  That *should* be redundant as we already have the `OpSet::parent_object()` method that you can use on any object to walk backward to the root.

There's going to need to be a `move` operation in PatchLog, PatchBuilder and Patch.  

Once that's done all the plumbing to make patches and hydrate documents (should 🤞 work).

PR #647, opened 2023-06-19
JS: don't materialize on view

orionz:
There was an unnecessary call to `view()` that would call an expensive `materialize()` that was not needed.  Removing for performance reasons.

### DIFF
--- a/javascript/src/types.ts
+++ b/javascript/src/types.ts
@@ -37,6 +37,7 @@ export type Doc<T> = { readonly [P in keyof T]: T[P] }
 export type PatchInfo<T> = {
   before: Doc<T>
   after: Doc<T>
+  source: string
 }
 
 /**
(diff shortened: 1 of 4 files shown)


orionz:
Added source field to `PatchInfo` so we can tell local from sync changes.  Noticed that `saveIncremental()` wasn't exported on unstable.

javascript/src/types.ts
```diff
@@ -37,6 +37,7 @@ export type Doc<T> = { readonly [P in keyof T]: T[P] }
 export type PatchInfo<T> = {
   before: Doc<T>
   after: Doc<T>
+  source: string
```
acurrieclark, line 40: Is it worth adding a `Source` type here which is a union of the possible sources? Something like
```ts
export type PatchSource = "from" | "change" | "changeAt" | "merge" | "change" | "loadIncremental" | "change" | "sync";
```
Could add it as the type on `progressDocument` too?

PR #672, opened 2023-07-10
js: Return heads of new changes from changeAt

alexjg:
`changeAt` is useful because it allows the user to choose when they synchronize state with the document by saving the heads of the last sync point and then applying changes since that last sync point using `changeAt`. This only goes one way however, to go in the other direction the user needs to be able to obtain the diff from the now applied changes to the state of the document including any other concurrent changes. To enable this modify `changeAt` to return a `{newDoc: Doc<T>, newHeads: Heads | null}` instead of just `Doc<T>`. The returned `newHeads` is the hash of the transaction created by the change callback and can be used as the `before` argument to `diff` to obtain a diff from the result of the changeAt call to the state of the document including concurrent changes.

### DIFF
--- a/javascript/src/stable.ts
+++ b/javascript/src/stable.ts
@@ -296,6 +296,7 @@ export function from<T extends Record<string, unknown>>(
   _opts?: ActorId | InitOptions<T>,
 ): Doc<T> {
   return _change(init(_opts), "from", {}, d => Object.assign(d, initialState))
+    .newDoc
 }
 
 /**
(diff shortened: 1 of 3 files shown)


orionz:
Trying to think of an alternate way to do this.   What if instead of changing the external API we add `lastChange` to `InternalState` and add a top level function `Automerge.getLastLocalChangeHash(doc)`.

Seems like generally useful information that we could easily track and make available in all situations instead of just after a changeAt().  It'd keep the api calls more consistent too.

alexjg:
Hmm. That API seems a little messy to me. The biggest problem is that the change function in `changeAt` might not do anything, and so `getLastLocalChangeHash` would then refer to whatever change was made before `changeAt`. This in turn means that users would have to do some fiddly thing like save `getLastLocalChange`, make their change, and then check if `getLastLocalChange` has a new value. Is there a reason to avoid returning the hash from `changeAt`?

PR #678, opened 2023-07-19
text/v1: return value in slice()

stephen:
I noticed that `slice()` on text doesn't actually return the value, so the return is always `undefined`.

This PR adds a fix and a test case.

### DIFF
--- a/javascript/src/text.ts
+++ b/javascript/src/text.ts
@@ -215,7 +215,7 @@ export class Text {
   }
 
   slice(start?: number, end?: number) {
-    new Text(this.elems.slice(start, end))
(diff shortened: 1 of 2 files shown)


alexjg:
LGTM. If you can fix the lint error I'll merge

stephen:
I might be reading wrong, but I see the lint error on the main branch as well: https://github.com/automerge/automerge/actions/runs/5553266478/jobs/10141602059

alexjg:
Ah right you are. I've pushed a fix to `main` so if you rebase on `main` this should be good to go.

stephen:
rebased, thanks!

PR #684, opened 2023-07-26
should be throwing range errors not strings

### LINKED ISSUE #659: Automerge errors in js from automerge-wasm don't have stack traces
While investigating #658, I found it very tricky to pinpoint where this error was being thrown from because it seems like the errors are raw strings with no stacktraces attached:
```ts
import * as Automerge from "@automerge/automerge";
try {
  Automerge.from({ d: new Automerge.Text("👀👀") });
} catch (err) {
  console.log(typeof err);
  console.log(err);
}
```

outputs:
```
string
index 1 is out of bounds
```

It would be great if errors across the wasm boundary captured a stacktrace to make hunting down errors more possible. Without it, I ended up sticking `try {} catch {}` everywhere to pinpoint the problem.

orionz:
We were throwing JsString's not RangeErrors from wasm. Fixes #659


stephen:
aha, thanks for looking at this! could you add a test similar to [this](https://github.com/automerge/automerge/pull/666/files#diff-ae42699ce73688e24115b44257b95383829fc413d462630eaccd366e9c1a385c)?

PR #692, opened 2023-08-03
change a reference to a pointer to handle null pointers properly

orionz:
Found a possible null pointer bug in automerge-c `AMitemsToMark()`


jkankiewicz:
> Found a possible null pointer bug in automerge-c AMitemsToMark()

To clarify, the out argument for `AMitemToMark()` , unlike all of the other `AMitemTo...()` functions that return a pointer value, was declared as `&mut *const` instead of `*mut *const` which caused the function to unintentionally return `false` when the out variable was `NULL` instead of when the out argument was `NULL`.

@orionz ran afoul of this bug because the C compiler he's using is atypically zeroing uninitialized pointer variables and that caused `AMitemToMark()` to ignore the out variable and return `false` instead of reassigning the out variable and returning `true`.

Worse yet, if `AMitemToMark()` were to be called with a `NULL` out argument value then it would crash the program on an access violation instead of returning `false`.

Issue #708, opened 2023-08-24
diff not returning complete patch sets with `next` string arrays


acurrieclark:
This might be a symptom of something grander, but this specific example doesn't produce the expected set of patches:

```ts
import { next } from "@automerge/automerge"

const before = next.from<DataType>({
  values: ["oh", "you", "pretty", "things"],
});

const after = next.change(before, (d) => {
  next.deleteAt(d.values, 2);
});

const patches = next.diff(after, next.getHeads(after), next.getHeads(before));

console.log(patches); // => [ { action: 'insert', path: [ 'values', 2 ], values: [ '' ] } ]
```

Here, I would expect patches to be one of the following:

```ts
// this is also the patch which is correctly produced if not using `next`
[ { action: 'insert', path: [ 'values', 2 ], values: [ 'pretty' ] } ]
```

or

```ts
[ 
  { action: 'insert', path: [ 'values', 2 ], values: [ '' ] } ,
  { action: 'splice', path: [ 'values', 2, 0 ], value: 'pretty' } ]
]
```
Either way, the value of the string is not actually included.

PR #714, opened 2023-08-31
Splice full string when diffing the reverse of a deletion

acurrieclark:
See #708


alexjg:
This is fixed on `main`.

PR #717, opened 2023-09-01
Test and fix for finding a string in a `next` list

acurrieclark:
When looking for a string in a list of `next` strings using indexOf, the current logic only matches if we are looking for the exact `String` object, not the string value.

This adds a check for the string value itself too along with a rudimentary test.


Stvad:
I've been hitting this issue (indexOf doesn't find string matches correctly in the automerge array) & wondering if there are any blockers for merging this?

PR #725, opened 2023-09-07
automerge-c: External build support

jkankiewicz:
@orionz and @alexjg, these changes enable an external build tool—like SCons—to delegate the building of the automerge-c library to CMake instead of requiring its CMake build configuration to be translated into the external build tool's native configuration.


jkankiewicz:
This PR is superseded by #780.

PR #755, opened 2023-10-04
Rename crate::exid::ExId to crate::obj_id::ObjId

teohhanhui:


teohhanhui:
> Now, the interesting question is whether this has any compatibility effects. It seems like it should definitely not, because the exported name and type have not changed, but I am cautious.

If we used type alias it'd have compatibility concerns: https://predr.ag/blog/re-exporting-enum-with-type-alias-breaking-change-not-major/

But we didn't. There are no compatibility concerns that I'm aware of.

> @teohhanhui did you encounter anywhere when you wrote code against this library that had to use `ExId` rather than `ObjId`?

No.

alexjg:
> This change doesn't fix everything. Wherever ExId is still present in impl block header or function / method signature...

Can you elaborate?

teohhanhui:
e.g.

https://github.com/automerge/automerge/blob/9400e195f1d01b0c10d535c1c1ccd321e684c362/rust/automerge/src/autocommit.rs#L574

results in

Screenshot 2023-10-04 at 23-51-40 AutoCommit in automerge - Rust

teohhanhui:
Looks like that will be fixed soon: https://github.com/rust-lang/rust/pull/113374

alexjg:
Hmm, do we need to make this change then?

teohhanhui:
In the linked PR, I don't see any tests for `impl` block header.

teohhanhui:
Looks like that's https://github.com/rust-lang/rust/issues/41072

alexjg:
I'm inclined to think that this is a rustdoc bug and we should just wait for it to be fixed as even with these changes we end up with `ExId` leaking out, wdyt?

PR #761, opened 2023-10-11
Export list of change sources as array as well as type

acurrieclark:
PatchSources are currently defined as a type, but it is also useful to be able to check that a source is valid at runtime. This PR keeps the type the same, but also exports the list of sources as an array.

### DIFF
--- a/javascript/src/next.ts
+++ b/javascript/src/next.ts
@@ -49,6 +49,7 @@ export {
   type AutomergeValue,
   type ScalarValue,
   type PatchSource,
+  PatchSources,
   type PatchInfo,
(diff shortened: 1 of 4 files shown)


acurrieclark:
@alexjg I added this pre-emptively as I thought it might be useful for the work I am doing on repo right now. However, as per our earlier conversation, it might actually be a little premature.

Happy to close if you think it unnecessary.

NB. Not entirely sure why it has failed the lint check; it seems to be a rust file which has failed?

alexjg:
Closing for now as it doesn't seem needed so far.

PR #762, opened 2023-10-13
Cache the last patch on the document's InternalState

pvh:
I don't love this! But here it is, a band-aid for a performance problem.

## How it works

What this does is every time you call progressDocument, that patch is also cached on the new document. (LRU=1) When you make a new call to progressDocument, the patch is replaced. 

To benefit from this, we check in diff if there's a saved patch with exactly those heads and if so, we return them!

This allows you to call diff() with the same arguments a few times in a row and memoize the result.

## What's missing / not great

Well, first I'm not a huge fan of shoving stuff into the InternalState object but... it seems relatively harmless.

The tests all pass but it could probably use a couple of extra tests to stress it. I did test that if I messed up the various inputs that it would fail in expected ways but there's not a test that targets this *specifically*.

I did some analysis of the code to make sure this was the correct intervention and @orionz, there were three calls which take patchCallback but *don't* go through progressDocument. Those are:

 * init & load (which use handle.materialize) and 
 * clone (which uses handle.applyPatches directly).

I think it would be sensible to consolidate those onto progressDocument for consistency but I suspect the approach used in load is cheap and fast and that's why it's preferred. Apologies for not having enough knowledge to make an informed call there. There's no downside to not doing this, particularly, other than initial loads in automerge-repo will not currently register themselves in the cache.

The other change @orionz might hate is that we now always use the applyAndReturnPatches form of apply_patches_impl.

## What we should probably do longer term

I think this whole API could benefit from some adjustment -- Orion and I have talked about exposing the diff_cursors as a primitive users could consume but we could probably do even better than that if we did a bit of thinking. (I find the diff_cursors kind of hard to reason about even as I've been working on this code.

## Who wants this?

@geoffreylitt, rather urgently. It's the difference been usable (40ms typing latency) and unusable (110ms typing latency).

Note that even if we accept this there's still more tuning to do. See the attached image.

image

### DIFF
--- a/javascript/src/internal_state.ts
+++ b/javascript/src/internal_state.ts
@@ -8,6 +8,7 @@ export interface InternalState<T> {
   handle: Automerge
   heads: Heads | undefined
   freeze: boolean
+  mostRecentPatch: any // TODO: type this
   patchCallback?: PatchCallback<T>
   textV2: boolean
 }
--- a/javascript/src/stable.ts
+++ b/javascript/src/stable.ts
@@ -475,19 +475,27 @@ function progressDocument<T>(
   }
   const state = _state(doc)
   const nextState = { ...state, heads: undefined }
-  let nextDoc
-  if (callback != null) {
-    const { value, patches } = state.handle.applyAndReturnPatches(
-      doc,
-      nextState,
-    )
-    if (patches.length > 0) {
-      callback(patches, { before: doc, after: value, source })
+
+  const { value: nextDoc, patches } = state.handle.applyAndReturnPatches(
+    doc,
+    nextState,
+  )
+
+  if (patches.length > 0) {
+    if (callback != null) {
+      callback(patches, { before: doc, after: nextDoc, source })
     }
-    nextDoc = value
-  } else {
-    nextDoc = state.handle.applyPatches(doc, nextState)
+
+    const newState = _state(nextDoc)
+
+    newState.mostRecentPatch = {
+      before: _state(doc).heads,
+      after: newState.handle.getHeads(),
+      patches,
+      source,
+    }
   }
+
   state.heads = heads
   return nextDoc
 }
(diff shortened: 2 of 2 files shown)


pvh:
Here's a copy of a performance trace from a couple of keystrokes in the current version of Tiny-Essay-Editor while it's running a very similar but not identical patch. Note that we still have a lot of other work to do to get down to consistently <10ms (particularly because CodeMirror & React are eating a lot of our time.)

https://cdn.discordapp.com/attachments/721451876171644989/1162186613057659031/with-cached-diffs?ex=653b05a0&is=652890a0&hm=1e5c5a3541545b4e26609419660b886265d561153903d4782d1685b93a1433f2&

acurrieclark:
@pvh following on from our conversation I have been thinking a lot about this. One thought I dwelt on for a while was that this route could open up the possibility of a new function to get the last set of patches from the doc. For example:

```ts
const doc = init();

const newDoc = change(doc => {
  doc.foo = "bar";
});

// new function
const {patches, source} = getLatestPatches(newDoc);

// patches => Patch[], source => PatchSource
```

At that point you would have all the information returned by the `patchCallback`; `doc` and `newDoc` are your before and after, `source` is the source in `PatchInfo` and you have your patches.

Crucially for `repo` this could be called on the document at any stage in the proceedings, so we can avoid my most recent PR almost entirely.

pvh:
> Crucially for `repo` this could be called on the document at any stage in the proceedings, so we can avoid my most recent PR almost entirely.

This was definitely the motivation but "getLatestPatches" feels like a fragile design -- in a complex system where other things could change your patches it could break at any time. That said, I DID stash the `source` on there with the thought that we could perhaps expose it somehow. That's for a followup though.

acurrieclark:
I saw that you had included `source` too, which is really what clarified my thought on it. Agree it's for a later update, but I think the principle should be sound.

I have always thought of a document as a snapshot of the state of the file at that time. If the patches remained the same beyond the next change, then I would hope it would be pretty solid?

```ts
const doc = init();

const newDoc = change(doc, (d) => {
  d.foo = "bar"
});

const finalDoc = change(newDoc, (d) => {
  d.foo = "baz"
});

console.log(getPatches(newDoc)); //. => foo was set to bar
console.log(getPatches(finalDoc)); // => foo was set to baz
```

alexjg:
This LGTM, I'll fix the lint error shortly and merge.

pvh:
Thanks! If you could ship a release I'll get it into TEE for Geoffrey.

pvh:
@alexjg with Orion's patch, we should revert this, right?

PR #768, opened 2023-10-14
Cache clocks for faster clock_at()

orionz:
`clock_at()` is scaling at O(n) and is becoming a problem with apps making a change per keystroke 

Changed Clock to use `im::HashMap` to make clones quick and memory usage shared between clocks.

Added a cache to ChangeGraph to cache evey 32nd clock.  When generating a clock walk backwards until hitting a cached clock and stop.

This should make clock_at() to be O(1)

### DIFF
--- a/rust/automerge/src/change_graph.rs
+++ b/rust/automerge/src/change_graph.rs
@@ -16,7 +16,10 @@ pub(crate) struct ChangeGraph {
     edges: Vec<Edge>,
     hashes: Vec<ChangeHash>,
     nodes_by_hash: BTreeMap<ChangeHash, NodeIdx>,
+    clock_cache: Vec<Clock>,
 }
+
+const CACHE_STEP: u32 = 32;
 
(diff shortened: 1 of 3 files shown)


rust/automerge/src/change_graph.rs
```diff
@@ -16,8 +16,11 @@ pub(crate) struct ChangeGraph {
     edges: Vec<Edge>,
     hashes: Vec<ChangeHash>,
     nodes_by_hash: BTreeMap<ChangeHash, NodeIdx>,
+    cache: Vec<Clock>,
```
pvh, line 19: clock_at_cache?

alexjg:
The code here LGTM but it introduces a 14% slowdown in the edit trace benchmark ofr me. We are over-focused on that benchmark I think but I think we should justify the slowdown with appeal to some other thing (e.g. speeding up a slow generate_sync_message)

pvh:
LMK when this is released, I think it will make a big difference in text editing performance in TEE.

PR #774, opened 2023-10-18
automerge-c: Merge automerge_core into automerge

jkankiewicz:
Merging the "automerge_core" library built by Cargo into the "automerge" library built by CMake simplifies installation of automerge-c because of its reduction to only one binary file.

It also simplifies linking when the automerge-c CMake project is built as a subdirectory of a dependent CMake project.


jkankiewicz:
This PR is superseded by #780.

Issue #779, opened 2023-10-24
Running Automerge (and Automerge Repo) in a Service Worker


romaricpascal:
Hello,

As part of an investigation for a side project, I've created [examples of Automerge and Automerge Repo running inside Service Workers](https://gitlab.com/romaricpascal/automerge-in-service-worker#automerge-in-service-workers) (bundled by Rollup, which [seems to have been an issue in the past](https://github.com/automerge/automerge/issues/367).

I was wondering if that's something you'd be interested to have as part of the examples in both repositories. I appreciate having them in your repositories does increase the maintenance burden of the project in the long run, though, so no hard feelings at all if you prefer keeping that out for now and close this issue.

Thanks for all the research and work on Automerge 😊

alexjg:
Oh this is fantastic stuff and I would love a PR to add them as examples.

Are there things we can change about the way automerge and automerge-repo are distributed which would make it easier to achieve this?

romaricpascal:
Great! Glad you're keen to have these examples, I'll create pull requests to add them to each repository 😊 

Regarding changes to how automerge and automerge-repo are distributed, it's tough to say which changes may help for all bundlers, as I only focused on making it work with Rollup. For that I had [to crudely update the `initSync()` call in `@automerge/automerge-wasm/web/index.js`](https://gitlab.com/romaricpascal/automerge-in-service-worker/-/blob/main/rollup.config.js?ref_type=heads#L30) so [it worked with `@rollup/plugin-wasm`](https://www.npmjs.com/package/@rollup/plugin-wasm#using-with-wasm-bindgen-and-wasm-pack). But according to other issues, it looks like [Parcel needs something slightly different](https://github.com/automerge/automerge/issues/525#issuecomment-1465871813) and it sounds like [Webpack is kind of happy with wasm-bindgen output](https://github.com/automerge/automerge/issues/367). No clue what would happen in Vite as well, which is growing in popularity.

From the point of view of running in a Service Worker, the code will need to get bundled in any case as [Firefox doesn't support modules in service workers](https://web.dev/articles/es-modules-in-sw#backwards_compatibility), at least into something that can be called via `importScripts`, if not bundled with the main script of the service worker. 

If `automerge` and `automerge-repo` were providing a pre-bundled version as a UMD module (or IIFE), that might make things simpler to run in Service Workers. Two things to solve come to mind in that situation though:
[...]

PR #782, opened 2023-10-25
Add commented example of Rollup bundling

romaricpascal:
Illustrates both running on the pages, as well as in a Service Worker, with commented config (and JS files for the service worker as there's a bit of coordination between the page and the worker that needs explaining).

Part of #779.

### DIFF
--- a/javascript/examples/rollup/public/in-service-worker/service-worker.js
+++ b/javascript/examples/rollup/public/in-service-worker/service-worker.js
@@ -1,0 +1,87 @@
+// Firefox is not heavy on forwarding logs from Service Workers
+// to the main window, so instead, we'll use a BroadcastChannel
+// to help debugging. A bit blunt, but that'll do
+const channel = new BroadcastChannel("service-worker-log")
+const originalLog = console.log
+console.log = function (...args) {
+  try {
+    channel.postMessage(args)
+  } catch (error) {
+    // There'll be one point where I'll try to log something
+    // that's not serialisable for `postMessage` so that'll
+    // avoid crashing.
+    originalLog(...args)
+  }
+}
+
+// Import the bundled script for automerge, which will add an `AutomergePromise`
+// allowing to wait for the automerge API once all the event listeners
+// of the service worker have been set up synchronously
+self.importScripts("/in-service-worker/automerge.js")
+
+// Little trickery to have a way to resolve the promise only late
+// With that pattern, we get a reference we can pass to `waitUntil`
+// or use as a way to delay responses to `isReady`, while delaying
+// when the Promise is resolved till the Automerge API has been loaded
+// Thankfully the function passed to the Promise constructor
+// runs synchronously, allowing to register the listeners OK
+const ready = new Promise(markReady => {
+  self.addEventListener("install", event => {
+    console.log("Installing service worker", new Date().toISOString())
+    self.skipWaiting()
+
+    event.waitUntil(ready)
+  })
+
+  self.addEventListener("activate", () => {
+    console.log("Service worker activated")
+  })
+
+  self.addEventListener("message", event => {
+    console.log("Message from page", event)
+    const handler = MESSAGE_HANDLERS[event.data.action]
+    if (handler) {
+      handler(event)
+    }
+  })
+
+  const MESSAGE_HANDLERS = {
+    async isReady(event) {
+      console.log("Checking if server is ready", ready)
+      await ready
+      event.source.postMessage({ action: "ready" })
+    },
+  }
+
+  // This need to happen only after all event listeners are set up
+  // This needs to run in the main script as there's no controlling
+  // when the service worker script will be torn down and started again
+  // Without this, when we close the browser, `automerge` and `doc`
+  // would be undefined, which is less than ideal :(
+  ;(async function () {
+    let automerge = await AutomergeAPI
+    let doc = automerge.from({
+      value: 1,
+    })
+
+    // You'd likely want to add some persistence here, to IndexedDB most likely
+    // as it's the storage available inside service workers
+
+    Object.assign(MESSAGE_HANDLERS, {
+      getValue(event) {
+        event.source.postMessage({ action: "setValue", value: doc.value })
+      },
+      increment(event) {
+        const { value } = event.data
+        doc = automerge.change(doc, draft => {
+          draft.value += value
+        })
+        console.log("Sending back", doc.value)
+        event.source.postMessage({ action: "setValue", value: doc.value })
+      },
+    })
+
+    console.log("Marking server as ready")
+    markReady()
+  })()
+})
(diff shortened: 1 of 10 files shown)


javascript/examples/rollup/rollup.config.js
```diff
@@ -0,0 +1,100 @@
...
+import { defineConfig } from "rollup"
+import tla from "rollup-plugin-tla"
+
+const inPageConfig = {
+  input: "src/index.js",
+  output: {
+    file: "public/main.js",
+  },
+  // Fixes the `this` instance at top level in automerge's `stable.js` file
+  context: "globalThis",
```
romaricpascal, line 14: This _might_ be something that could be fixed in `automerge`'s published package

javascript/examples/rollup/rollup.config.js
```diff
@@ -0,0 +1,100 @@
...
+  plugins: [
+    // Force automerge-wasm to resolve to its web version
+    alias({
+      entries: [
+        {
+          find: "@automerge/automerge-wasm",
+          replacement: "node_modules/@automerge/automerge-wasm/web/index.js",
+        },
+      ],
+    }),
```
romaricpascal, line 24: Rollup (well, its node-resolve plugin), would try to load the `bundler/automerge_wasm.js` of `@automerge/automerge-wasm`. I tried amending its `__wbg_set_wasm(wasm);` to `__wbg_setwasm(await wasm.default())` (similarly to the replacement of the `initSync` call for `web/index.js`). This is a step towards complying with `@rollup/plugin-wasm` way of handling the wasm module, but it seems there's another part of the code that needs an `await` but I'm not sure which: when running in the browser, an Error gets thrown that `__wbindgen_add_to_stack_pointer` is not a function.

Thanks to it synchronous `initSync` call the `web/index.js` file make things a bit smoother with Rollup it seemes.

alexjg:
This is great stuff. I'm not an expert in service workers so I'm pinging @pvh to take a look but from my side it LGTM.

javascript/examples/rollup/public/in-service-worker/service-worker.js
```diff
@@ -0,0 +1,87 @@
...
+    if (handler) {
+      handler(event)
+    }
+  })
+
+  const MESSAGE_HANDLERS = {
+    async isReady(event) {
+      console.log("Checking if server is ready", ready)
+      await ready
+      event.source.postMessage({ action: "ready" })
```
romaricpascal, line 52: Something that I came across since I raised the PR. Not sure if you'd prefer it as an opt-in rather than a default, let me know 😊 

```suggestion
      // Make the service worker start handling fetch events from the page straight away
      // which can be necessary when users reload `Ctrl/Command-Shift-R`, 
      // which bypasses service worker
      // See: https://web.dev/articles/service-worker-lifecycle#shift-reload
      await self.clients.claim()
      event.source.postMessage({ action: "ready" })
```

pvh: No, if you need this you need it. This is good but @romaricpascal I don't see you handling controllerchange here.

romaricpascal: 🤔 I'll need to have some further thoughts on whether it's a need in all situations. Same would go for the `self.skipWaiting()` on line 31. 

I'm wondering if these two statements are not app dependent, unlike the rest of the code which is really tied to automerge needing to load asynchronously and coordinating the page with that given `navigator.serviceWorker.ready` wouldn't provide the info that automerge has actually loaded.

Depending on your app, you may or may not want to activate a new worker at a custom time rather than straight after install (eg. coodinating with the UI via a `BroadcastChannel` rather than automatically), and may or may not want to claim clients straight away (though this seems more necessary for shift-reloads and first run), which in turn may or may not lead you to want to use `controllerchange` (which only matters if the JS on the page and the one in the worker are different versions of the app). This seems to fall a bit in the same area as the `update()` method on registration and `updatefound` events, where apps will have different use depending on the feature they provide.

romaricpascal: @pvh I've had a bit more thinking on what's the minimum necessary and that's where I got to:
1. I'd say the `clients.claim()` is necessary. In the context of running Automerge in a Service Worker, you'll likely want to be sure that your page is controlled by the service worker, which means you want the worker to take control on first run (or Shift-reloads).
2. The `self.skipWaiting` however, is a mater of how you want a new version of the worker to take over, so I'd be keen to leave it out of the example as it'll depend on how the app wants to handle its service worker updates (which will vary from one to another).
3. Similarly `controllerchange` only occurs when a new controller is replacing an existing one if I understood correctly, so again a matter of how the app wants to handle its service worker updates.
4. And `update` and its related `updatefound` event also depend on how the app plans to handle update.

In the end, I'm thinking of:
- removing the `self.skipWaiting` bit from the example
- adding a note in the service worker section of the readme explaining that:
    - this example only deals with bundling automerge for use in a service worker and the coordination with loading the WASM
    - how the service worker updates will be dependent from one application to another, and you can find a trove of information on the different mechanisms at your disposal in [The service worker lifecycle article on web.dev](https://web.dev/articles/service-worker-lifecycle).

Does that sounds a good way forward?

pvh:
One very important case is when the service worker restarts -- this triggers neither "controllerChange" nor "activate". I haven't quite nailed this yet but it causes real failures in my application. Here's my implementation (which currently does not work in FF because it uses `import`): https://github.com/pvh/trail-runner/blob/main/service-worker.js

romaricpascal:
I got caught by surprise as well by the service worker being active but having restarted and lost whatever it had in memory from when it was previously running. I've only been hit by this when closing/re-opening the whole browser (which made sense to me, as I'd hoped the worker wouldn't stay alive when the browser is off, at least when not configured to receive notifications). In which scenario have encountered it?

That's another reason why the page doesn't just use [the `ready` Promise from the ServiceWorkerContainer](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/ready) but [a `postMessage` handshake to know when the worker has actually loaded](https://github.com/automerge/automerge/pull/782/files#diff-acf465e537c7d41d839ba235d88767a3bbfd8dc35edd67d2042bff4e4d06d6baR15). With only the `ready` promise, the page would think things are OK, while [some things still need initialising worker side](https://github.com/automerge/automerge/pull/782/files#diff-b3ee05e2cda575f6c661c5227a4bf45c9f3fcb06367dad91e1ba7df41fd96568R56-R60). 

With this and the pages getting claimed during that handshake (because if claiming only when activated, Shift-reloaded pages would not become controlled), the page can only start whatever interacts with the worker only when it's controlled by a worker and the worker is ready.

alexjg:
I'm going to close this as I think service worker examples belong in the [automerge-repo](https://github.com/automerge/automerge-repo) codebase rather than in this repository.

PR #793, opened 2023-11-07
Run cron advisories job on `rust` folder

MarijnS95:
Ever since commit dd3c6d130 ("Move rust workspace into ./rust") this job seems to have been failing on not finding a `Cargo.toml` in the root.

This was fixed for the "main" CI on pushes and pull requests in commit e8309495c ("Update `cargo deny` to point at `rust` subdirectory"), but not for the scheduled cron job leading to unnecessary failures in (fork) repos.


MarijnS95:
In hindsight I could have probably moved this job from `ci.yaml` and deduplicate it into `advisory-cron.yaml` and make that trigger on `main` pushes/pull_requests?

PR #795, opened 2023-11-09
javascript: Better errors when setting invalid values

alexjg:
Context: The change callback which is used to modify an automerge document is passed a proxy which looks like a POJO but intercepts mutating calls and converts them into operations on the underlying document.

Problem: When setting a key of one of the proxies in the change handler to `undefined` the proxy throws a RangeError. This is necessary because automerge doesn't support `undefined` - only `null`. However, the `RangeError` doesn't contain any information which allows the developer to figure out where they passed in an invalid value.

Solution: Include a JSON pointer to the path the developer attempted to insert the invalid value at so they can identify what part of their data they need to change.

### DIFF
--- a/javascript/src/proxies.ts
+++ b/javascript/src/proxies.ts
@@ -138,7 +138,11 @@ type ImportedValue =
   | [Record<string, any>, "map"]
   | [boolean, "boolean"]
 
-function import_value(value: any, textV2: boolean): ImportedValue {
+function import_value(
+  value: any,
+  textV2: boolean,
+  path: Prop[],
+): ImportedValue {
   switch (typeof value) {
     case "object":
       if (value == null) {
@@ -185,7 +189,9 @@ function import_value(value: any, textV2: boolean): ImportedValue {
         return [value, "str"]
       }
     default:
-      throw new RangeError(`Unsupported type of value: ${typeof value}`)
+      throw new RangeError(
+        `Unsupported type ${typeof value} for path ${printPath(path)}`,
+      )
   }
 }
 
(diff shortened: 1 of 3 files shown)


javascript/src/proxies.ts
```diff
@@ -948,3 +964,8 @@ function assertString(value: Text | string): asserts value is string {
     throw new Error("value was not a string")
   }
 }
+
+function printPath(path: Prop[]): string {
+  // print the path as a json pointer
+  return "/" + path.map(encodeURIComponent).join("/")
```
kid-icarus, line 970: Feels a little more ergonomic to use dot notation IMO, wdyt?

```suggestion
  return "." + path.map(encodeURIComponent).join(".")
```

kid-icarus:
The only other improvement I can think of here is to list out all of these errors. If I have an object with, lets say 5 undefined values, I'd have to resolve them one by one. Other than that this is very helpful, thank you!

javascript/src/proxies.ts
alexjg: Yeah that's probably a good shout, although at that point I should remove the comment about it being a JSON pointer (I mean, it's not really anyway because of the `encodeURIComponent` bit, maybe it's fine for this to be a bit ambiguous if there are key names with dots in them, it's just an error message)

alexjg: Actually, I've changed my mind. I think using JSON pointer is probably a bit better than making up a scheme which will in the end be quite similar to JSON pointer. (I was thinking about array indices, which look bad using a `.` separator and then thinking that I can't just use `[<index>]` because I'm sure there's a fair number of keys out there with `[` in them and then I ran out of energy and figured it's probably best to use something that is at least vaguely standardized.

alexjg:
Listing out the errors would be quite a bit more tricky because this code doesn't know whether it's running in `from` or  as part of a change callback. Consequently we just abort the whole operation immediately. To batch up the errors we would need to distinguish between the two cases so that in the case of `from` we can continue processing the rest of the imported value. This is a bit more work and maintenance burden than a better error message and so I think I'll forego it. This PR should mean that it's clear enough what went wrong and where that no one has to resort to console logging or hacking automerge, which is my main goal.

PR #796, opened 2023-11-10
Always send at least one sync message

alexjg:
Context: The sync protocol maintains a sync state which is updated every time we send or receive a message. This sync state contains a field for the heads the other end has reported. This is useful for applications to understand what local changes the other hand has not yet acknowledged.

Problem: In some cases we never send a response to the first message in the sync protocol and so the other end never learns what our heads are. The reason this happens is that we have some optimisations in the protocol to not send a message if we can tell that the other end has all the same changes as us. One scenario where this happens is if you load a document you previously saved and then synchronise that document with a sync server which already has the document and there have been no new changes.

Solution: Add a field to the sync state to track whether we have ever sent a response and ensure we always send a response if we have not sent one.

### DIFF
--- a/javascript/test/sync_test.ts
+++ b/javascript/test/sync_test.ts
@@ -65,7 +65,7 @@ describe("Data sync protocol", () => {
         assert.deepStrictEqual(message.changes, [])
       })
 
-      it("should not reply if we have no data as well", () => {
+      it("should not reply after the first round if we have no data as well", () => {
         let n1 = Automerge.init(),
           n2 = Automerge.init()
         let s1 = initSyncState(),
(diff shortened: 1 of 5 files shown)


ept:
When you say "track whether we have ever sent a response", I assume you mean "track whether we have sent a sync message since we last unserialised the sync state"? That is, if we disconnect and reconnect again, and therefore serialise and unserialise the sync state, we also reset the flag and therefore we will once again send a sync message after reconnection?

Is the idea that we will track a timestamp of when we last received some heads from the peer? In that case, we might want to update that timestamp not only once on the sync message when we connect, but also continuously as long as the connection is open. (That is, the fact that a connection is open but no change has been sent over the connection implicitly tells us that the old heads are still valid.)

pvh:
I don't think I'd use an "open" connection as a metric since "open" is not really a thing we can be sure of without exchanging messages. We might think we have a connection but be unable to send messages through it anymore, or the connection could be half-duplex (as with a congested asymmetrical connection) or any number of other problems.

Further, we will probably have network adapters which don't maintain a connection state, such as UDP. I think we should just track "when did we last hear from this peer".

pvh:
(It's also worth noting that this interface will likely change as we evolve support for multi-document synchronization. We could have positive confirmation of synchronization for many documents by exchanging a single MST root.)

PR #797, opened 2023-11-10
Unified ops

orionz:
Move all ops into a single append only vec per doc for less fragmented memory and fewer allocations.


rust/automerge/src/op_set.rs
```diff
@@ -229,14 +301,27 @@ impl OpSetInternal {
         }
     }
 
+    pub(crate) fn load(&mut self, obj: ObjId, op: Op) -> OpIdx {
+        self.osd.push(obj, op)
+    }
+
+    // want to move to this everywhere
+    pub(crate) fn load2(&mut self, obj: ObjId, op: Op, range: &mut OpIdxRange) -> OpIdx {
```
alexjg, line 309: Can we just move to this now and get rid of `load`?

rust/automerge/src/op_set/op.rs
```diff
@@ -0,0 +1,460 @@
...
+        self.0 as usize
+    }
+
+    pub(crate) fn as_op2(self, osd: &OpSetData) -> Op2<'_> {
+        Op2::new(self.0 as usize, osd)
+    }
+}
+
+#[derive(Debug, Copy, Clone)]
+pub(crate) struct Op2<'a> {
```
alexjg, line 29: Can we name this  `OpRef`?

rust/automerge/src/op_set/op.rs
```diff
@@ -0,0 +1,460 @@
...
+impl OpIdx {
+    pub(crate) fn new(index: usize) -> Self {
+        Self(index as u32)
+    }
+
+    pub(crate) fn get(&self) -> usize {
+        self.0 as usize
+    }
+
+    pub(crate) fn as_op2(self, osd: &OpSetData) -> Op2<'_> {
```
alexjg, line 23: Rather than having this `as_op2` method, can we instead have a method on `OpSetData::as_op_ref(idx: OpIdx)`. I feel like that would better communicate that the result is a reference to something indexed by `OpIdx` (see my other comment about renaming `Op2 -> OpRef`).

rust/automerge/src/query/nth.rs
```diff
@@ -55,35 +63,37 @@ impl<'a> Nth<'a> {
...
+        self.list_state.pos()
     }
 }
 
 impl<'a> TreeQuery<'a> for Nth<'a> {
-    fn equiv(&mut self, other: &Self) -> bool {
-        self.index() == other.index() && self.key() == other.key()
-    }
+    /*
+        fn equiv(&mut self, other: &Self) -> bool {
```
alexjg, line 82: nit: remove

alexjg (commented):
This all looks good to me with the exception of the names. I left a few comments regarding that.

PR #800, opened 2023-11-14
Faster load

orionz:
* Rewrote the reconstruct_document code to take advantage of unified_ops.
* Changed the way `Change::new()` worked to not have to re-decode all the ops 
* Removed indexes from maps and rewrote a few queries to make that possible.  This should make all map access much faster.  
* Delay index creation until after load is complete.
* Moved succ and pred into an OpDep unified table 
* Renamed `Op` to `OpBuilder` and `Op2` to `Op`

Open issues:

* The fact that indexes are an option now could require new tests.  What happens if list api calls are made against unindexed maps?
* Rollback is poorly tested and many of these changes intersected with it.  The unified_ops allows for a very clean rollback codepath. A rollback rewrite is recommended.
* Reconstruct document could be further optimized.  Build the ChangeGraph in the first pass - remove PartialChange.  Dont index if we are doing an incremental load.  Etc.
* Add and remove succ/pred really needs new tests.

Further performance work:

* It should be easy now to remove the history and generate changes on demand.  Should cut memory useage by 60%
* Index.visible is super slow and could be optimized 
* Index.ops could be replaced with a bloom filter and save a ton of time and memory.


rust/automerge/src/op_tree/node.rs
```diff
@@ -422,11 +462,15 @@ impl OpTreeNode {
                         cumulative_len += child.len() + 1;
                     }
                     Ordering::Equal => {
-                        return self.index.change_vis(vis);
+                        //return self.index.change_vis(vis);
```
alexjg, line 465: nit: remove commented code

rust/automerge/src/query/nth.rs
```diff
@@ -52,9 +52,11 @@ impl<'a> Nth<'a> {
         marks.current().cloned()
     }
 
-    pub(crate) fn pred(&self) -> OpIds {
-        self.osd.sorted_opids(self.ops.iter().map(|op| *op.id()))
-    }
+    /*
```
alexjg, line 55: nit: remove commented code

rust/automerge/src/storage/convert/op_as_docop.rs
```diff
@@ -135,7 +135,8 @@ impl<'a> Iterator for OpAsDocOpSuccIter<'a> {
     type Item = DocOpId;
 
     fn next(&mut self) -> Option<Self::Item> {
-        if let Some(s) = self.op.succ().get(self.offset) {
+        // FIXME
```
alexjg, line 138: What does this refer to?

rust/automerge/src/op_tree.rs
```diff
@@ -292,7 +307,53 @@ impl OpTreeInternal {
...
+
+    pub(crate) fn seek_idx<'a>(
+        &'a self,
+        idx: OpIdx,
+        encoding: ListEncoding,
+        clock: Option<&Clock>,
+        osd: &'a OpSetData,
+    ) -> Option<FoundOpId<'a>> {
+        let op = idx.as_op(osd);
+        if op.key().is_map_key() {
```
alexjg, line 349: What's the reason to use `is_map_key` here instead of just matching on `key()`? The latter seems clearer to me.

rust/automerge/src/storage/convert/op_as_docop.rs
orionz: Ah - good call - ill expand upon the FIXME - nth() is no longer fast b/c  the iterator changed - fixing wasn't trivial b/c the iterator is anonymous

PR #803, opened 2023-11-19
WIP: send the whole document in the first sync message

alexjg:
**Note: this currently does not work, there are breaking tests in the WASM interop tests.**

This PR modifies the sync protocol to send the whole document in the first sync message if the other end tells us they have nothing in _their_ first sync message.

This is conceptually simple but required a fair amount of fiddling to stay backwards compatible. The problem is this: the sync protocol has no versioning or message tagging mechanism, we just check the first byte of the message to see if it's a particular magic byte and if so we parse and process it. This means that peers running the new sync protocol can't just send the whole document down the wire, we need to know that the other end will support it first. The way I've achieved that here is to add a notion of "support capabilities", which are the things each end will support. The first message a peer sends is a standard v1 sync message, but appended at the end of it are a list of supported capabilities. Initially there are two capabilities a "message v1" and "message v2"  capability. The sending peer appends it's supported capabilities to the first sync message and the receiving peer stores those capabilities in it's sync state.

The end result of this is that after the first sync message the receiving peer knows whether the other end supports v2 sync messages and if it does support them, and also has no heads, sends it the whole document.

Another thing that has made this PR more complicated than it should be is that I copied the entire original sync implementation into a module called `sync::v1_compat_tests` (which is only compiled for tests) so that we can run compatibility tests against the old implementation.


alexjg:
@jkankiewicz there's a part of `automerge-c` which is impacted by this. The sync message may now contain a compressed document instead of a list of changes and so `AMSyncMessageChanges` might need to change. I can hack something in there right now but do you have a preferred way of addressing that?

jkankiewicz:
> I can hack something in there right now but do you have a preferred way of addressing that?

@alexjg, I'll take a look at what you did and let you know what I think.

alexjg:
Well, right now it's a `todo!` 😅

jkankiewicz:
The C API's unit tests must be disabled in the CI workflow; the unit test case  `test_should_send_a_sync_message_implying_no_local_data()` is still calling the no-longer-extant `AMsyncMessageChanges()`.

PR #804, opened 2023-11-22
rust: Improve some intradoc links

MarijnS95:
Found some markdown that was either:
- Missing backticks;
- Had the ` `` ` around `[]` so that it wouldn't turn into a link but display as a single-item array;
- Missing `[]` so that it would not link to an item/function;
- Some invalid `Self::` references that should have pointed to another type (rustdoc points this out and it helps the user understand and click what the docs are referring to);
- Added `()` at the end of functions for readability (and forces `rustdoc` to ensure that the target item really is an `fn`).

### DIFF
--- a/rust/automerge/Cargo.toml
+++ b/rust/automerge/Cargo.toml
@@ -20,12 +20,12 @@ sha2 = "^0.10.0"
 thiserror = "^1.0.16"
 itertools = "^0.10.3"
 flate2 = "^1.0.22"
-uuid = { version = "^1.2.1", features=["v4", "serde"] }
-smol_str = { version = "0.2", features=["serde"] }
+uuid = { version = "^1.2.1", features = ["v4", "serde"] }
+smol_str = { version = "0.2", features = ["serde"] }
 tracing = { version = "^0.1.29" }
 fxhash = "^0.2.1"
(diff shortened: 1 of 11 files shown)


MarijnS95:
Note that I only went over some of the high-value files/classes. Given this low-hanging fruit a larger and more rigorous pass is likely needed.

---

I personally prefer to have all function-links suffixed with `()`, because that:

1. Signifies to the user that we're referring to an _action_ rather than an object/variant/value/field;
2. Signifies to `rustdoc` that the link in question is only valid if the target really is a function.

alexjg:
Apologies for missing this, would you mind rebasing on to `main` to fix the conflicts and then I'll go ahead and merge.

MarijnS95:
No problem. @alexjg what are your thoughts about https://github.com/automerge/automerge/pull/804#issuecomment-1822446543?

MarijnS95:
Conflict with https://github.com/automerge/automerge/pull/818/files - which I also solved in this PR already - resolved! :tada: 

> Apologies for missing this, would you mind rebasing on to `main` to fix the conflicts and then I'll go ahead and merge.

Maybe the right time to merge this, so that I don't have to keep resolving simple conflicts :)

alexjg:
Thanks for the contribution and the persistence! Apologies for taking a while to merge.

PR #809, opened 2023-11-28
js: don't use put with text, fixes #677

### LINKED ISSUE #677: Automerge.Text manipulations can get mangled when en/decoded through changes
### Description
While investigating #675, I found that some of my Automerge.Text manipulations get mangled after being loaded via `loadIncremental`, `applyChanges`, or the sync protocol.

### Repro
```ts
const Automerge = require("@automerge/automerge");

let original = Automerge.from({ t: new Automerge.Text("") });
original = Automerge.change(original, (doc) => changeText(doc.t, `Billy Wald "You’re great"`));
original = Automerge.change(original, (doc) => changeText(doc.t, `Unknown`));
original = Automerge.change(original, (doc) => changeText(doc.t, `billy culinary dropout`));

// Case 1: original output (ok)
console.log("original:", original.t.toString());

// Case 2: using applyChanges (fails)
const changes = Automerge.getAllChanges(original);
const [replayed] = Automerge.applyChanges(Automerge.init(), changes);
console.log("replayed using applyChanges:", replayed.t.toString());
```

Case 1 will output the expected string `billy culinary dropout` but case 2 outputs `billy culiknary dropowt`. The full code including `changeText` impl using [`odiff`](https://www.npmjs.com/package/odiff) is in this gist: https://gist.github.com/stephen/5aeac63971520dca4473b5b15f2c42e6

I have a few other example sets of text manipulation that also cause this output.

yarolegovich:
I traced the issue described in the linked ticket (#677).
For it to happen we need to have multiple nodes in the operations tree.
```ts
const multipleOpNodes = new Array(32).fill("0").join("")
d.text.insertAt(0, multipleOpNodes)
```
Then we need to reset the `never_seen_puts` flag:
```ts
d.text.set(0, "1")
```
And finally we need another put at position that gets us to the second child node of operations tree:
```ts
d.text.set(15, "2")
```
to end up in [this branch](https://github.com/automerge/automerge/blob/main/rust/automerge/src/query/list_state.rs#L90).

Because neither `process_text_node` nor `process_list_node` is called position won't be shifted by [the number of seen nodes](https://github.com/automerge/automerge/blob/main/rust/automerge/src/query/list_state.rs#L120) and instead of inserting text at position 15 we'll insert it at position 0.

My fix is to use splice instead of set in text proxy:
```patch
- return (this[index] = value)
+ context.splice(objectId, index, 1, value)
```
as suggested in the comment of the code I referenced before:
```
// text nodes are intended to only be interacted with splice()
// meaning all ops are inserts or deleted inserts
// the indexes are written with this assumption in mind
// if conflicted put()'s with different character widths exist
// we cannot trust the indexs and need to descend
```

This fix will work for JS, but why was it implemented like this in Rust? 
Won't it be better to throw InvalidOp when trying to `.put` on a text object?
```rust
#[test]
fn wrong_behaviour() {
    let mut doc = AutoCommit::new();
    let oid = doc.put_object(&automerge::ROOT, "t", ObjType::Text).unwrap();
    doc.splice_text(&oid, 0, 0, "00000000000000000000000000000000").unwrap();
    doc.put(&oid, 0, "1").unwrap();
    doc.put(&oid, 15, "2").unwrap();

    let changes: Vec<Change> = doc.get_changes(&[])
        .into_iter()
        .map(|c| c.clone())
        .collect();
    let mut another_doc = AutoCommit::new();
    another_doc.update_diff_cursor();
    another_doc.apply_changes(changes).unwrap();

    assert_eq!(doc.text(&oid), another_doc.text(&oid));
}
```
This test will fail 👆

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ build/
 .vim/*
 /target
 .DS_Store
+**/.idea/*
--- a/javascript/.mocharc.json
+++ b/javascript/.mocharc.json
@@ -1,0 +1,8 @@
+{
+  "extensions": ["ts"],
+  "spec": ["test/*"],
+  "node-option": [
+    "experimental-specifier-resolution=node",
+    "loader=ts-node/esm"
+  ]
+}
--- a/javascript/src/proxies.ts
+++ b/javascript/src/proxies.ts
@@ -906,7 +906,7 @@ function textMethods(target: Target) {
   const { context, objectId } = target
   const methods = {
     set(index: number, value: any) {
-      return (this[index] = value)
+      context.splice(objectId, index, 1, value)
     },
     get(index: number): AutomergeValue {
       return this[index]
--- a/javascript/test/text_v1.ts
+++ b/javascript/test/text_v1.ts
@@ -331,4 +331,19 @@ describe("Automerge.Text", () => {
   it("should support slice", () => {
     assert.strictEqual(s1.text.slice(0).toString(), s1.text.toString())
   })
+
+  it("should be able to move insert, delete and set changes to a new document", () => {
+    let doc = Automerge.from({ text: new Automerge.Text("") })
+    doc = Automerge.change(doc, d => {
+      const multipleOpNodes = new Array(32).fill("0").join("")
+      d.text.insertAt(0, multipleOpNodes)
+      d.text.set(0, "1")
+      d.text.set(15, "2")
+    })
+    const [newDoc] = Automerge.applyChanges(
+      Automerge.init<{ text: Automerge.Text }>(),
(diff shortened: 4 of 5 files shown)


alexjg:
Thanks for tracking down the cause of this issue, that's a bunch of work! The reason we use `set` rather than `splice` is for backwards compatibility. In the original (i.e. not `next`) API of automerge it is possible to insert objects into a text sequence and there is code out in the wild which uses this feature so for now we have to support it. At some point we will release automerge `3.0` and we'll remove this.

I think to fix #677 then we'll need to fix the underlying logic bug.

PR #827, opened 2023-12-18
WeakMap metadata plus many load/materialize/apply_patch performance upgrades

orionz:
* move wasm metadata from defineProperty to WeakMap (40% speedup of materialize)
* cleanup js internal metadata APIs
* dont generate patches if they aren't needed (major speedup on progressDocument() - large scene no longer tuns out of memory)
* more materialize cleanup and speedup
* fix `add_index` performance regression. (20% speedup of reconstruct_document)

### Issues:

~~The WeakMap is stored on the global object.  I could possibly store it in some kind of rust based global instead.  Are global contexts in all versions of js?~~
Weakmap is now managed externally to the wasm api

Names have changed - maybe not for the better - the weak_map is now called `metadata`, what was called `metadata` is now called `user_data` and all the hidden properties plus user_data are collectivly called `object_metadata`.   Would like a second opinion on this before making it final.

### DIFF
--- a/rust/automerge-wasm/src/export_cache.rs
+++ b/rust/automerge-wasm/src/export_cache.rs
@@ -4,7 +4,8 @@ use crate::Automerge;
 use automerge as am;
 use automerge::ChangeHash;
 use fxhash::FxBuildHasher;
-use js_sys::{Array, Function, JsString, Object, Reflect, Symbol, Uint8Array};
+use js_sys::{Array, Function, JsString, Object, Reflect, Uint8Array, WeakMap};
+use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::ops::RangeFull;
 use wasm_bindgen::prelude::*;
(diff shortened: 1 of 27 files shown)


rust/automerge-wasm/src/export_cache.rs
```diff
@@ -339,77 +418,111 @@ impl<'a> ExportCache<'a> {
...
+        for cow in self.objs.values() {
+            Object::freeze(&cow.outer);
+        }
     }
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct COWJsObj {
+    pub(crate) metadata: ObjMetadata,
+    pub(crate) outer: Object,
```
alexjg, line 499: I think the "inner" and "outer" terminology here is a bit confusing now that this is called COWJsObj. Do I understand it correctly that "outer" is the value we will be passing back to the user whilest "inner" is the unwrapped value (i.e. we need to combine it with the metadata.datatype to understand it)?

rust/automerge/src/exid.rs
```diff
@@ -68,6 +69,31 @@ impl ExId {
         }
     }
 
+    #[inline(never)]
+    pub fn to_unsafe_u8(&self) -> Vec<u8> {
```
alexjg, line 73: I guess this is necessary for performance? If so then given that these are public functions we should document what they do and _why_ they are unsafe.

PR #834, opened 2024-01-08
rust: Upgrade `itertools` dependency to `0.12`

MarijnS95:
It seems all uses in this library are still compatible.  The `0.10.5` release is over a year old and `0.11` about 7 months old, and we'd like to get rid of some duplicate dependencies in our tree.


MarijnS95:
Note that some crates use version ranges here, we might also write `>= 0.10, <0.13` and extend that when new releases are published. It'd be "required" to test in the CI that `-Zminimal-versions` keeps building then though.

alexjg:
This LGTM, thanks for fixing the lints in #836. If you can rebase the checks will re-run and we can merge.

MarijnS95:
@alexjg yes, the CI succeeds on this now too :tada:

PR #864, opened 2024-03-04
chore: fix lint issues

kid-icarus:
This implements a few clippy suggestions that were causing the lint GitHub action to fail.

### DIFF
--- a/rust/automerge/src/op_set/op.rs
+++ b/rust/automerge/src/op_set/op.rs
@@ -312,7 +312,7 @@ impl<'a> Op<'a> {
         }
     }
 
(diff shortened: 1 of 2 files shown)


alexjg:
You're doing great work here, thankyou!

I see `cargo deny` is complaining for reasons which should be fixed by #863 , would it be worth adding this change to that PR and I'll merge rebase it all into `main`?

kid-icarus:
> I see `cargo deny` is complaining for reasons which should be fixed by #863 , would it be worth adding this change to that PR and I'll merge rebase it all into `main`?

Sure! I'll merge these fixes in #863

kid-icarus:
Done :)

PR #866, opened 2024-03-06
Remove libc dependency

jdcole:
libc dependency blocked `wasm32-unknown-unknown` builds


alexjg:
I'm going to close this for now as it's been a while since it was submitted (many apologies for taking so long to get to it) and I expect that the author has moved on. @jdcole if you still want to merge this then I would accept the PR once the formatting is fixed.

PR #869, opened 2024-03-11
Rich Text preview alpha 1

alexjg:
This is a preview of the rich text API we are designing for automerge, which is based on an extension of Peritext. This PR is a draft and is more a place to discuss the API and design then to review the code.

You can play with this implementation by installing `@automerge/automerge@2.1.11-richtext-alpha.1`.

The goal of this project is to expose an API for rich text which is interoperable across different text editors, platforms, and programming languages. This prototype targets integration with Prosemirror, but the design is intended to be general and in particular to not be specific to HTML.

## The blocks extension to Peritext

Peritext models rich text as a sequence of characters along with formatting spans which annotate ranges of those characters. Formatting spans work well for representing inline formatting such as bold and italic characters, or links; but don't work so well for block structure such as lists, heading, and block quotes. To support blocks we add "block markers" which appear inline in the sequence of characters.

A "block" is an object with three properties: 

* A type, such as "paragraph"
* A list of parent block types
* A map of attributes. Attribute values are simple scalar values, such as strings, numbers, or booleans.

For example, a document with a single paragraph might look like this:

```
"hello world"
{type: "paragraph", parents: [], attributes: {}}
```

And a document with an ordered list might look like this:

```
{type: "ordered-list-item", parents: [], attributes: {}}
"item one"
{type: "ordered-list-item", parents: [], attributes: {}}
"item two"
```

A document with a nested list would look like:

``` 
{type: "ordered-list-item", parents: [], attributes: {}}
"item one"
{type: "ordered-list-item", parents: ["unordered-list-item"], attributes: {}}
"nested item one"
```

A point to note here is that there are no block markers in the document for parents which do not contain text. This allows blocks which are created at the same level in the block structure to merge without coordination.

## The blocks API

Block markers appear inline in the text. This means that a block marker counts as one character in the text which is important because it means that when you are modifying the text you need to take account of the block markers when calculating indexes. 

Rendering text using `Automerge.text` will render the block markers as unicode object replacement characters - which have a width of 1. This means that you _can_ modify the text using the standard string methods, but it will look a bit strange because all the text will be joined with object replacement characters rather than newlines.

Text is still inserted using `Automerge.splice` and marks are created using `mark`, but there are new APIs for handling blocks. `Automerge.spans` will give you the text grouped by formatting span and split up by block marker. `Automerge.splitBlock`, `Automerge.joinBlock`, and `Automerge.updateBlock` allow you to modify blocks precisely whilst `Automerge.updateBlocks` allows you to update the block structure of the entire text in one go.

Changes to blocks are emitted as `SplitBlock`, `JoinBlock`, and `UpdateBlock` patches.

### Prelude

```typescript
type Value = string | number | boolean | null | Date | Uint8Array

type Block = {
  type: string,
  parents: string[],
  attrs: {[key: string]: Value}
}

export type Span = 
    { type: "text", value: string, "marks"?: {[key: string]: Value }} 
  | { type: "block", value: Block }
```

### `Automerge.splitBlock`

```typescript
function splitBlock<T>(
  doc: Doc<T>,
  path: Prop[],
  index: number | Cursor,
  block: Block,
) {
```

This inserts a new block marker at the given index

### `Automerge.joinBlock`

```typescript
export function joinBlock<T>(
  doc: Doc<T>,
  path: Prop[],
  index: number | Cursor,
)
```

Delete the block marker at the given index, throws if the character at the index is not a block marker

### `Automerge.updateBlock`

```typescript
function updateBlock<T>(
  doc: Doc<T>,
  path: Prop[],
  index: number | Cursor,
  block: Block,
)
```

Update the block marker at the given index. This will be a no-op if nothing has changed.

### `Automerge.updateBlocks`

```typescript
function updateBlocks<T>(
  doc: Doc<T>,
  path: Prop[],
  blocks: (Block | string)[],
) {
```

Infer the sequence of `splice`, `splitBlock`, `joinBlock`, and `updateBlock` operations required to transform the current block structure of the document to match the given `blocks`. Note that this will modify the text as well as the block structure but it does not modify formatting spans.

### `Automerge.spans`

```typescript
function spans<T>(doc: Doc<T>, path: stable.Prop[]): Span[]
```

Return the text of the document split into spans of text and block markers. Text spans will be grouped by formatting span. For example, the following document:

```typescript
const doc = Automerge.change(Automerge.init(), doc => {
  doc.text = "hello world"
  Automerge.splitBlock(doc, ["text"], 0, {type: "paragraph", parents: [], attrs: {}})
  Automerge.splitBlock(doc, ["text"], 7, {type: "paragraph", parents: [], attrs: {}})
  Automerge.mark(doc, ["text"], {start: 1, end: 3}, "bold", true)
  Automerge.mark(doc, ["text"], {start: 3, end: 9}, "italic", true)
})
console.log(Automerge.spans(doc, ["text"]))
```

prints:

```
[
  {
    type: 'block',
    value: { type: 'paragraph', parents: [], attrs: {} }
  },
  { type: 'text', value: 'he', marks: { bold: true } },
  { type: 'text', value: 'llo ', marks: { italic: true } },
  {
    type: 'block',
    value: { type: 'paragraph', parents: [], attrs: {} }
  },
  { type: 'text', value: 'w', marks: { italic: true } },
  { type: 'text', value: 'orld' }
]
```

Note how the text between spans is grouped by the active marks.

### Patches

The following patches are added to represent block changes

```typescript
export type SplitBlockPatch = {
  action: 'splitBlock'
  path: Prop[],
  index: number,
  type: string,
  parents: string[],
  attrs: { [key: string]: Value },
}

export type UpdateBlockPatch = {
  action: 'updateBlock'
  path: Prop[],
  index: number,
  new_type: string | null,
  new_parents: string[] | null,
  new_attrs: { [key: string]: string | Value } | null,
}

export type JoinBlockPatch = {
  action: 'joinBlock'
  path: Prop[],
  index: number,
}
```

#### Block types

The objective of this work is to provide an interoperable API for rich text, but this implementation leaves an important part unspecified - the set of block types. In the Prosemirror implementation which I have been using to drive this work I have restricted the allowed block types to the following:

* `paragraph`
* `heading` with a numeric `level` attribute between 1 and 6
* `blockquote`
* `ordered-list-item`
* `unordered-list-item`
* `code-block`
* `image` with `src`, `alt`, and `title` attributes

I don't think we're going to find a set of block types which satisfies everyone, but is it possible to say something about what a block is that would allow applications that do not understand the block type to still display a useful UI? For example, in HTML we could say that if you don't know the block type, just render a `div`.

#### Embedded content

I implemented the `image` tag in the Prosemirror bindings as a block with a `src` attribute. Images are one case of a more general feature which is "embedded" content. Other examples would be including videos, maps, tables, or some custom UI based on the content of another automerge document. Implementing embedded content as a block has interoperability concerns.

Firstly, embedded content usually doesn't contain children, so it seems strange to implement it as a block marker. In the Prosemirror bindings I special cased the `image` block type so that the implementation doesn't consider an `image` block to split the block it follows. This won't work in a general setting where there are unknown block types.

Secondly, embedded content sometimes appears inline with the text but also sometimes in a block which breaks the flow of text. We could treat embedded content as always inline and require that the user wrap the embedded content with block markers, but this seemto me like 

a) it would make the editing experience strange for UIs which don't know about the embedded type, you could end up in strange situations where you deleted e.g. the closing block marker and the whole document becomes part of the embed
b) It would make it impossible to have embedded content which should _only_ be in a block context - such as a table.

Finally, embedded content is opaque to UIs which don't know about the type of the content. There's no way for the UI to render anything other than a block marker.

My current thinking is that the best solution to this is to introduce an "embed" type, which would be like a block marker but would appear as a different type in the `spans` the document returns. I would imagine the API looking something like the following:

```typescript
type Embed = {
  type: string,
  alt: string,
  isInline: boolean,
  attrs: {[key: string]: Value}
}

function Automerge.insertEmbed(doc, path, index, embed: Embed)
function Automerge.updateEmbed(doc, path, index, embed: Embed)
function Automerge.deleteEmbed(doc, path, index)

type Span = 
    { type: "text", value: string, "marks"?: {[key: string]: Value }} 
  | { type: "block", value: Block }
  | { type: "embed", value: Embed }

export type InsertEmbedPatch = {
  action: 'insertEmbed'
  path: Prop[],
  index: number,
  type: string,
  alt: string,
  isInline: boolean,
  attrs: { [key: string]: Value },
}
export type UpdateEmbedPatch = {
  action: 'updateEmbed'
  path: Prop[],
  index: number,
  new_type: string | null,
  new_alt: string | null,
  new_isInline: boolean | null,
  new_attrs: { [key: string]: string | Value } | null,
}
export type DeleteEmbedPatch = {
  action: 'deleteEmbed'
  path: Prop[],
  index: number,
}
```

An alternative would be to just add an `isEmbed` attribute to the standard block marker which would reduce the size of the API at the expense of slightly confusing the concept of a block.

#### The API for updating formatting spans

The `updateBlocks` API is very useful for creating robust adapters to existing text editors. In the Prosemirror bindings I special cased changes which are easy to turn into automerge changes (e.g. inserting a character) and use `updateBlocks` for anything more complicated. This means that the implementation doesn't become to coupled to how exactly Prosemirror represents particular kinds of change.

However, the `updateBlocks` API does not update formatting spans. I did play for a bit with trying to infer the formatting spans from the `Span[]` passed to `updateBlocks` but this runs into a problem. The automerge API deliberately doesn't expose what the expand behavior of a mark is, once you've set a mark the idea is that you leave the logic of how it should expand up to Automerge. This in turn allows us to ensure that everyone converges on the same UI for a given set of mark operations. 

Unfortunately, this means we can't infer everything we need from a `Span[]`. We can figure out what marks are active but we don't know how they are meant to expand. In the Prosemirror bindings this hasn't turned out to be a problem because Prosemirror likewise has a whole separate set of operations for manipulating marks which dovetail quite well with automerge - but I am concerned about how this will work in less similar text editors. This might be an argument for exposing the expand behavior and just recommending that people don't use it for rendering.


ept:
Love this! Some thoughts:

- Should we allow a block to be an arbitrary map object? I feel the proposed structure (with `type`, `parents`, and `attrs` fields) is perhaps a bit prescriptive. For example, Vicente @almeidavc's rich text editor uses a different model for nested blocks (instead of an array of parent block *types*, he references parent block *IDs*, which results in different trade-offs regarding the behaviour of changing indentation levels in nested bullet point lists). It would be nice to be able to support such alternative uses of block markers. Any rich text editors that want to interoperate will of course have to use the same structure for representing blocks, but this could be a convention rather than enforced by types at the Automerge level.
- Should we allow nested objects inside `attrs`? For example, imagine a custom `<blockquote>` block type that allows an attribute containing an array of authors, or a structured publication object, as the source of the quotation. This sort of feature doesn't need to be built into our default editor integrations, of course, but I could imagine some apps wanting to be able to customise block types to suit the needs of their specific domain.
- Instead of `updateBlock`, should we give a block marker an object ID and allow them to be updated using the same function calls as any other object? We don't do this for marks because marks are not objects, but my feeling is that block markers are more like objects than marks are. I think we shouldn't allow a mark to be updated (to change a mark, simply overwrite it with a new mark), but updating a block marker is meaningfully different, and should be supported in my opinion.
- For embedded content, I'm happy with either an Embed type that is distinct from a Block, or an `isEmbed` property. We might want to distinguish between an inline embed that is part of the paragraph (or whatever block of text) that surrounds it (such as an inline mathematical formula or a [sparkline](https://en.wikipedia.org/wiki/Sparkline)), and a block embed that is a block of its own (such as a figure or table). For a block embed, the question is: if the embed is immediately followed by text (not a block marker), what sort of block should it be put in? Perhaps a new block that has the same properties as the last block marker that precedes the embed.
- For an embedded table, it would be nice to be able to put all the tabular data as nested objects within the embed object (an alternative would be to have a reference to a separate document, but I think it would be good to make it possible to have everything in one self-contained document that can map to a single file on disk). Another reason for wanting nested objects.
- The question of expanding marks and `updateBlocks` is tricky. Need to think more about that one.

jessegrosjean:
I'm curious how IDs would work with block objects. Do blocks have associated IDs?

> For example, Vicente @almeidavc's rich text editor uses a different model for nested blocks (instead of an array of parent block types, he references parent block IDs, which results in different trade-offs regarding the behavior of changing indentation levels in nested bullet point lists)

I guess this is a more general question, but In a design like this who generates those block IDs and how are they kept unique? Is common practice to just use a UUID generator, or do you need to also incorporate actorID to be unique?

almeidavc:
> I'm curious how IDs would work with block objects. Do blocks have associated IDs?
> 
> > For example, Vicente @almeidavc's rich text editor uses a different model for nested blocks (instead of an array of parent block types, he references parent block IDs, which results in different trade-offs regarding the behavior of changing indentation levels in nested bullet point lists)
> 
> I guess this is a more general question, but In a design like this who generates those block IDs and how are they kept unique? Is common practice to just use a UUID generator, or do you need to also incorporate actorID to be unique?

In the model @ept is referring to, the id used to reference a parent block is the objectId produced by Automerge (the id of the operation that created the blocked consisting of the sequence number and the actor id). One can use these ids as they are unique.

ixxie:
I'm very new to automerge, but I recently worked on a project with the Notion [block API](https://developers.notion.com/reference/block). I'd like to share a couple of insights from this experience, since it may help.

## Life is easy when block types mirror HTML tags

Whenever the Notion API did this, it was easy to render in the app. Whenever it didn't, life was difficult. 

It's true that some people will render in other markup languages, but HTML is arguably the most common. This also means plenty of tooling to transform HTML to other markup languages exist.

Some frameworks like Svelte have [special elements](https://svelte.dev/docs/special-elements#svelte-element) that let you render an HTML tag dynamically. So ideally, block type strings would even be *exactly* the HTML tag they represent.

This approach has the added benefit that it gives you guidance in the API design, so you don't have to think about it too much.

## An extensibility mechanism would have been awesome

I wanted to develop functionality with 'custom' block types. However, the API didn't really support this, so I had to work around it.

If it's not an insane demand, some type of extensibility mechanism could be nice. Since the focus here is rich text rather than arbitrary blocks, perhaps a simple and effective mechanism could be to offer a `classes` attribute. In a web application, this would obviously correspond to CSS classes.

alexjg:
Closing in favor of #898

PR #874, opened 2024-03-15
Only commit conversion transaction when there are valid conversions

LiraNuna:
Fixes https://github.com/automerge/automerge/issues/872

When using `StringMigration::ConvertToText`, a transaction is made when loading the document. This isn't necessary and just causes ballooning of the document for no reason.

This PR corrects this issue by only committing the transaction if there are strings to convert.

### DIFF
--- a/rust/automerge/src/automerge.rs
+++ b/rust/automerge/src/automerge.rs
@@ -1599,12 +1599,16 @@ impl Automerge {
                 _ => {}
             }
         }
-        let mut tx = self.transaction();
-        for Conversion { obj_id, prop, text } in to_convert {
-            let text_id = tx.put_object(obj_id, prop, ObjType::Text)?;
-            tx.splice_text(&text_id, 0, 0, &text)?;
+
+        if !to_convert.is_empty() {
+            let mut tx = self.transaction();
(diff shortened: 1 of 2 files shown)


rust/automerge/src/automerge.rs
```diff
@@ -1599,12 +1599,16 @@ impl Automerge {
...
-            let text_id = tx.put_object(obj_id, prop, ObjType::Text)?;
-            tx.splice_text(&text_id, 0, 0, &text)?;
+
+        if !to_convert.is_empty() {
+            let mut tx = self.transaction();
+            for Conversion { obj_id, prop, text } in to_convert {
+                let text_id = tx.put_object(obj_id, prop, ObjType::Text)?;
+                tx.splice_text(&text_id, 0, 0, &text)?;
+            }
+            tx.commit();
```
LiraNuna, line 1609: Alternatively, only this line could be surrounded with the `if`, but I think it is more readable to surround the entire thing.

alexjg:
This looks great, I'll fix the `cargo deny` errors now then you should be able to rebase on that to get our particularly grumpy CI to give us a thumbs up.

alexjg:
Alright, if you rebase on `main` CI should pass.

PR #895, opened 2024-04-01
adds an exemplar document

heckj:
Follow up from https://github.com/automerge/automerge-swift/issues/139, which identified a bug in the Swift language overlay covering the representation of timestamps in the user data of an Automerge document. I also included some of the complex unicode representations that fouled up the swift language code earlier re: Emojis and text representation.

For cross-platform validation of data representation. Covers times with millisecond granularity, and text with multi-byte emoji characters embedded within.

Text does not include revisions, marks, or any block attributes


lightsprint09 (commented):
Looks good to me. I would add list of every type as well

interop/README.md
```diff
@@ -0,0 +1,49 @@
...
+        133,
+        111,
+        74,
+        131
+    ],
+    "counter": 5,
+    "fp": 3.14159267,
+    "int": -4,
+    "location": "https://automerge.org/",
+    "notes": [
```
alexjg, line 33: This is actually not quite right any more. I updated the `automerge-cli` package to display `Text` as JSON strings.

alexjg:
One small comment about the README but otherwise looks good. I think the next step on this road is to define a JSON encoding of the automerge model which we can then run tests against. This would allow us to say "write a program which accepts an automerge document on the command line and outputs the standard JSON encoding" and then we can run interop tests against that program.

This is pretty low priority though and just getting an exemplar document in with a description is much better than where we were.

interop/README.md
heckj: Ooh - nice, I must have an older target lurking there. I'll rebuild and update it.

Issue #919, opened 2024-05-06
Error on load (with 5.10 release, after syncing back and forth a bit with sync.automerge.org) - error inflating document chunk ops: an operation referenced a missing actor id


heckj:
I don't have a solid reproduction of this issue - I was working on debugging the sync and connections for automerge-repo-swift (using the demo app MeetingNotes) when I encountered this. The libraries were all recently updated to the 0.5.10 core library release, not sure of the state of `sync.automerge.org`. I'm suspecting there was an issue in the flight back and forth.

It appeared when I loaded a binary after debugging sessions, and the one on fisk started throwing this error:

```
error inflating document chunk ops: an operation referenced a missing actor id
```

I'm attaching the raw file (as compressed/ZIP version since GitHub it cranky about file types) for inspection:

[rawautomergefile.zip](https://github.com/automerge/automerge/files/15227475/rawautomergefile.zip)

heckj:
duplicate of #897 - with an example file to show the result.

alexjg:
This issue should be fixed in `0.5.11`. I'm going to close this but please reopen if it occurs again.

PR #920, opened 2024-05-07
Fix corruption bug

### LINKED ISSUE #897: Error inflating document chunk ops: an operation referenced a missing actor id
Hello!

I've run across an issue where automerge 0.5.9 generates invalid saves that give `error inflating document chunk ops: an operation referenced a missing actor id` when attempting to load them. I've managed to reproduce this in a minimized test case:

```rust
use anyhow::Result;
use automerge::{ActorId, AutoCommit, ReadDoc, ROOT};
use automerge::transaction::Transactable;

// error inflating document chunk ops: an operation referenced a missing actor id
fn main() -> Result<()> {
    let actor0 = ActorId::from(&[0]);
    let actor1 = ActorId::from(&[1]);
    let actor2 = ActorId::from(&[2]);

    let mut doc0 = AutoCommit::new().with_actor(actor0);
    doc0.put(ROOT, "a", 1)?;

    // swap these actors and no error occurs
    let mut doc1 = doc0.fork().with_actor(actor2);
    let mut doc2 = doc0.fork().with_actor(actor1);

    doc1.put(ROOT, "b", 2)?;
    doc2.merge(&mut doc1)?;
    doc2.delete(ROOT, "c")?;

    // error occurs here
    doc2.save_and_verify()?;

    println!("{:?}", doc2.get(ROOT, "a"));
    println!("{:?}", doc2.get(ROOT, "b"));
    println!("{:?}", doc2.get(ROOT, "c"));

    Ok(())
}
```

Interestingly it seems to depend on the ordering of ActorIDs as if you swap `actor1` and `actor2` the error doesn't occur. With random ActorIDs the problem is not 100% reproducible.

It also seems to be related to the `doc2.delete(ROOT, "c")` operation which has no effect since the key `"c"` isn't set in the document.

Anyhow, is this a bug or am I misusing the API somehow? The behavior doesn't seem to be present in 0.5.6. Thank you!

alexjg:
Context: when we save a document we encode all the actor IDs in the
document at the start of the encoded document and then refer to those
actor IDs by their index into the encoded array. Similarly, in the
runtime representation of ops we don't directly store the actor ID but
instead refer to an index into a table of actor IDs held elsewhere. This
means that when we write the document to storage we need to translate
from the runtime indexes to the stored indexes.

Problem: in some cases we were creating actor indexes in the runtime
table of actors which we never wrote to the document. This meant that
the runtime actor indexes could get out of sync with the actors we write
to the document. This resulted in the actor indexes being wrong when
loading the document and consequently in "missing actor ID" errors.

Solution: when encoding actor IDs don't use the runtime table as a
translation table directly. Instead build a new table which contains
only the actor IDs we know we are writing.

Fixes #897 #919


heckj:
THANK YOU THANK YOU THANK YOU - this one was biting me constantly during testing!

PR #922, opened 2024-05-09
Allow empty keys in maps

### LINKED ISSUE #852: Why can't keys be empty?
Per [1], keys can't be empty strings in Automerge maps. I've run into a situation in the wild where my JS data naturally contains the empty string as a key, so this will be a bit awkward to work around.

I'm curious if there's a reason Automerge excludes `""` as a key, or if this is just a provisional decision.

(Thanks!)

[1] https://github.com/automerge/automerge/blob/a0ac03edd4c73b78ee3e1bf3b4583bc16804232f/rust/automerge/src/transaction/inner.rs#L419

LiraNuna:
According to discussion over on #852, there really isn't a reason why keys can't be empty. 
This PR allows this by simply removing the check and error.

Resolves #852

### DIFF
--- a/javascript/test/error.ts
+++ b/javascript/test/error.ts
@@ -15,15 +15,4 @@ describe("Automerge errors", () => {
 
     assert(error instanceof Error)
   })
-
-  it("Automerge.from throws an error, not a string", () => {
-    let error
-    try {
-      Automerge.from({ "": "bad key" })
-    } catch (err) {
-      error = err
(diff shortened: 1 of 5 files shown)


LiraNuna:
@alexjg it seems that `structuredClone` was added in node18 but I guess for compatibility reasons Automerge targets 16. We could remove the test or figure out a way to run tests with 18

PR #928, opened 2024-05-14
When resolving a cursor allow to check if the op that the cursor points to is present in the document

paulsonnentag:
I want to be able to resolve a cursor in a previous document, but only if the op that the cursor points to is actually present in that version of the document. I want to use cursors to implement comments that can point to regions of the text. If I'm looking at an older version of the document I want to filter out comments that point to characters that didn't exist at that point in time. I think at some point automerge threw an exception if the cursor couldn't be resolved. I think it's better to have an API that signals that the cursor couldn't be resolved through a return value.


pvh:
This is a good one for @alexjg to take a gander at.

alexjg:
This is straightforward to implement, we just need to figure out the API. At the moment `getCursorPosition` always returns a `number`, or throws an exception if the cursor doesn't point at something in the text object in question. If the document is currently pointing at a version before the cursor was created, we resolve to the first visible character preceding the target of the cursor.

Now, we could update the `getCursorPosition` call to take an argument like `resolveFuture: boolean`, which if `false` means that `getCursorPosition` returns `null`. We can even write the typescript types such that this isn't a breaking change. 

I think a better option would be to add an `isCursorVisible(doc, <heads>, cursor)` call, which could be used to check for this kind of thing.

FintanH:
Going to close this since the commits seem to only resolve to a single test, and it's been 2 years.

PR #929, opened 2024-05-15
Add tests documenting merge behavior for independently created docs 

HerbCaudill:
I created these tests while trying to clear up my mental model of how Automerge handles merging documents that were created independently (e.g. on different peers). (See [this Discourse thread](https://discord.com/channels/1200006940210757672/1230453235207110666) for context.)

It turned out that Automerge's behavior is very different from my prior intuitions, so I thought it would be worth having these tests in the codebase as a way of documenting how it actually works.

### DIFF
--- a/javascript/test/basic_test.ts
+++ b/javascript/test/basic_test.ts
@@ -233,6 +233,7 @@ describe("Automerge", () => {
       assert.deepEqual(doc3.list[2], 3)
       assert.deepEqual(doc3, { list: [1, "a", 3] })
     })
+
     it("handle simple lists", () => {
       let doc1 = Automerge.init<any>()
       let doc2 = Automerge.change(doc1, d => {
@@ -243,6 +244,7 @@ describe("Automerge", () => {
       let [docB2] = Automerge.applyChanges(docB1, changes)
       assert.deepEqual(docB2, doc2)
     })
+
     it("handle text", () => {
       let doc1 = Automerge.init<any>()
       let doc2 = Automerge.change(doc1, d => {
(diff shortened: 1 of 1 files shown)


javascript/test/basic_test.ts
```diff
@@ -801,4 +808,41 @@ describe("Automerge", () => {
...
+    it("merges non-conflicting fields, using Automerge.from", () => {
+      // shorthand for the above
+      let doc1 = Automerge.from<any>({ x: 1 })
+      let doc2 = Automerge.from<any>({ y: 2 })
+      const merged = Automerge.merge(doc1, doc2)
+      // both properties are preserved
+      assert.deepEqual(merged, { x: 1, y: 2 })
+    })
+
+    it("can't merge independently created arrays", () => {
```
pvh, line 835: technically these are conflicts -- if you check there you'll see them.

pvh: also @ept and @alexjg is it really wise for us to have a hardcoded object ID for the root?

ept: In my mind, whether the root object has a fixed object ID or not is an implementation detail; the real question is whether we want to allow merging of documents that were independently created and share no change history. If we wanted to rule out merging of independently created documents, a variable root ID that is created in the first change would be one way of achieving that, but a simpler solution would be to simply compare two docs' oldest change hash when merging, and refuse to merge if they don't match. But even if we restrict merges of independently created docs, that doesn't really solve the issue that Herb's tests here highlight, because the creation of the conflicting arrays need not be in the first change of the document's history, it could also happen later. The real solution, in my opinion, is to [merge concurrently created objects under the same key](https://github.com/automerge/automerge-classic/issues/4).

alexjg:
I'm going to close this PR as we already have tests which exercise these features and I think we should address the documentation problem by adding better docs to the [conflicts page](https://automerge.org/docs/documents/conflicts/) or otherwise improving the documentation.

Issue #891, opened 2024-03-24
undefined `context` in `import_value`


joshuahhh:
Just ran into a reference to an undefined variable `context` in some Automerge code. It looks like TypeScript didn't catch it because Mocha defines "context" as a global. (I'm not good enough at TS to know how to restrict the Mocha types to test files, unfortunately.)

```
file:///.../node_modules/@automerge/automerge-repo/node_modules/@automerge/automerge/dist/mjs/proxies.js:102
            else if (isSameDocument(value, context)) {
                                           ^

ReferenceError: context is not defined
    at import_value (file:///.../node_modules/@automerge/automerge-repo/node_modules/@automerge/automerge/dist/mjs/proxies.js:102:44)
    at Object.set (file:///.../node_modules/@automerge/automerge-repo/node_modules/@automerge/automerge/dist/mjs/proxies.js:185:35)
    at <anonymous> (/.../src/server/run.ts:382:18)
    at _change (file:///.../node_modules/@automerge/automerge-repo/node_modules/@automerge/automerge/dist/mjs/stable.js:356:9)
    at Module.change (file:///.../node_modules/@automerge/automerge-repo/node_modules/@automerge/automerge/dist/mjs/stable.js:233:16)
    at callback (file:///.../node_modules/@automerge/automerge-repo/dist/DocHandle.js:301:30)
    at file:///.../node_modules/@automerge/automerge-repo/dist/DocHandle.js:152:36
    at /.../node_modules/xstate/lib/utils.js:415:23
    at Array.reduce (<anonymous>)
    at Object.updateContext (/.../node_modules/xstate/lib/utils.js:403:48)
    at handleAction (/.../node_modules/xstate/lib/actions.js:586:34)
    at processBlock (/.../node_modules/xstate/lib/actions.js:627:24)
    at Array.map (<anonymous>)
    at Object.resolveActions (/.../node_modules/xstate/lib/actions.js:652:52)
    at StateNode.resolveTransition (/.../node_modules/xstate/lib/StateNode.js:917:36)
    at StateNode.transition (/.../node_modules/xstate/lib/StateNode.js:840:17)
    at /.../node_modules/xstate/lib/interpreter.js:962:28
    at Object.provide (/.../node_modules/xstate/lib/serviceScope.js:12:16)
    at Interpreter._nextState (/.../node_modules/xstate/lib/interpreter.js:961:34)
    at /.../node_modules/xstate/lib/interpreter.js:108:31
[...]

PR #932, opened 2024-05-16
Correctly pass `context` parameter to `import_value`

### LINKED ISSUE #914: [BUG] ReferenceError: context is not defined in `proxies.ts`
If you manage to trigger a range error, you get another exception instead, namely:

`ReferenceError: context is not defined`

The issue is on [this line](https://github.com/automerge/automerge/blob/4999578a7ca138e59efa12a90a1dd9380472e9f1/javascript/src/proxies.ts#L170), where the variable context is not defined.

LiraNuna:
The test for a recursive document as a value was broken as a parameter `context` was omitted from the function arguments. This was not caught by TypeScript as `context` is a part of mocha in tests.

This PR corrects this issue and adds an appropriate tests to catch future issues like this.

Resolves #914

### DIFF
--- a/javascript/src/proxies.ts
+++ b/javascript/src/proxies.ts
@@ -142,6 +142,7 @@ function import_value(
   value: any,
   textV2: boolean,
   path: Prop[],
+  context: Automerge,
 ): ImportedValue {
   switch (typeof value) {
     case "object":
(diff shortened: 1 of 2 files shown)


joshuahhh:
#891 author here. Thanks for fixing this. Do you know how to prevent Mocha from making `context` a global everywhere?

LiraNuna:
I was hoping to figure it out and fix that in a separate PR. Not sure why mocha types were applied to source code.

PR #944, opened 2024-06-30
automerge-c: Expose Cursor API

jkankiewicz:
@alexjg, I've finally exposed the Cursor API to C programs.

In order to make the full unit test suite pass again, I also included the changes from PR #885 that I authored.

@heckj can either rebase PR #885 onto main after this PR is approved or he can open a new PR for his changes.

### DIFF
--- a/rust/automerge-c/src/doc.rs
+++ b/rust/automerge-c/src/doc.rs
@@ -417,17 +417,11 @@ pub unsafe extern "C" fn AMgetCursor(
 ) -> *mut AMresult {
     let doc = to_doc!(doc);
     let obj_id = to_obj_id!(obj_id);
-    match heads.as_ref() {
-        None => {
-            return to_result(doc.get_cursor(obj_id, position, None));
-        }
+    return match heads.as_ref() {
+        None => to_result(doc.get_cursor(obj_id, position, None)),
         Some(heads) => match <Vec<am::ChangeHash>>::try_from(heads) {
-            Ok(heads) => {
-                return to_result(doc.get_cursor(obj_id, position, Some(heads.as_slice())));
-            }
-            Err(e) => {
-                return AMresult::error(&e.to_string()).into();
-            }
+            Ok(heads) => to_result(doc.get_cursor(obj_id, position, Some(heads.as_slice()))),
+            Err(e) => AMresult::error(&e.to_string()).into(),
         },
     };
 }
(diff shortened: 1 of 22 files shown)


jkankiewicz:
@alexjg, clippy is complaining about needless return statements but when I remove them then automerge-c won't build on my local machine.
I'm using rustc 1.79.0.

heckj:
You can ditch my changes in #885 and we can just close it entirely if they pass in this. The whole purpose was to simply check the cmake build actually running again.

rust/automerge-c/src/doc.rs
```diff
@@ -386,6 +387,102 @@ pub unsafe extern "C" fn AMgetChangesAdded(doc1: *mut AMdoc, doc2: *mut AMdoc) -
...
+/// \brief Gets the absolute position of an `AMcursor` within a list object or
+///        text object.
+///
+/// \param[in] doc A pointer to an `AMdoc` struct.
+/// \param[in] obj_id A pointer to an `AMobjId` struct or `AM_ROOT`.
+/// \param[in] cursor A pointer to an `AMcursor` struct.
+/// \param[in] heads A pointer to an `AMitems` struct with `AM_VAL_TYPE_CHANGE_HASH`
+///                  items to select a historical object or `NULL` to select the
+///                  current object.
+/// \return A pointer to an `AMresult` struct with an `AM_VAL_TYPE_UINT` item.
```
whereswaldon, line 445: Perhaps this is the wrong place for it, but it would be helpful to specify the units of the return value. Is it measuring code points, bytes, or something else?

rust/automerge-c/src/item.rs
```diff
@@ -1234,34 +1299,36 @@ pub enum AMvalType {
...
+    SyncState = 1 << 17,
     /// A *nix timestamp (milliseconds) value.
-    Timestamp = 1 << 17,
+    Timestamp = 1 << 18,
     /// A 64-bit unsigned integer value.
-    Uint = 1 << 18,
+    Uint = 1 << 19,
     /// An unknown type of value.
-    Unknown = 1 << 19,
+    Unknown = 1 << 20,
```
whereswaldon, line 1331: Perhaps I am misunderstanding how these values are used, but redefining these flags seems like it has the potential to break any serialized/binary representations of automerge objects created via the C API. Am I misunderstanding the function of this type?

whereswaldon (commented):
Thanks so much for working on this! I have a couple of questions about the implementation if you don't mind @jkankiewicz.

rust/automerge-c/src/item.rs
jkankiewicz: > redefining these flags seems like it has the potential to break any serialized/binary representations of automerge objects created via the C API.

@whereswaldon, those values are not serialized because they're just a way to refer to Rust enum variants from within C; all serialization/deserialization is handled by the underlying Rust library.
I've inserted tags in alphabetical order into that enum before and doing so has never broken a unit test.
Being that everyone is taught to prefer symbolic values over magic numbers I doubt that anyone would choose to use an enum tag's opaque integer value instead of its descriptive name.

rust/automerge-c/src/doc.rs
jkankiewicz: Since PR #641, indices are assumed to be in terms of UTF-8 characters but you can build a version that uses UTF-32 indexing using the `AUTOMERGE_C_UTF32` compile time flag.

jkankiewicz:
> To be clear, indices are byte positions within a UTF8 stream by default?

@whereswaldon, for an `AM_ OBJ_TYPE_TEXT` object, yes. For an ` AM_OBJ_TYPE_LIST` object it's always the index of a value regardless of how the library is compiled.

rust/automerge-c/src/doc.rs
jkankiewicz: > To be clear, indices are byte positions within a UTF8 stream by default?

@whereswaldon, for an `AM_ OBJ_TYPE_TEXT` object, yes. For an ` AM_OBJ_TYPE_LIST` object it's always the index of a value regardless of how the library is compiled.

whereswaldon: Makes sense, thanks for clarifying.

PR #952, opened 2024-07-23
Make `ChangeOptions.time` default to current timestamp 

marionauta:
Also, document that this value should be in milliseconds.

### DIFF
--- a/javascript/src/stable.ts
+++ b/javascript/src/stable.ts
@@ -85,8 +85,11 @@ import { stableConflictAt } from "./conflicts.js"
 export type ChangeOptions<T> = {
   /** A message which describes the changes */
   message?: string
-  /** The unix timestamp (in milliseconds) of the change (purely advisory, not used in conflict resolution), defaults to current timestamp */
-  time?: number
+  /**
+   * The unix timestamp (in milliseconds) of the change (purely advisory, not used in conflict resolution).
+   * When omitted it defaults to the current timestamp. When set to `undefined` no timestamp is used.
+   */
+  time?: number | undefined
   /** A callback which will be called to notify the caller of any changes to the document */
   patchCallback?: PatchCallback<T>
 }
@@ -527,8 +530,8 @@ function _change<T>(
     state.handle.isolate(scope)
     heads = scope
   }
-  if (options.time === undefined) {
-    options.time = new Date().getTime()
+  if (!("time" in options)) {
+    options.time = Date.now()
   }
   try {
     state.heads = heads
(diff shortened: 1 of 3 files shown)


pvh:
Hey @marionauta, thanks for the PR! I had been planning to land this upstream in automerge-repo instead but maybe it's better to fix it here. Either way though, we should have a test which checks the times correctly round-trip, that you can specify a time, and that you can still specify no time!

This change may also break folks who rely on the old behaviour so we should be a bit thoughtful about how we release it.

pvh:
Just to be explicit, can you please add tests to make sure this works as expected and to avoid any future regressions? Should be very straightforward :)

marionauta:
@pvh Just added some tests.

I changed the API slightly to allow for no timestamps (as it was before):

- If time parameter is omitted, it defaults to the current timestamp.
- If time is set, it uses that timestamp.
- If time is `undefined` (`{ time: undefined }`) no timestamp will be stored.

In the tests I had to check for no timestamp with a 0, because that's what `decodeChange` does.

ept:
Automerge 0.x used [1 second resolution](https://github.com/automerge/automerge-classic/blob/main/frontend/index.js#L92) (not millisecond resolution) timestamps. Is this an intentional departure from 0.x? Personally I can't think of any use case that needs millisecond resolution, but I can think of problems:

- storing milliseconds significantly increases the file size (even second-resolution timestamps are already one of the bigger parts of the compressed text editing trace)
- millisecond resolution keystroke timings may leak personally identifiable information (allowing authors to be identified) as keystroke timings differ from person to person – in fact keystroke timings have been proposed as a biometric identifier

marionauta:
I asked on Discord and [was told](https://discord.com/channels/1200006940210757672/1200129501867298878/1265327923506712727) the timestamp should be in milliseconds.

pvh:
Thanks for the tests, @marionauta! I will defer to @alexjg on the specifics of timestamp resolution but it certainly wouldn't be difficult to round to the nearest 1000 per @ept's note. 

Also, I agree it would indeed be unfortunate to give up all the file size benefits we have earned through hard work in order to be able to leak our user's biometric identity!

alexjg:
Ah, my apologies @marionauta , I was actually thinking of the `Timestamp` value type - which is a milllisecond precision unix timestamp - not the commit timestamp. I agree with the arguments here that this should be a second precision.

marionauta:
I updated the functions, doc strings and tests to use seconds precision.

alexjg:
Release in v2.2.6

PR #974, opened 2024-09-14
Update flake and build instructions

kentookura:
Since the build assumes nightly rust builds (by using the -Z flag), the rust overlay now selects the latest nightly version.

The other changes in the flake:
-  rnix-lsp is unmaintained and thus removed
- pkgconfig has been renamed

The changes to `rust/automerge-c/Cargo.toml` are somehow generated by the build step, I did not manually change the version number.

The build instructions are updated to use ninja as that was the only thing that worked for me. If this is undesirable, we can indicate that ninja is an alternative way to build.

### DIFF
--- a/rust/automerge-c/README.md
+++ b/rust/automerge-c/README.md
@@ -11,7 +11,7 @@ a root directory of your choosing (e.g. "/usr/local") like so:
 
 ```shell
 cmake -E make_directory automerge-c/build
-cmake -S automerge-c -B automerge-c/build
+cmake -S automerge-c -B automerge-c/build -G Ninja
 cmake --build automerge-c/build
 cmake --install automerge-c/build --prefix "/usr/local"
 ```
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ To build this codebase you will need:
 - `yarn`
 - `cmake`
 - `cmocka`
+- `ninja`
 
 You will also need to install the following with `cargo install`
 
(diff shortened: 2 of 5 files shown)


rust/automerge-c/README.md
```diff
@@ -11,7 +11,7 @@ a root directory of your choosing (e.g. "/usr/local") like so:
 
 ```shell
 cmake -E make_directory automerge-c/build
-cmake -S automerge-c -B automerge-c/build
+cmake -S automerge-c -B automerge-c/build -G Ninja
```
alexjg, line 14: @jkankiewicz is this necessary?

kentookura: As I said above, I added this because this was the only way that it built for me. I will instead update the README to indicate that if the cmake command fails, one can try to use Ninja instead.

alexjg:
Updating in favor of #1070 as that doesn't touch the C stuff, which should be fixed separately I think.

jkankiewicz:
> @jkankiewicz is this necessary?

@alexjg , no it should not be necessary to specify Ninja as the underlying build system; it should be possible to build automerge-c with whatever underlying build system is CMake's default for the platform one is building automerge-c on. On a Linux distribution, for example, CMake's default underlying build system is usually Make.

@kentookura, which platform were you building automerge-c on that required you to specify `-G Ninja`?

kentookura:
This must have been on Linux, specifically NixOS.

PR #977, opened 2024-10-15
Prevents coupling UTF-16 text encoding to WASM targets

miguelangel-dev:
### Related PRs
   - https://github.com/automerge/automerge/pull/597
   - https://github.com/automerge/automerge/pull/641
   - https://github.com/automerge/automerge/pull/626

### Description
We've encountered several issues with the `Splice Text`, `Cursor`, and `Marks` operations in the automerge-swift project. The root cause is that in cross-platform projects, WASM systems are not guaranteed to use UTF-16 text encoding. For example, in [automerge-swift](https://github.com/automerge/automerge-swift) the default encoding is Unicode Scalar, which is also used in the public API. This leads to **text width** misalignments when sharing code with other WASM systems.

I guess UTF-16 was initially enforced in WASM because of its design with JavaScript in mind, but text encoding can vary depending on the platform. This PR relaxes that constraint by introducing a compile-time flag (`utf16-indexing`). By default, the WASM target will still be compiled with UTF-16 indexing to ensure backward compatibility with existing projects, but this can be overridden if clients need it.

side note: PR enables to change WASM Text encoder without changing the current behaviour.

### QA
The original set of tests continues to pass, ensuring backward compatibility with existing projects. [That test](https://github.com/automerge/automerge/blob/890e2eefbcc84365dcb49a4701a28980b5da5e2a/rust/automerge-wasm/test/test.mts#L2016) confirms that the original UTF-16 functionality remains intact.


heckj:
Hey @miguelangel-dev - if you use the utf8 binding in compilation, does that change its representation to Swift's string interface and break any of the text encoding/iteration there? Wasnt sure if this was limited to the WASM interactions, or if this compilation flag would have broader effects.

miguelangel-dev:
> if you use the utf8 binding in compilation, does that change its representation to Swift's string interface and break any of the text encoding/iteration there?
Exactly. On the surface, the bindings will operate within the defined encoder, but under the hood, independently of what it receives, it will be considered UTF8/UTF16/Unicode scalar depending on the compilation flag.  By default, it will use `Rust's char`, which represents a single Unicode scalar value. In the opposite, **wasm** due to the interaction with JavaScript bindings, it has been fixed to UTF-16, avoiding the need for domain transformations. However, since the public API for Swift bindings is using Unicode scalars, there’s a misalignment! In pure WASM environments, it will receive Unicode and treat it as UTF-16, causing issues with Cursors, Marks, and Splice Text operations.

This change would allow us to adjust the automerge-swift Cargo configuration. By setting `default-features = false and enabling features = ["wasm"]`, we can normalize the system to use Unicode scalars across all WASM environments.

side-note: not sure how robust is the current UTF-8/UTF-16 implementation, and if it brings some unexpected cases, but it could bring the option to make a migration to Graphemes clusters in the futures without altering the logic/structure of Automerge Core.

miguelangel-dev:
could someone force the bots?

rust/automerge/src/text_value.rs
```diff
@@ -1,124 +1,122 @@
...
-}
-
-#[cfg(feature = "utf8-indexing")]
-impl TextValue {
-    pub(crate) fn new(s: &str) -> Self {
-        let mut v = SequenceTree::new();
-        for ch in s.bytes() {
-            v.push(ch)
+cfg_if! {
+    if #[cfg(feature = "utf8-indexing")] {
```
miguelangel-dev, line 6: it is only used in `C` - [reference](https://github.com/automerge/automerge/pull/641)

scripts/ci/build-test
```diff
@@ -1,7 +1,10 @@
 #!/usr/bin/env bash
 set -eoux pipefail
 
-cd rust
-cargo build --workspace --features=optree-visualisation,wasm
```
miguelangel-dev, line 5: **wasm** will be executed into `./script/ci/wasm-tests`

rust/automerge/src/text_value.rs
```diff
@@ -1,124 +1,122 @@
...
-
-#[cfg(target_family = "wasm")]
-#[derive(Clone, PartialEq, Default)]
-pub struct TextValue(SequenceTree<u16>);
-
-#[cfg(feature = "utf8-indexing")]
-#[derive(Clone, PartialEq, Default)]
-pub struct TextValue(SequenceTree<u8>);
-
-#[cfg(not(any(target_family = "wasm", feature = "utf8-indexing")))]
```
alexjg, line 17: Switching from this style to using `cfg-if` doesn't seem worth it to me. We're taking on an extra dependency for a reasonably small increase in readability IMO. As I understand it the difference is really that we would go from `#[cfg(feature = <...>)]` to the nested if chains below right?

miguelangel-dev: 100% aligned with your comment. Basically, I do not want to introduce any if-else, or another dependency for solving that. 

It’s unclear whether this is a limitation of Rust or an intentional design choice, but in Rust, features are not mutually exclusive. This means you can’t easily express conditions like <if featureA and !featureB>. There's no direct syntax for writing this, aside from using:
- When there is a conflict, [choose one feature over another](https://doc.rust-lang.org/cargo/reference/features-examples.html#feature-precedence). The [cfg-if](https://crates.io/crates/cfg-if) package can help with writing more complex `cfg` expressions. **It is the recommended, and followed approach**
- Split the functionality into separate packages.
- Architect the code to allow the features to be enabled concurrently, and use runtime options to control which is used. For example, use a config file, command-line argument, or environment variable to choose which behavior to enable. **Not a valid option**

👉🏻 Reference from official Rust doc [[Mutually exclusive features]](https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features)

PR #984, opened 2024-11-14
add support for .at() in list proxies

### LINKED ISSUE #979: Add Array.at to document array
It would be great to get better parity with ecma script standard on arrays and add `.at` to arrays stored in a document.

Example:
```ts
const repo = new Repo()
const handle = repo.create({ items: ['item1'] })

handle.change(doc => {
  if (typeof doc.items.at(0) !== 'string') throw Error('invalid document')
})
```

mattheson:
closes #979


mattheson:
`yarn run packaging-tests` with clean main fails on webpack_esm_fullfat on my computer. possibly related to https://github.com/webpack/webpack/issues/13559? will investigate more

PR #994, opened 2024-12-19
new cursor API

mattheson:
closes https://github.com/automerge/automerge/issues/881

this pr implements two features:
1. start/end cursors which respectively correspond to `0` and `sequence.length` positions
2. ability to specify how the cursor behaves when its referenced item is removed (`MoveCursor`)


mattheson:
@alexjg there seems to be an issue with the `js_tests` runner for `yarn run packaging-tests`.

`yarn run packaging-tests` works on my pc

rust/automerge/src/automerge.rs
```diff
@@ -1506,18 +1519,95 @@ impl Automerge {
...
+                            Ok(found.index)
+                        } else {
+                            // FIXME: this should probably be an `OpSet` query
+                            // also this implementation is likely very inefficient
+
+                            // current implementation walks upwards through `key` of op pointed to by cursor
+                            // and checks if `key` is visible by using `seek_list_opid()`.
+
+                            let mut key = match found.op.key().elemid() {
+                                None => {
```
alexjg, line 1569: This check and the one a few lines down in the loop really indicate that we are not in a sequence CRDT. I.e. if we are in a sequence and the `elemid` of an op returns `None` then something has gone really wrong somewhere. On that basis I think we should just `expect` here but make sure to include a check at the beginning of this function that the object type we are examining is a sequence (`ObjTyp::is_sequence)`.

mattheson: fixed in 67a0dd7e33db142617b6ba5a5fae74ac872298fb

mattheson:
not sure what the `build_wasm` problem is... ci is being difficult with us

PR #997, opened 2025-01-07
rust: allow specifying text encoding at document creation

alexjg:
Problem: when users perform operations like "insert at index <x> in text <y>" Automerge has to transform the index into a position in the OpSet. Different platforms prefer different encodings for text, so we need to know what encoding to use to translate the index the user has given us into a position in the OpSet. Currently this configuration is selected by compile time feature flags, but in some cases users don't know what encoding the platform they are targeting uses until runtime (this is the case for Rust code which is compiled to WASM, but not always used in a browser).

Solution: Make the text encoding a parameter at document creation.

### DIFF
--- a/rust/automerge/tests/text_encoding.rs
+++ b/rust/automerge/tests/text_encoding.rs
@@ -1,0 +1,636 @@
+use std::borrow::Cow;
+
+// Test that looking up by index works across diferent ways of calculating
+// the index
+//
+// Places in the codebase where we use indexes for manipulating text:
+//
+// # Reading values
+//
+// - ReadDoc::length and ReadDoc::length_at
+// - ReadDoc::marks and ReadDoc::marks_at (via the Mark::start and Mark::end)
+//   fields
+// - ReadDoc::get_cursor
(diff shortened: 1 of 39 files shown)


rust/automerge/tests/text_encoding.rs
```diff
@@ -0,0 +1,625 @@
...
+            grapheme: 6,
+        },
+    }
+    .run();
+}
+
+#[test]
+fn splice_text() {
+    Scenario {
+        text: "hello 🫠 world",
```
miguelangel-dev, line 192: mmm... could you repeat the test using `🧑‍🧑‍🧒‍🧒` instead of that emoji?

**Insertion** is always based on `rust::chars` [**[ref](https://github.com/automerge/automerge/blob/a3185700b3400d9f00060ee1dcaeb8675cbab1db/rust/automerge/src/transaction/inner.rs#L601)**], which splice text based on Unicode scalars. Probably it is not expected, and it seems tests are passing because the picked emoji is 1:1 between Unicode-scalars and grapheme clusters.

PR #998, opened 2025-01-08
ci: fix js tests

mattheson:
<del>bumping puppeteer fixes current ci issues</del>

<del>bumping wrangler because of an odd ci failure. wrangler managed to launch but test hung for 5 minutes:</del>
https://github.com/mattheson/automerge/actions/runs/12664743678/job/35293364913


mattheson:
well... looks like that didn't fix it

PR #1015, opened 2025-04-15
New compressed OpSet

orionz:
This is a rewrite of the Automerge internals to store data internally in columnar compressed format. 

### Hexane 

The first component is Hexane.  This started off as a collection of structs to manage the columnar compressed data in a general purpose way.  I decided to spin it out into its own crate to help keep its API's clean and generic but also in the hopes that it may be useful for others in the future.

The two primary exports of the library are ColumnData and the trait ColumnCursor.   ColumnData stores and owned the compressed data and offers an interface to splice it or read it.

```rust
     let mut column = ColumnData::<UIntCursor>::new();
     column.splice(0,0,[1,2,3,4,5,6]);
     assert_eq!(column.iter_range(3..4).collect::<Vec<_>>(),vec![Some(Cow::Owned(4))]);
     assert_eq!(column.get(3), Some(Some(Cow::Owned(4))));
     let bytes = column.save();
     let other = ColumnData::<UIntCursor>::load(&bytes)l;
```

These are the basics.  There are some more advanced features for doing smart searches I will talk about later. 

The shortcomings of the library are mostly in API design.  I would really like the cursors to specify the exact format of the inputs and outputs.  Currently the API always assumes null's are possible and always assumes owned or borrowed return values are possible.  Thus the return value of get() or iter() is `Option<Cow<'_,Cursor::Item>>`.  Being able to spcify that the return value us just `u32` or `Option<&str>` would be wonderful someday.

Second - write performance could be better.  I always end up copying and replacing the internal slab even when twiddling a single byte in the compressed data would be possible.  But so far none of the benchmarks have pointed to this as a problem so it stays until what time it becomes a problem.

Third - the DeltaCursor code is quity messy.  I think that it could be cleaned up by making a smarter primitive `DeltaInt { abs: i64, delta: i64 }`.  Currently the primitive knows its absolute value or its delta but never both at the same time and it makes for quite a mess.

Lastly - there are some automerge based expectations built into the slab indexes (mainly that every slab keeps a min and max u64 value attached to it) where as a proper version of this library would have a cursor specific set of index data (min and max string?). None of that has been needed by automerge so it has not been built out.

Getting Hexane working and compliant with our columnar format was a lot of hard work and code but it has generally been a very reliable and robust foundation to build on.

### Reconstruct OpSet

todo!()

### Change Batch Loaded 

todo!()

### The new Ref Types 

todo()

### Doc iter()

todo!()


rust/hexane/src/cursor.rs
```diff
@@ -0,0 +1,590 @@
...
+        slab: &'a Slab,
+        encoder: &mut Encoder<'a, Self>,
+        post: Self::PostState<'a>,
+        cursor: Self,
+    ) -> Option<Self>;
+
+    // ENCODER
+    fn finish<'a>(slab: &'a Slab, writer: &mut SlabWriter<'a, Self::Item>, cursor: Self);
+
+    fn copy_between<'a>(
```
alexjg, line 238: I think we should use a struct for the arguments of this function as I think it would be easier to read. I suggest:

```
struct CopyBetweenArgs<'a, C: ColumnCursor> {
    slab: &'a [u8],
    writer: &mut SlabWriter<'a, C::Item>,
    start: C,
    end: C,
    last_run: Run<'a, C::Item>,
}
```

This would make it easier to not confuse the start and end arguments and also make it clearer what `last_run` is for (assuming I've understood it correctly).

rust/hexane/src/slab/writer.rs
```diff
@@ -0,0 +1,531 @@
...
+                }
+            }
+        }
+        debug_assert_eq!(width, buffer.len());
+        result.push(Slab::new(buffer, len, acc, abs));
+        assert_eq!(self.num_slabs, result.len());
+        result
+    }
+
+    pub fn copy(
```
alexjg, line 503: I think we should add a struct for these arguments, otherwise it will be very easy to confuse the `lit` and `size` arguments (I have confused them several times whilst reading this code base).

Something like:

```
struct CopyArgs{
    slab: &'a [u8],
    range: Range<usize>,
    lit_length: usize,
    total_items: usize,
    acc: Acc,
    bool_state: Option<bool>,
  }
  ```
  
  Actuall `bool_state` needs another name, I've already forgotten what it indicates.

alexjg (commented):
Here's an initial round of comments focusing on the slab tree and writer in `hexane`.

alexjg:
Closed in favor of #1023

PR #1026, opened 2025-05-08
js: fix the automerge.wasm.base64 export

nikgraf:
Note: I haven't tried it yet.

### DIFF
--- a/javascript/HACKING.md
+++ b/javascript/HACKING.md
@@ -134,7 +134,7 @@ Altogether then we have the following exports:
 - `/`: The full package with WebAssembly initialization
 - `/slim`: Only the JavaScript code, no WebAssembly initialization
 - `/automerge.wasm`: The WebAssembly file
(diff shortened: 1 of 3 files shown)


alexjg:
Based on my experiments this still needs a bit of work. I created a basic next.js project and then added `"@automerge/automerge": "<path to this project>/javascript"` to `package.json`, then did the `/slim` and `base64.wasm` import dance. This works fine in an app built using `next build` but fails to resolve the package in `next dev`. I don't know a lot about Next.js, does it's dev mode respect subpath exports?

nikgraf:
@alexjg this is a simple Next.js example where dev and build version work https://github.com/nikgraf/automerge-nextjs-example

Note: There are several improvements/variations I plan to in the future, but initial step would be to get rid of this type error https://github.com/nikgraf/automerge-nextjs-example/blob/main/src/components/test.tsx#L8-L9

alexjg:
The test failures here are flakes I think, I'm not rerunning yet as I want @orionz to have a chance to look at them first.

PR #1032, opened 2025-05-12
rust: Make a slightly less flaky test for a marks bug

alexjg:
I've noticed a flaky test in `op_set2::change::batch::tests::batch_marks_patch`. I haven't managed to nail down exactly what the root cause is (though it looks like the fast and slow seek op queries have inconsistent behavior around skipping mark operations) but I have pulled out the randomly generated ops in the original test and put them in a test in this PR. The test still flakes around 50% of the time, but that's much more consistent than the ~1/8 (not rigorously tested) failures in the original test.


alexjg:
Closing as the fix for this is now on `main`

PR #1036, opened 2025-05-15
optimize doc.len

orionz:
Automerge.length() has been optimized.  For a n=100,000 on my machine Automerge.length() went from 20ms -> 0.5ms on HEAD and 3ms at(heads)

### DIFF
--- a/rust/automerge/Cargo.toml
+++ b/rust/automerge/Cargo.toml
@@ -15,8 +15,8 @@ utf8-indexing = []
 utf16-indexing = []
 
 [dependencies]
-#hexane = { path = "../hexane" }
-hexane = "0.1.1"
+hexane = { path = "../hexane" }
+#hexane = "0.1.1"
 hex = "^0.4.3"
 leb128 = "^0.2.5"
 sha2 = "0.11.0-pre.5"
--- a/rust/automerge/Cargo.toml
+++ b/rust/automerge/Cargo.toml
@@ -15,8 +15,7 @@ utf8-indexing = []
 utf16-indexing = []
 
 [dependencies]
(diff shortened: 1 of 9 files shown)


rust/automerge/Cargo.toml
```diff
@@ -15,8 +15,8 @@ utf8-indexing = []
 utf16-indexing = []
 
 [dependencies]
-#hexane = { path = "../hexane" }
-hexane = "0.1.1"
+hexane = { path = "../hexane" }
```
alexjg, line 18: You can put the version in here as well, it'll just choose the local version if available, then fall back to crates.io.

rust/automerge/src/op_set2/change/batch.rs
```diff
@@ -1192,7 +1192,9 @@ mod tests {
 
     fn make_rng() -> SmallRng {
         let seed = rand::random::<u64>();
-        //let seed = 3727965917273105553;
```
alexjg, line 1195: I have this in a branch I'm using:

```
let seed = std::env::var("AUTOMERGE_TEST_SEED")
    .ok()
    .and_then(|s| s.parse::<u64>().ok())
    .unwrap_or_else(rand::random::<u64>();
```

rust/automerge/src/op_set2/op_set/mark_index.rs
```diff
@@ -182,6 +211,7 @@ impl MarkIndexColumn {
     }
 }
 
+/*
 impl From<i64> for MarkIndexValue {
```
alexjg, line 215: Can we remove this commented code?

PR #1040, opened 2025-05-18
js: patch wasm-bindgen code with vite-ignore

mattheson:
for https://github.com/automerge/automerge/issues/1037

### DIFF
--- a/javascript/scripts/build.mjs
+++ b/javascript/scripts/build.mjs
@@ -93,8 +93,9 @@ function buildWasm(outputDir) {
    * "src/wasm_bindgen_output/bundler/"
    *
    * @param {string} target
+   * @param {string[]=} vitePatchFiles
    */
-  function runWasmBindgen(target) {
+  function runWasmBindgen(target, vitePatchFiles) {
     console.log(`running wasm-bindgen for '${target}' target`)
     const outputPath = path.join(outputDir, target)
     fs.mkdirSync(outputPath, { recursive: true })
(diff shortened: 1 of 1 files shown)


javascript/scripts/build.mjs
```diff
@@ -93,8 +93,9 @@ function buildWasm(outputDir) {
    * "src/wasm_bindgen_output/bundler/"
    *
    * @param {string} target
+   * @param {string[]=} vitePatchFiles
    */
-  function runWasmBindgen(target) {
+  function runWasmBindgen(target, vitePatchFiles) {
```
alexjg, line 98: Rather than passing a parameter here I think we should just do the vite patching if `target == "web"`. My reasoning is that the only place this is a problem is in the `web` target because the init method generated by wasm-bindgen for the `web` target tries to guess where the wasm file is and do a dynamic import for it whereas all the other targets just hard code the location. Thus I think it's clearer that there's something about the web target we are interested in.

javascript/scripts/build.mjs
```diff
@@ -104,10 +105,28 @@ function buildWasm(outputDir) {
         cwd: __dirname,
       },
     )
+
+    // add /* @vite-ignore */ comment to URL imports for wasm
+    // prevents vite from bundling multiple wasm files
+    // see https://github.com/automerge/automerge/issues/1037
```
alexjg, line 111: As well as referencing the issue I think it would be good to write in some prose about the silly things vite is doing here which makes it necessary, as well as the fact that the vite comment is an undocumented thing we are taking advantage of.

It's good to reference issues, but you never know when GitHub will be down.

alexjg:
I sent a PR to your fork with the packaging test in it [here](https://github.com/mattheson/automerge/pull/4). If you could merge that and then squash everything into one commit I'll go ahead and merge.

Issue #951, opened 2024-07-19
Unrelated string changes in objects garble string changes through `changeAt()`


cindyhu2023:
Hi!

I'm seeing unexpected string merging results when I use `changeAt()` to updated a string property in a nested object. More specifically, when I have 2 objects, certain size of regular updates (through `change()`) to the string property on one object affect the `changeAt()` results on the other object. For context, I'm trying to use `changeAt` to achieve "undo" on some previous changes while preserving the changes afterwards. 

Here is the code snippet to reproduce:

```typescript

describe('Automerge undo', () => {
  it('can undo string property in nested objects with changeAt', () => {
    let doc = Automerge.init('00')

    doc = Automerge.change(doc, (d: any) => {
      d.object1 = {color: 'unset'}
      d.object2 = {name: ''}
    })

    doc = Automerge.change(doc, (d: any) => {
      d.object1.color = 'red'
    })
    const afterRed = Automerge.getHeads(doc)

    doc = Automerge.change(doc, (d: any) => {
      d.object1.color = 'white'
    })
    const afterWhite = Automerge.getHeads(doc)

    // change object1 color again
    doc = Automerge.change(doc, (d: any) => {
      d.object1.color = 'blue'
    })
    const afterBlue = Automerge.getHeads(doc)
    assert.deepEqual(doc.object1.color, 'blue')

    // update string property on unrelated object
    doc = Automerge.change(doc, (d: any) => {
      d.object2.name = 'a'.repeat(31)  // 31 is the breaking point
    })

    // revert blue to white
    doc = Automerge.changeAt(doc, afterBlue, (d: any) => {
      d.object1.color = 'white'
    }).newDoc
    assert.deepEqual(doc.object1.color, 'white')

    // revert white to red
    doc = Automerge.changeAt(doc, afterWhite, (d: any) => {
      d.object1.color = 'red'
    }).newDoc
    assert.deepEqual(doc.object1.color, 'red')

    // revert red to unset
    doc = Automerge.changeAt(doc, afterRed, (d: any) => {
      d.object1.color = 'unset'
    }).newDoc
    assert.deepEqual(doc.object1.color, 'unset')
  })
})

```

Test case failed with following error:
```
  Automerge
    1) can undo string property in nested objects with changeAt


  0 passing (9ms)
  1 failing

  1) Automerge
       can undo string property in nested objects with changeAt:

      AssertionError: expected 'unsetredte' to deeply equal 'unset'
      + expected - actual

      -unsetredte
[...]

PR #1073, opened 2025-05-21
rust: expose a weird patch bug

alexjg:
This PR adds a test which exposes the core of the issue in #951 

I still don't know what the root cause is, but this is a more minimal reproduction. @orionz any ideas?


alexjg:
Fixed in #1074

PR #1097, opened 2025-06-02
crate change_graph all at once, faster build paths, fast incLoad

orionz:
Several small performance imnprovements on load

### DIFF
--- a/rust/automerge/src/autocommit.rs
+++ b/rust/automerge/src/autocommit.rs
@@ -184,9 +184,12 @@ impl AutoCommit {
     /// [`Self::reset_diff_cursor()`]
     pub fn update_diff_cursor(&mut self) {
         self.ensure_transaction_closed();
-        self.patch_log.set_active(true);
-        self.patch_log.truncate();
-        self.diff_cursor = self.doc.get_heads();
+        let heads = self.doc.get_heads();
+        if !heads.is_empty() {
+            self.patch_log.set_active(true);
+            self.patch_log.truncate();
+            self.diff_cursor = heads;
+        }
     }
 
     /// Returns the cursor set by [`Self::update_diff_cursor()`]
--- a/rust/automerge/src/automerge.rs
+++ b/rust/automerge/src/automerge.rs
@@ -1641,8 +1641,18 @@ impl Automerge {
             if !visible_objs.contains(&obj.id) {
                 continue;
             }
-            for op in ops.visible(at.clone()).top_ops() {
-                //if op.visible_at(at.as_ref()) {
+            let mut index = 0;
+            let encoding = self.text_rep(obj.typ);
+            //for op in ops.visible(at.clone()).top_ops() {
+            let mut iter = ops.visible2(self.ops(), at.as_ref()).peekable();
+            while let Some(op) = iter.next() {
+                // cheap top ops
+                if let Some(p) = iter.peek() {
+                    if p.elemid_or_key() == op.elemid_or_key() {
+                        continue;
+                    }
+                }
+                let next_width = op.width(encoding);
                 if let OpType::Make(_) = op.op_type() {
                     visible_objs.insert(op.id.into());
                     let (mut path, parent_obj_id) = if obj.id.is_root() {
@@ -1653,19 +1663,13 @@ impl Automerge {
                     };
                     let prop = match op.key {
                         KeyRef::Map(prop) => Prop::Map(prop.into()),
-                        KeyRef::Seq(_) => {
-                            let encoding = self.text_rep(obj.typ);
-                            let found = self
-                                .ops
-                                .seek_list_opid(&obj.id, op.id, encoding, at.as_ref())
-                                .unwrap();
-                            Prop::Seq(found.index)
-                        }
+                        KeyRef::Seq(_) => Prop::Seq(index),
                     };
                     path.push((parent_obj_id.clone(), prop));
                     let obj_id = self.ops.id_to_exid(op.id);
                     paths.insert(obj_id, path);
                 }
+                index += next_width;
             }
         }
         paths
--- a/rust/automerge/src/op_set2/change.rs
+++ b/rust/automerge/src/op_set2/change.rs
@@ -5,7 +5,7 @@ use super::types::{ActionCursor, ActorCursor, ActorIdx};
 use crate::change_graph::ChangeGraph;
 use crate::storage::change::{ChangeOpsColumns as ChangeOpsColumns2, Verified};
 use crate::storage::{Change, ChunkType, Header};
-use crate::types::ActorId;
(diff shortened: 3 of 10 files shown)


rust/automerge/src/autocommit.rs
```diff
@@ -184,9 +184,12 @@ impl AutoCommit {
     /// [`Self::reset_diff_cursor()`]
     pub fn update_diff_cursor(&mut self) {
         self.ensure_transaction_closed();
-        self.patch_log.set_active(true);
-        self.patch_log.truncate();
-        self.diff_cursor = self.doc.get_heads();
+        let heads = self.doc.get_heads();
+        if !heads.is_empty() {
```
alexjg, line 188: What does this change achieve? It seems like a big change in semantics (if you have empty heads you don't make the patch log active?)

rust/automerge/src/automerge.rs
```diff
@@ -1641,8 +1641,18 @@ impl Automerge {
             if !visible_objs.contains(&obj.id) {
                 continue;
             }
-            for op in ops.visible(at.clone()).top_ops() {
-                //if op.visible_at(at.as_ref()) {
+            let mut index = 0;
+            let encoding = self.text_rep(obj.typ);
+            //for op in ops.visible(at.clone()).top_ops() {
+            let mut iter = ops.visible2(self.ops(), at.as_ref()).peekable();
```
alexjg, line 1647: Not related to this PR, but just a note that we should really find a better name than `visible2` :slightly_smiling_face:

rust/automerge/src/op_set2/change/collector.rs
```diff
@@ -387,10 +391,12 @@ impl<'a> ChangeCollector<'a> {
         let mut seq = vec![0; num_actors];
         let mut changes = Vec::with_capacity(self.changes.len());
         let mut heads = BTreeSet::new();
-        let mut change_graph =
-            ChangeGraph::with_capacity(self.changes.len(), self.num_deps, num_actors);
+        //let mut change_graph =
```
alexjg: nit: can we delete these comments?

rust/automerge/src/op_set2/change/collector.rs
```diff
@@ -432,13 +438,22 @@ impl<'a> ChangeCollector<'a> {
 
             let change = Change::new(finished);
 
-            change_graph.add_change(&change, actor)?;
+            //change_graph.add_change(&change, actor)?;
```
alexjg: nit: remove commented code

rust/automerge/src/op_set2/change/collector.rs
```diff
@@ -432,13 +438,22 @@ impl<'a> ChangeCollector<'a> {
...
 
         let max_op = self.max_op;
 
+        let change_graph = ChangeGraph::from_iter(
+            changes.iter().zip(actors.into_iter()),
+            self.num_deps,
+            num_actors,
+        )?;
+
+        //assert_eq!(change_graph, change_graph2);
```
alexjg: nit: remove commented code

rust/automerge/src/automerge.rs
orionz: Agreed - my intent was to go back and replace all the instances of `visible()` - ill make sure to include that in the list_opt patch im working on

rust/automerge/src/autocommit.rs
orionz: So there's a pretty meaningful optimization where AutoCommit does not bother collecting patch log data until your first call to `diff_incremental()` simply because `current_document()` is so much faster than `diff()`.  However this optimization was never being used because AutomergeRepo was creating the document and then materializing it (just a `{}`) before doing the `loadIncrental()` - which is a pretty normal way to go about things.  Given that this small optimization was being missed on the most meaningful `diff_incremental()` of them all - the one on inital load - I decdied to detect this and basically ignore a diff_incremental() if the document is empty.  Also we were producing patches of things being created then deleted etc - when on document load you really just want the final value.

orionz:
Looked at the problematic essay file - some of the fixes in the PR helped a great deal.  Also found some really odd memory thrashing in the actor remapper code.  Added a patch to fix that.

Here's a bench on main:
```
         :loadIncremental:  13763ms
         ::save: 109ms
         ::loadIncremental again: 7541ms
         ::just load: 2177ms
```

Here's the same bench on this branch
```
         :loadIncremental:  2377ms
         ::save: 107ms
         ::loadIncremental again: 1698ms
         ::just load: 1240ms
```

rust/automerge/src/op_set2/change.rs
```diff
@@ -172,20 +189,20 @@ impl Eq for OpBuilder<'_> {}
...
 {
     if ops.is_empty() {
-        return (ChangeOpsColumns::default().into(), vec![]);
+        return ChangeOpsColumns::default().into();
     }
 
-    let (actor_map, actors) = remap_actors(ops, meta, actors);
+    mapper.remap_actors(ops, meta);
 
     let _remap = move |actor: Option<Cow<'_, ActorIdx>>| {
```
alexjg: Can we remove the leading underscore, it suggests that the variable isn't used later in the function when in fact it is.

rust/automerge/src/op_set2/change.rs
```diff
@@ -41,16 +41,35 @@ pub(crate) fn build_change<T, G>(
...
 ) -> Change<'static, Verified>
+where
+    T: AsChangeOp,
+    G: GetHash,
+{
+    let mut mapper = ActorMapper::new(actors);
+    build_change_inner(ops, meta, graph, &mut mapper)
+}
+
+#[inline(never)]
```
alexjg: Does the `inline(never)` make a big difference?

rust/automerge/src/op_set2/change.rs
```diff
@@ -41,16 +41,35 @@ pub(crate) fn build_change<T, G>(
     graph: &G,
     actors: &[ActorId],
 ) -> Change<'static, Verified>
+where
+    T: AsChangeOp,
+    G: GetHash,
+{
```
alexjg, line 46: Might be worth a comment that we are doing this so we can reuse the memory for the actor map as it made a big difference to performance. Otherwise 6 months down the road we'll "simplify" this away.

rust/automerge/src/op_set2/change.rs
orionz: The inline never gives me flamegraphs with callstacks that are useful.  They dont seem to affect performance at all so I've not minded leaving them in but I suppose good hygene might be stripping them out after a performance session and then adding them back in next time Im hunting something down...
Thoughts?

alexjg: Yeah I would be nervous about leaving them in until we have benchmarks running in CI on all major architectures.

Issue #574, opened 2023-04-19
React Native?


andrioid:
Is there any way of running this on platforms that don't support WASM yet?

This library would be a perfect fit for offline React Native applications and LLVM for persistence.

alexjg:
I've laid out some of my thoughts on this [here](https://github.com/automerge/automerge/issues/525#issuecomment-1465168972)

I would be extremely happy to see contributions related to this :smiley:

andrioid:
https://github.com/cawfree/react-native-webassembly looks promising, and I hope I'll have time to try it out there.

pwwolf:
Given this doesn't work in RN without presumably some hacking, is there any reason not to use the pure javascript implementation in RN? I was able to run automerge@1.0.1-preview.7 fine, but I was concerned there might be binary compatibility differences between that and another client running the wasm version.

Tymek:
Maybe I'll look into it. It's an interesting challenge. What I found so far:

### `window.WebAssembly` does not live in native land
Automerge Slim Next is using this global in `./wasm_bindgen_output/web/automerge_wasm.js`.
1. I doubt using above mentioned `react-native-webassembly` as global will just work. It's still worth a try.

There are other paths thru this integration hell. Options I found are:

2. Native module
https://github.com/dgca/expo-rust-demo/tree/main
Rust ➡️ Java+Swift+WASM ➡️ Native module
Most stable, but needs to be written 3 times, separately for each language.

3. [Turbo module](https://github.com/reactwg/react-native-new-architecture/blob/9845e0e0d709732a07c001684a7b11aa2e6eb1d5/docs/turbo-modules.md)
"Generate" native module from C++. Rust ➡️ Cpp ➡️ Native. Looks a bit round-about 🤷🏻‍♂️ 

4. [JSI module](https://blog.notesnook.com/getting-started-react-native-jsi/)
Write-once, but more experimental. See https://github.com/laptou/jsi-rs/tree/main

Similar issues where solved here: https://github.com/Nozbe/WatermelonDB/tree/master/native

pvinis:
any of those would be great as a first version for react native/expo. get people hooked, then make a better v2 😅

PR #1099, opened 2025-06-03
Node-API binding

kraenhansen:
I facilitated a discussion with @alexjg and @nikgraf at Local First Conf '25 and we believe adding a Node-API binding for Automerge will be valuable for a couple of reasons:

- Servers running Automerge via Node.js will be able to exceed the 4GB memory limit imposed by WASM.
- Less indirection (no crossing into JavaScript and interpreting of memory buffers) as the JavaScript engine is manipulated from native code.
- As [Hermes gets support for Node-API](https://github.com/facebook/hermes/pull/1377) and React Native eventually does too, we expect this work to serve the basis for fixing #574.

I intent for this PR and the `node-api-binding` branch it targets, to be long-living and the base for future PRs gradually adding a Node-API binding.

I suggest using [napi.rs](https://napi.rs/) for code generation from Rust macros, of:
- Node-API function calls
- TypeScript declarations (which is not the case for wasm-bindgen)

My working theory is that it will be easiest to port [the existing WASM binding](https://github.com/automerge/automerge/tree/main/rust/automerge-wasm/src).

Although ["WebAssembly" support is a declared goal of the napi.rs project](https://napi.rs/blog/announce-v2.en#whats-the-next-step) that's not "in scope" of this project and I don't expect the outcome of this project to replace the current wasm-binding.


alexjg:
This is great to see, thanks for the work! 

One thing that immediately jumps out at me is how much code will need to be duplicated between here and the wasm bindings. We should probably think about finding a way to share common logic between the two codebases to make it easier to maintain. That's not a reason to block this work though - let's get a thing that works first.

A smaller note which might make your life easier. In the next major release of Automerge - which is what i think we should support here - we do not support the `TextRepresentation::Array` mode. Anywhere you see that in the wasm codebase you can ignore it and just use the `TextRepresentation::String` branch. I will be removing that code before we release 3.0 anyway,

PR #1111, opened 2025-07-07
List opt

orionz:
Optimizations for seeking to an index in a list.   

The list test battery is seeking improvements on the order of 15 ms -> 30 µs 

Also optimized `seek_list_opid` and `find_op_by_id_and_vis` to similar effect.

Open work: 
* The new `advance_acc_by` method is faster, better testes and less confusing than the old `with_acc().nth()` interface.   Remove `with_acc()` and update code that uses it.

### DIFF
--- a/rust/automerge/src/op_set2/change/batch.rs
+++ b/rust/automerge/src/op_set2/change/batch.rs
@@ -30,13 +30,23 @@ struct BatchApply {
 }
 
 struct Untangler<'a> {
-    gosub: SmallHashMap<usize, Vec<usize>>,
-    stack: Vec<usize>,
+    // the tangle of change ops we need to navigate
+    change_ops: &'a mut [ChangeOp],
+    // these are entry points into the change_op tangle
+    // when we see a doc_op.id equal to key, put this vec onto the stack
+    // ops inserted after HEAD are put onto the stack immidately
     entry: SmallHashMap<OpId, Vec<usize>>,
+    // same concept as entry but internal to the change_ops array
+    gosub: SmallHashMap<usize, Vec<usize>>,
+    // stack of change ops ready to be processed
+    stack: Vec<usize>,
+    // these are change ops updating a pre-existing doc op elemid's
+    // and are handled differently than inserts
     updates: SmallHashMap<ElemId, Vec<usize>>,
     updates_stack: Vec<usize>,
-    change_ops: &'a mut [ChangeOp],
     pred: &'a mut PredCache,
+    // Top and Conflicts keep track of changes that need to
+    // be made to the index.top and index.visible columns
     top: Top,
     conflicts: &'a mut Vec<Adjust>,
     value: ValueState<'a>,
--- a/rust/automerge/src/op_set2/op_set.rs
+++ b/rust/automerge/src/op_set2/op_set.rs
@@ -20,6 +20,7 @@ use super::types::{Action, ActorCursor, ActorIdx, KeyRef, MarkData, OpType, Scal
 
 use std::borrow::Cow;
 use std::cmp::Ordering;
+use std::num::NonZeroUsize;
 use std::ops::{Range, RangeBounds};
 use std::sync::Arc;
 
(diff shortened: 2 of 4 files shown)


rust/automerge/src/op_set2/change/batch.rs
```diff
@@ -35,6 +35,10 @@ struct Untangler<'a> {
     entry: SmallHashMap<OpId, Vec<usize>>,
     updates: SmallHashMap<ElemId, Vec<usize>>,
     updates_stack: Vec<usize>,
+    change_ops: &'a mut [ChangeOp],
```
alexjg: I feel like this structure has become complex enough that it would be worth having some prose comments on it explaining how it works. Each time I have to review it I spend a really long time re-reading to understand what the different pieces of state here do.

rust/automerge/src/op_set2/op_set.rs
```diff
@@ -332,37 +428,62 @@ impl OpSet {
         query.resolve(index - 1).ok()
     }
 
+    pub(crate) fn query_insert_at_list(
+        &self,
+        obj: &ObjId,
+        index: usize,
```
alexjg: Might be worth using `NonZeroUsize` here to avoid panicking in the `advance_by` call when in debug mode, and just generally to make the contract of this function clear.

rust/automerge/src/op_set2/op_set.rs
```diff
@@ -332,37 +428,62 @@ impl OpSet {
...
+        encoding: ListEncoding,
+        _clock: Option<&Clock>,
```
alexjg: Can we remove this argument?

rust/automerge/src/op_set2/op_set.rs
```diff
@@ -433,55 +562,112 @@ impl OpSet {
...
+        let mut iter = self.cols.insert.iter_range(pos..range.end);
+        let acc = iter.calculate_acc().as_usize();
+        let insert_op = iter.next().flatten().as_deref().copied().unwrap_or(false);
+
+        if insert_op {
+            iter.advance_acc_by(0);
+            Some(pos..(iter.pos()))
+        } else {
+            let mut iter = self.cols.insert.iter_range(0..range.end);
+            iter.advance_acc_by(acc - 1);
```
alexjg, line 579: Can `acc` ever be `0`? If so we should guard against that here or use `saturating_sub`.

rust/automerge/src/op_set2/op_set.rs
```diff
@@ -433,55 +562,112 @@ impl OpSet {
...
+        assert_eq!(range.end, end_pos);
         OpsFound {
             index,
             ops: vec![],
             end_pos,
+            range,
+        }
+    }
+
+    fn list_register_at_pos(&self, pos: usize, range: Range<usize>) -> Option<Range<usize>> {
```
alexjg: This doesn't need to return `Option` right?

rust/automerge/src/op_set2/op_set/index.rs
```diff
@@ -160,12 +158,12 @@ impl IndexBuilder {
         self.flush();
 
         /*
-                let text = self
-                    .widths
-                    .iter()
-                    .zip(self.top.iter())
-                    .map(|(w, t)| if *t { Some(*w) } else { None })
-                    .collect();
+        let text = self
```
alexjg: nit: can we remove this commented code?

rust/hexane/src/columndata.rs
```diff
@@ -475,11 +489,64 @@ impl<'a, C: ColumnCursor> ColumnDataIter<'a, C> {
         Some(new_acc - starting_acc)
     }
 
+    fn reset_iter_to_acc2(&mut self, acc: Acc) -> Acc {
```
alexjg: Can this function replace `reset_iter_to_acc` or is there still work to do? I.e. is the double reset still an issue?

PR #1112, opened 2025-07-07
js: Add applyPatches

alexjg:


acurrieclark:
In my implementation of this, I figured the best API was to make `applyPatches` a function which worked **inside** a change function, rather than one which **contained** one. That way it could be surrounded by other logic and a number of changes could be grouped together if required.

It also felt that it would be more in keeping with the other functions in the toolkit which are designed the same way.

alexjg:
Good shout, updated.

chee:
hmm, i don't know what this looked like before so i don't know if i'm directly contradicting, but: what do you think of a signature that takes a single patch

```ts
for (const patch of payload.patches) {
  applyPatch(doc, patch)
}
```

so that a developer might intercept and reroute specific patches if they wish?

alexjg:
I've added `applyPatch`.

PR #1114, opened 2025-07-10
js: Fix applyPatches

alexjg:


sergeuz:
Is it by design in this PR that `applyDelPatch()` doesn't support deleting a key from a map? I think I saw it mentioned somewhere in the docs that `del` patches are supposed to support it.

PR #1120, opened 2025-07-15
Alternative noop fix

alexjg:
I found the naming and control flow of the `check_for_noop` function in #1117 a bit difficult to follow for the following reasons:

* The "check" sounds like this is a non-mutating function. But in fact it modifies both the query and the action parameter
* The returned boolean is actually the opposite of what I expected. I expected the op to be a noop if `check_for_noop` returns `true`, but in fact `check_for_noop` returns `true` if we do actually need to emit a delete operation for a conflicted value. I.e. it is _not_ a noop.
* The logic in this function really depends on this logic in the caller:

  ```rust
          let noop = query.check_for_noop(&mut action);
          if query.ops.is_empty() && action == OpType::Delete {
              return Ok(None);
          }
  ```
  To understand the noop check you need to know that the caller is going to do this next, otherwise the logic of popping the last element from `query.ops` seems like it will always emit a `Delete` operation - even for non conflicted writes.
  
I think we have a terminology problem. An operation which clears a conflicted value is not a noop - it's just not visible in the current materialized view of the document. This PR introduces a `ResolvedAction` enum to distinguish between the case of an op which updates the materialized view, and one which just resolves conflicts, then we pass that to the `TxOp::{list, map}` constructors. This also avoids passing around the extra `noop` argument to those functions.

### DIFF
--- a/rust/automerge/src/op_set2/op.rs
+++ b/rust/automerge/src/op_set2/op.rs
@@ -1,5 +1,5 @@
 use super::hexane::{ColumnDataIter, DeltaCursor, IntCursor};
-use super::op_set::{MarkIndexBuilder, ObjInfo, OpSet};
+use super::op_set::{MarkIndexBuilder, ObjInfo, OpSet, ResolvedAction};
 use super::types::{
     Action, ActorCursor, ActorIdx, KeyRef, MarkData, OpType, PropRef, PropRef2, ScalarValue,
 };
@@ -169,6 +169,7 @@ pub(crate) struct TxOp {
     pub(crate) obj_type: ObjType,
     pub(crate) index: usize,
     pub(crate) pos: usize,
+    pub(crate) noop: bool,
     pub(crate) bld: OpBuilder<'static>,
 }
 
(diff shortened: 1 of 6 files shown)


rust/automerge/src/op_set2/op_set.rs
```diff
@@ -1177,24 +1177,72 @@ impl OpsFound<'_> {
...
+pub(crate) enum ResolvedAction {
+    // An operation which resolves a conflict but does not change the observed state
+    // I.e. it is invisible to the materialized view
+    ConflictResolution(types::OpType),
+    // A normal operation which is visible in the document
+    VisibleUpdate(types::OpType),
+}
+
+impl ResolvedAction {
+    pub(crate) fn action(&self) -> types::OpType {
```
orionz: This function is only used to determine if the action is an increment... how about just replace it with `is_increment()` so we have one less clone()... OpType can contain strings so clone here is more than a copy.

rust/automerge/src/op_set2/op.rs
```diff
@@ -262,12 +262,15 @@ impl TxOp {
...
         pos: usize,
         index: usize,
-        _action: types::OpType,
+        action: ResolvedAction,
         elemid: ElemId,
         pred: Vec<OpId>,
-        noop: bool,
     ) -> Self {
-        let (action, value, expand, mark_name) = _action.clone().decompose();
+        let (action, noop) = match action {
```
orionz: how about `let (action, value, expand, mark_name, noop) = action.decompose();` and merge these 5 lines into 1 since thats all thats happening

orionz: Also it looks like the `clone()` before decompose is not needed.  Great to remove that while where in here.

rust/automerge/src/op_set2/op_set.rs
```diff
@@ -1177,24 +1177,72 @@ impl OpsFound<'_> {
...
+    ///
+    /// `Some(ResolvedAction)` if there is an op which needs to be inserted into
+    /// the opset, or `None` otherwise
+    pub(crate) fn resolve_action(
+        &mut self,
+        original_action: types::OpType,
+    ) -> Option<ResolvedAction> {
         if let Some(op) = self.ops.last() {
-            if let types::OpType::Put(v) = action {
+            if let types::OpType::Put(v) = &original_action {
```
orionz, line 1196: tiny nit - but I think you can `let Put(v) = ... {   } else { Some(ResolvedAction::VisibleUpdate(original_action))) }` and now all your if branches are covered and all the returns can go away.

PR #1148, opened 2025-08-05
Fix issue with TypeScript typings

### LINKED ISSUE #1147: TypeScript error in 3.1.1
Hi!

We're seeing the following TypeScript error when building with Automerge 3.1.1, across a variety of TypeScript versions:

```
wasm_types.d.ts:332:2 - error TS1036: Statements are not allowed in ambient contexts.
```

It looks like it's due to this section in that `d.ts` file:

```
export interface Stats {
  numChanges: number;
  numOps: number;
  numActors: number;
  cargoPackageName: string;
  cargoPackageVersion: string;
  rustcVersion: string;
};
```

The semi-colon on that final line is the culprit - I believe removing it will clear this issue. I tried to find where this file is generated in the code, in order to prepare a PR, but didn't have any luck. I'll take a closer look later if it's helpful?

cgauld:
fixes #1147

### DIFF
--- a/rust/automerge-wasm/src/lib.rs
+++ b/rust/automerge-wasm/src/lib.rs
@@ -376,7 +376,7 @@ export interface Stats {
   cargoPackageName: string;
   cargoPackageVersion: string;
   rustcVersion: string;
-};
+}
 
 export type UpdateSpansConfig = {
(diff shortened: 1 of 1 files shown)


cgauld:
My pleasure :-)

PR #1151, opened 2025-08-15
Add a new bundle type...

orionz:
Add new bundle type.  Use `document.bundle(hashes)`.  Bundle has `to_changes()` to conver it into `Vec<Change>`.  Can be loaded but it just internally calles `to_changes()`.  I have not written an optimized bundle loader.  Put a placeholder in for Author in.

Needs feedback ...

### DIFF
--- a/rust/automerge/src/automerge.rs
+++ b/rust/automerge/src/automerge.rs
@@ -20,7 +20,7 @@ use crate::exid::ExId;
 use crate::iter::{DocIter, Keys, ListRange, MapRange, Spans, Values};
 use crate::marks::{Mark, MarkAccumulator, MarkSet};
 use crate::patches::{Patch, PatchLog, TextRepresentation};
-use crate::storage::{self, change, load, CompressConfig, Document, VerificationMode};
+use crate::storage::{self, change, load, Bundle, CompressConfig, Document, VerificationMode};
 use crate::transaction::{
     self, CommitOptions, Failure, Success, Transactable, Transaction, TransactionArgs,
 };
@@ -727,6 +727,9 @@ impl Automerge {
                 );
                 Self::new()
             }
+            storage::Chunk::Bundle(_b) => {
+                todo!()
+            }
             storage::Chunk::CompressedChange(stored_change, compressed) => {
                 tracing::trace!("first chunk is compressed change");
                 change = Some(
@@ -904,6 +907,13 @@ impl Automerge {
         tracing::trace!(changes=?changes.iter().map(|c| c.hash()).collect::<Vec<_>>(), "merging new changes");
         self.apply_changes_log_patches(changes, patch_log)?;
         Ok(self.get_heads())
+    }
+
+    pub fn bundle<I>(&self, hashes: I) -> Result<Bundle, AutomergeError>
+    where
+        I: IntoIterator<Item = ChangeHash>,
+    {
+        Bundle::for_hashes(&self.ops, &self.change_graph, hashes)
     }
 
     /// Save the entirety of this document in a compact form.
--- a/rust/automerge/src/op_set2/change.rs
+++ b/rust/automerge/src/op_set2/change.rs
@@ -14,8 +14,10 @@ use std::num::NonZero;
 use std::ops::Range;
 
 pub(crate) mod batch;
+//pub(crate) mod bundle;
 pub(crate) mod collector;
 
+//pub(crate) use bundle::{Bundle, BundleMetadata};
 pub(crate) use collector::{BuildChangeMetadata, ChangeCollector, CollectedChanges};
 
 pub(crate) trait GetHash {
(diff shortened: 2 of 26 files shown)


rust/automerge/src/automerge.rs
```diff
@@ -717,6 +717,9 @@ impl Automerge {
                 );
                 Self::new()
             }
+            storage::Chunk::Bundle(_b) => {
+                todo!()
```
alexjg: This can be implemented now right?

rust/automerge/src/op_set2/change.rs
```diff
@@ -14,8 +14,10 @@ use std::num::NonZero;
 use std::ops::Range;
 
 pub(crate) mod batch;
+//pub(crate) mod bundle;
```
alexjg: nit: remove commented code

rust/automerge/src/op_set2/change.rs
```diff
@@ -14,8 +14,10 @@ use std::num::NonZero;
...
 pub(crate) mod collector;
 
+//pub(crate) use bundle::{Bundle, BundleMetadata};
```
alexjg: nit: remove commented code

rust/automerge/src/storage/bundle/builder.rs
```diff
@@ -0,0 +1,895 @@
...
+    start_op: u64,
+    max_op: u64,
+}
+
+#[derive(Debug)]
+pub struct BundleChangeIter<'a>(BundleChangeIterUnverified<'a>);
+
+impl<'a> BundleChangeIter<'a> {
+    pub(crate) fn new(columns: &RawColumns<compression::Uncompressed>, data: &'a [u8]) -> Self {
+        Self(BundleChangeIterUnverified::try_new(columns, data).unwrap())
```
alexjg, line 479: Might be worth calling this something like `new_from_verified` and documenting that it will panic if passed incorrectly formatted columns.

rust/automerge/src/storage/bundle/builder.rs
```diff
@@ -0,0 +1,895 @@
...
+    deps: CursorIter<'a, DeltaCursor>,
+    extra_count: CursorIter<'a, UIntCursor>,
+    extra: &'a [u8],
+}
+
+impl<'a> Iterator for BundleChangeIterUnverified<'a> {
+    type Item = Result<BundleChange<'a>, ParseError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.try_next().transpose()
```
alexjg: I think this has slightly unexpected behavior w.r.t the `Iterator` trait. If the iterator returns an error it will return `Some(Err(..))`, which means it would be valid to call `Iterator::next` again. The next return value from `next` will make no sense whatsoever as we probably bailed half way through iteration in `try_next`. I think we should probably track an errored state somewhere so if `next` is called after returning an error then we just return `None`.

rust/automerge/src/storage/bundle/builder.rs
```diff
@@ -0,0 +1,895 @@
...
+    fn next(&mut self) -> Option<OpBuilder<'a>> {
+        self.iter.next().map(|v| v.unwrap())
+    }
+}
+
+impl<'a> Iterator for OpIterUnverified<'a> {
+    type Item = Result<OpBuilder<'a>, ParseError>;
+
+    fn next(&mut self) -> Option<Result<OpBuilder<'a>, ParseError>> {
+        self.try_next().transpose()
```
alexjg: As above I think we should track an error state here so that we can return `None` if a previous `try_next` returned an `Err`.

rust/automerge/src/storage/bundle/storage.rs
```diff
@@ -0,0 +1,144 @@
...
+
+        let (i, ops_meta) = RawColumns::parse(i)?;
+        let ops_meta = ops_meta
+            .uncompressed()
+            .ok_or(parse::ParseError::Error(ParseError::CompressedOpCols))?;
+        let (_, ops) = parse::range_of(|i| parse::take_n(ops_meta.total_column_len(), i), i)?;
+        OpIterUnverified::try_new(&ops_meta, ops.value)
+            .map_err(|e| parse::ParseError::Error(ParseError::InvalidColumns(Box::new(e))))?;
+
+        // TODO how do I error if there are extra bytes here?
```
alexjg: I wonder if we even should error. Having extra content which we just ignore leaves the door open in the future for extensions which older implementations just ignore.

rust/hexane/src/delta.rs
```diff
@@ -737,4 +737,41 @@ pub(crate) mod tests {
...
+            .map(|d| d.unwrap())
+            .collect::<Vec<_>>();
+        assert_eq!(iter1, iter2);
+    }
+
+    // there is currently a bug where literal runs on slab boundaries save wrong
+    // currently not fixed as save_to is only used on locked encoders
+    /*
+    #[test]
+    fn delta_cursor_encoder_bug() {
```
alexjg, line 757: Rather than commenting this out can we use the [ignore attribute](https://doc.rust-lang.org/reference/attributes/testing.html#the-ignore-attribute) to not run it for now?

PR #1154, opened 2025-08-26
new feature "single-threaded" can be used to switch from Arc to Rc in…

axkibe:
… strictly single threaded use cases.

Hello, in cases when automerge is surely be ran from a single-thread only, there is a performance benefit to use Rc instead of Arc.

 I just experimented with it in a single threaded environment when I have Rc all over the place and getting Marks from the spans just felt weird having suddenly an Arc to pass around. (Or having to try_unwrap() or copy it from an Arc to Rc).

I didn't measure how much performance it really brings, so I'm not super invested into this.. I guess it might also introduce the need of adaption to other packages that build on this? So anyway, just wondering what you think about this.

Maybe the naming of the feature is not the best, since someone might think it would start threads otherwise..

### DIFF
--- a/rust/automerge/Cargo.toml
+++ b/rust/automerge/Cargo.toml
@@ -11,6 +11,7 @@ readme = "./README.md"
 [features]
 optree-visualisation = ["dot"]
 wasm = ["js-sys", "wasm-bindgen", "web-sys", "getrandom/wasm_js", "hexane/wasm"]
+single-threaded = []
 utf8-indexing = []
 utf16-indexing = []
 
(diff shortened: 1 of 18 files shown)


alexjg:
I appreciate the contribution, but without concrete performance numbers I don't think this is worth the additional complexity. If you have a specific usecase which is too slow due to the use of `Arc` then let's talk about how to fix that.

axkibe:
It's a fair call, I tested a few times with cargo test and the difference is below the measurement variance.

PR #1182, opened 2025-10-01
add diffPath to js interface

orionz:
Exposed the diff_obj function in javascript and added javascript tests.

Not sure I chose the right API - would like feedback.

Not wanting to break the rust api there is now a `diff_obj()` method that acts the same but takes an ExId argument.

For the wasm api I added a third option obj argument at the end of the function call b/c this doesn't add a new api call and doesn't break existing code.  It's odd tho b/c its the only api call with obj at the end and where obj is option.  Other options would be to add a `diff_obj` method that mirrors the rust API or to make obj manditory and make it the first argument so all current code calling `doc.diff(a,b)` would need to become `doc.diff("/",a,b)` which isnt a big deal since the only code calling the wasm api is the javascript api.

Lastly I added path as an option 3rd parameter to the javascript api.  So you can call

`Automerge.diff(doc, before, after)` and `Automerge.diff(doc, before, after, path)`. this is also odd b/c all other Automerge methods taking a path put it after the object and its never optional.  Maybe this is fine - maybe we need a new method `diffObject()` - but we dont want to break this API and make path manditory b/c everyone uses it.

Thoughts?

### DIFF
--- a/rust/automerge/src/automerge.rs
+++ b/rust/automerge/src/automerge.rs
@@ -1188,11 +1188,22 @@ impl Automerge {
         patch_log.make_patches(self)
     }
 
-    /// Create patches representing the change in the current state of an object in the document between the
-    /// `before` and `after` heads.  If the arguments are reverse it will observe the same changes
-    /// in the opposite order.
+    /// Create patches representing the change in the current state of an object
+    /// in the document between the `before_heads` and `after_heads` heads. If
+    /// the arguments are reverse it will observe the same changes in the
+    /// opposite order.
     ///
-    /// if `recursive` is false the diff does not extend into child objects
+    /// # Arguments
+    ///
+    /// * `obj` - The object to start the diff at.
+    /// * `before_heads` - heads from [`Self::get_heads()`] at beginning point
+    ///   in the documents history
+    /// * `after_heads` - heads from [`Self::get_heads()`] at ending point in
+    ///   the documents history.
+    /// * `recursive` - if false, do not also diff child objects
+    ///
+    /// Note: `before_heads` and `after_heads` do not have to be chronological.
+    /// Document state can move backward.
     pub fn diff_obj(
         &self,
         obj: &ExId,
--- a/rust/automerge/src/autocommit.rs
+++ b/rust/automerge/src/autocommit.rs
@@ -289,7 +289,7 @@ impl AutoCommit {
     ///
     /// By default the diff requires a sequental scan of all the ops in the doc.
     ///
-    /// [Self::diff()] is the equivelent to [Self::diff_obj(&ROOT, before, after, true)]
+    /// [Self::diff()] is the equivelent to [Self::diff_obj(&ROOT, before, after)]
     ///
     /// Managing the diff index has a small but non-zero overhead.  It should be
     /// disabled if no longer needed.  If a signifigantly large change is applied


orionz:
Move everything into a "diffPath" method and added a "recursive" flag.

javascript/src/implementation.ts
```diff
@@ -1020,6 +1020,19 @@ export function diff(doc: Doc<unknown>, before: Heads, after: Heads): Patch[] {
...
+ * This is an experimental API
+ */
+export function diffPath(doc: Doc<unknown>, path: Prop[], before: Heads, after: Heads, recursive: boolean): Patch[] {
+  checkHeads(before, "before")
+  checkHeads(after, "after")
+  const state = _state(doc)
+  const objPath = absoluteObjPath(doc, path, "diff")
+  return state.handle.diff_path(objPath, before, after, recursive)
+}
+
```
msakrejda, line 1043: > Move everything into a "diffPath" method and added a "recursive" flag.

Yeah, I agree this makes the most sense given the adjacent API surface. But maybe that `recursive` parameter should be an optional `opts` object with a `recursive` field? It's hard to read invocations of functions that accept boolean parameters.

msakrejda: Nice, I think that looks good.

javascript/src/implementation.ts
```diff
@@ -1020,6 +1022,19 @@ export function diff(doc: Doc<unknown>, before: Heads, after: Heads): Patch[] {
   return state.handle.diff(before, after)
 }
 
+/**
+ * Create a set of patches representing the change from one set of heads to another at a given path
+ *
+ * This is an experimental API
```
alexjg, line 1034: Can we add an [`@experimental`](https://typedoc.org/documents/Tags._experimental.html) typedoc tag?

rust/automerge-wasm/src/lib.rs
```diff
@@ -339,6 +341,10 @@ export type LoadOptions = {
   convertImmutableStringsToText?: boolean;
 };
 
+export type DiffOptions = {
+  recursive?: boolean;
```
alexjg, line 346: Can we add some doc comments here explaining what "recursive" does?

rust/automerge-wasm/src/lib.rs
```diff
@@ -329,6 +329,8 @@ interface Automerge {
     getMissingDeps(heads?: Heads): Heads;
 
     getCursorPosition(obj: ObjID, cursor: Cursor, heads?: Heads): number;
+
+    diffPath(path: any, before: Heads, after: Heads, options?: DiffOptions): Patch[];
```
alexjg: We could type this as `path: Prop[] | string` I think

rust/automerge-wasm/src/lib.rs
```diff
@@ -1111,6 +1117,28 @@ impl Automerge {
         Ok(interop::export_patches(&self.external_types, patches)?)
     }
 
+    #[wasm_bindgen(js_name = diffPath, skip_typescript)]
+    pub fn diff_path(
+        &mut self,
+        path: JsValue,
```
alexjg: Here we could add `#[wasm_bindgen(unchecked_param_type = "Prop[] | string")]`.

rust/automerge/src/automerge.rs
```diff
@@ -1170,11 +1175,29 @@ impl Automerge {
...
 
+    /// Create patches representing the change in the current state of an object in the document between the
+    /// `before` and `after` heads.  If the arguments are reverse it will observe the same changes
+    /// in the opposite order.
+    pub fn diff_obj(
+        &self,
+        obj: &ExId,
+        before_heads: &[ChangeHash],
+        after_heads: &[ChangeHash],
+        recursive: bool,
```
alexjg, line 1212: We should add some docs about what `recursive` does.

orionz:
Added some documentation and testing around the recursive flag.  So yes - if recursive is false but the object is updated you'll get a patch setting the object the `{}` but not the patches to fill out the objects interior.  This behavior feels consistent to me.  Any objection?

PR #1184, opened 2025-10-17
lean load

orionz:
An alternate loading strategy that uses about half as much memory for loads with large changes.

```
                       old             new     
essay              458M @  970ms     346M @  870ms   
mobydick           112M @  380ms     112M @  302ms
monday-notes        25M @   43ms      22M @   42ms
pathological      2767M @  8.4s     2511M @  8.7s 
unsinkable        7162M @ 15.5s     4884M @ 12.7s 
```

This provides some modest improvements and will make it easier to implement fully incremental loader that will be closer to optimal memory use.

The optimal loader would consume a `change_meta_iter` and on `op_iter` and only read from the change iter when it finds an op with no matching builder.  When the last op is loaded into the builder then the loader (an iterator itself) would emit the change.  This way - as long as ops are read in in document order then no additional memory would need to be allocated.  Worse case would be what we currently observer. 

Note: I had a version of this that got unsinkable down to 3.5G (good enough for 32bit memory) but had a signifgant performance impact.  I decided to ship this and may come back to implement the fully progressive version at a later date.

### DIFF
--- a/rust/automerge/src/op_set2/change/collector.rs
+++ b/rust/automerge/src/op_set2/change/collector.rs
@@ -163,7 +163,7 @@ impl<'a> OpEncoderStrategy<'a> {
         mapper.reset();
         match self {
             Self::Ops(mut v) => v.finish(change, mapper),
-            Self::Enc(e) => e.finish(change, mapper),
+            Self::Enc(e) => Ok(e.finish(change, mapper)),
             Self::Null => Err(Error::InvalidState),
         }
     }
@@ -269,6 +269,7 @@ impl<'a> VecEncoder<'a> {
     }
 
     fn add(&mut self, index: usize, op: OpBuilder<'a>) {
+        assert!(self.ops[index].is_none());
         self.data[index] = Some(op);
     }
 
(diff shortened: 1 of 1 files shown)


rust/automerge/src/op_set2/change/collector.rs
```diff
@@ -88,31 +117,411 @@ pub(crate) struct ChangeBuilder<'a> {
...
+            expand,
+            mark_name,
+        }
+    }
+
+    fn finish(
+        mut self,
+        change: &BuildChangeMetadata<'_>,
+        mapper: &mut ActorMapper<'_>,
+    ) -> Result<ChangeCols, Error> {
```
alexjg: I don't think this needs to be fallible

rust/automerge/src/op_set2/change/collector.rs
```diff
@@ -88,31 +117,411 @@ pub(crate) struct ChangeBuilder<'a> {
...
 
     pub(crate) fn max_op(&self) -> u64 {
-        self.start_op + self.ops.len() as u64 - 1
+        self.start_op + self.encoder.num_ops() - 1
     }
 
     pub(crate) fn add(&mut self, op: OpBuilder<'a>) {
         let counter = op.id.counter();
-        //if counter >= self.start_op && counter <= self.max_op() && op.id.actor() == self.actor {
-        assert!(self.ops[(counter - self.start_op) as usize].is_none());
```
alexjg, line 113: It looks like we've lost this assertion. It feels like a useful one to keep though. Maybe I'm missing where it moved to? I expected to see it in `VecEncoder::add`.

alexjg:
Some tiny comments but mostly LGTM, happy to merge without addressing if you prefer @orionz.

PR #1204, opened 2025-11-23
Fix type hints in applyPatch and applyPatches functions

### LINKED ISSUE #1183: Inconsistent types in API docs
I've noticed that some of the JS API docs have nice type hints, but others don't.

Example of good type hints: [applyChanges](https://automerge.org/automerge/api-docs/js/functions/applyChanges.html) is parameterized with `<T>`, which is then used for the `doc: Doc<T>` and `opts?: ApplyOptions<T>` arguments.

The following APIs have inconsistent or missing types.

### `doc: Doc<unknown>` / `doc: unknown` / `doc: any` / etc

* [applyPatch](https://automerge.org/automerge/api-docs/js/functions/applyPatch.html)
* [applyPatches](https://automerge.org/automerge/api-docs/js/functions/applyPatches.html)
* [diff](https://automerge.org/automerge/api-docs/js/functions/diff.html)
* [getObjectId](https://automerge.org/automerge/api-docs/js/functions/getObjectId.html)
  * The docs say "This is useful to determine if something is actually an automerge document, if doc is not an automerge document this will return null." I'd argue that people should use [isAutomerge](https://automerge.org/automerge/api-docs/js/functions/isAutomerge.html) for this. Now, `isAutomerge` indeed uses `getObjectId` behind the scenes, but this internal use of `getObjectId` shouldn't compromise the public-facing type hint offered by `getObjectId`.
* [hasHeads](https://automerge.org/automerge/api-docs/js/functions/hasHeads.html)
* [inspectChange](https://automerge.org/automerge/api-docs/js/functions/inspectChange.html)
* [saveSince](https://automerge.org/automerge/api-docs/js/functions/saveSince.html)
* [stats](https://automerge.org/automerge/api-docs/js/functions/stats.html)
* [updateText](https://automerge.org/automerge/api-docs/js/functions/updateText.html)

### Others
* [from](https://automerge.org/automerge/api-docs/js/functions/from.html) — `Record<string, unknown>` is a pretty loose way to describe a POJO.
* [getCursor](https://automerge.org/automerge/api-docs/js/functions/getCursor.html) — `CursorPosition` doesn't seem to be exported, which is a shame since I wanted to use it… which is what led me to look at the use of types across the docs in the first place. Yak, meet clippers.

Savithru7142:
Changed doc parameter type from 'unknown' to 'Doc<unknown>' in both applyPatch and applyPatches functions to add proper type hints.

Fixes #1183

### DIFF
--- a/javascript/src/apply_patches.ts
+++ b/javascript/src/apply_patches.ts
@@ -18,7 +18,7 @@ import {
   UnmarkPatch,
 } from "./wasm_types.js"
 
-export function applyPatch(doc: unknown, patch: Patch) {
+export function applyPatch(doc: Doc<unknown>, patch: Patch) {
   let path = resolvePath(doc, patch.path)
   if (patch.action === "put") {
     applyPutPatch(doc, path, patch)
@@ -183,7 +183,7 @@ function applyUnmarkPatch(
   )
 }
 
-export function applyPatches(doc: unknown, patches: Patch[]) {
+export function applyPatches(doc: Doc<unknown>, patches: Patch[]) {
   for (const patch of patches) {
     applyPatch(doc, patch)
   }


alexjg:
I don't think this is quite right as `applyPatch` doesn't have to be applied to an Automerge document and there are several parts of the code which do different things if in an Automerge document Vs not.

Was the `unknown` type annotation causing trouble for you in some way?

PR #1207, opened 2025-11-26
throw out of memory error on large loads for 32 bit wasm

orionz:
This PR has several parts

1. added OOM errors for large loads in 32bit wasm - this will happen ONLY when generating the hash graph for a document file (this is by far the biggest memory spike).
2. now copying change columns directly into hexane - same as with the op columns - leading to much more efficient per change loads
3. removal of large amounts of legacy column loading code plus general code cleanup
4. updated hexane to have a proper validation function

### DIFF
--- a/rust/automerge/src/op_set2/op_set/op_iter.rs
+++ b/rust/automerge/src/op_set2/op_set/op_iter.rs
@@ -12,11 +12,6 @@ use crate::{
     types::{ElemId, ObjId, OpId},
 };
 use hexane::{
-    BooleanCursor, ColGroupIter, ColumnData, ColumnDataIter, DeltaCursor, IntCursor, RawReader,
-    SlabWeight, StrCursor, UIntCursor,
-};
-
-use hexane::{
     BooleanCursor, ColGroupIter, ColumnData, ColumnDataIter, ColumnDataIterState, DeltaCursor,
     IntCursor, RawReader, SlabWeight, StrCursor, UIntCursor,
 };


kevinder:
Hi @orionz thanks for making this change. We are running automerge (wasm) on Cloudflare DO and seeing automerge hit OOM very often. Would you recommend we try using this change, or would you have any other advice on possible ways to avoid large memory spikes? We can consistently repro with about 10 concurrent users making edits on a new document, where OOM occurs consistently after the server is alive for 8-10 minutes. The only solution we've found that works is to restart the server before it hits the 8 minute mark or so, but this is not viable for other reasons. Thanks!

rust/automerge/src/op_set2/change/collector.rs
```diff
@@ -527,64 +555,88 @@ impl<'a> ChangeBuilder<'a> {
...
     pub(crate) fn from_bundle_changes(
         changes: Vec<BundleChange<'a>>,
         actors: &'a [ActorId],
     ) -> ChangeCollector<'a> {
         let changes = changes.into_iter().map(|c| c.into()).collect();
-        Self::from_change_meta(changes, 0, actors)
+        Self::from_change_meta(changes, actors)
+    }
+
+    pub(crate) fn try_from_change_meta(
```
alexjg, line 602: This is basically a copy of `from_change_meta` except that it might throw OutOfMemory right? There's a little bit of important logic duplicated here (sorting the builders and setting the builder index). Could we either a) use try_from_meta everywhere and unwrap or b) split the function into two parts with the common part shared between both `from_change_meta` and `try_from_change_meta`?

rust/automerge/src/op_set2/change/collector.rs
alexjg: The name of this method feels wrong now that we're no longer returning a change graph. Maybe we should just call it `collect()`?

rust/automerge/src/automerge.rs
```diff
@@ -1034,9 +1044,7 @@ impl Automerge {
         self.change_graph.get_hash_for_actor_seq(actor, seq)
     }
 
-    pub(crate) fn update_history(&mut self, change: &Change, num_ops: usize) {
-        self.max_op = std::cmp::max(self.max_op, change.start_op().get() + num_ops as u64 - 1);
-
+    pub(crate) fn update_history(&mut self, change: &Change, _num_ops: usize) {
```
alexjg: Can we remove the `_num_ops` parameter?

rust/automerge/src/storage/document.rs
```diff
@@ -316,18 +300,112 @@ impl<'a> Document<'a> {
...
+        }
+    }
+
+    pub(crate) fn reconstruct(
+        &self,
+        mode: VerificationMode,
+        text_encoding: TextEncoding,
+    ) -> Result<Automerge, ReconstructError> {
+        let mut op_set = OpSet::load(self, text_encoding)?;
+        let mut change_graph = ChangeGraph::load(self)?;
```
alexjg: My understanding of the need for a split between `load` and `set_hashes` is that it's so that we can load the change columns and pass those change columns to the change collector to actually compute the hashes, then call set_hashes once that's complete.

This split feels awkward though, it means that a `ChangeGraph` can be in this weird state where it has changes, but no graph structure and you just have to figure that out by reading through the code which constructed it. Could we instead split the columns out from `ChangeGraph`, so that we do something like

```rust
let change_cols = ChangeGraphCols::load(self)?;

...

let change_collector = ChangeCollector::try_new(&change_cols)?;

...

let change_graph = ChangeGraph::new(change_cols, change_collector.build_changegraph(&op_set)?.changes)
```

rust/automerge/src/change_graph.rs
```diff
@@ -902,3 +1049,171 @@ mod tests {
...
+    T: TryInto<u32> + Copy + Default,
+{
+    match i {
+        Err(e) => Err(e),
+        Ok(Some(i)) => Ok((*i).try_into().unwrap_or(0)),
+        Ok(None) => Ok(0),
+    }
+}
+
+/*
```
alexjg: nit: remove commented code

rust/automerge/src/change_graph.rs
```diff
@@ -902,3 +1049,171 @@ mod tests {
...
+    pub(super) const MAX_OP_COL_SPEC:     ColumnSpec = ColumnSpec::new_delta(MAX_OP_COL_ID);
+    pub(super) const TIME_COL_SPEC:       ColumnSpec = ColumnSpec::new_delta(TIME_COL_ID);
+    pub(super) const MESSAGE_COL_SPEC:    ColumnSpec = ColumnSpec::new_string(MESSAGE_COL_ID);
+    pub(super) const DEPS_COUNT_COL_SPEC: ColumnSpec = ColumnSpec::new_group(DEPS_COL_ID);
+    pub(super) const DEPS_VAL_COL_SPEC:   ColumnSpec = ColumnSpec::new_delta(DEPS_COL_ID);
+    pub(super) const EXTRA_META_COL_SPEC: ColumnSpec = ColumnSpec::new_value_metadata(EXTRA_COL_ID);
+    pub(super) const EXTRA_VAL_COL_SPEC:  ColumnSpec = ColumnSpec::new_value(EXTRA_COL_ID);
+
+/*
+    pub(super) const ALL_COLUMN_SPECS: [ColumnSpec; 9] = [
```
alexjg: nit: remove commented code

orionz:
> Hi @orionz thanks for making this change. We are running automerge (wasm) on Cloudflare DO and seeing automerge hit OOM very often. Would you recommend we try using this change, or would you have any other advice on possible ways to avoid large memory spikes? We can consistently repro with about 10 concurrent users making edits on a new document, where OOM occurs consistently after the server is alive for 8-10 minutes. The only solution we've found that works is to restart the server before it hits the 8 minute mark or so, but this is not viable for other reasons. Thanks!

This patch should cause and out of memory condition during load() to safely return an error.  This OOM condition only seems to happen on 32bit wasm.   If you have an opportunity to use native rust this would go away.  We're still trying to find ways to make the load process spike memory usage less but large documents currently can use well over 4G of ram to recomute the change graph.

rust/automerge/src/storage/document.rs
orionz: I agree - but the refactor is also pretty akward and the only place the change graph can be in this strange state is in one load function.  Furthermore all this code (should be) going away soon in favor of sedementry so maybe we can punt the cleanup with a nice comment rather than have to clean it up and then delete it?

orionz: Actually I think I found a simple compromise that solves the issue without a lot of work...

```rust 
pub(crate) struct ChangeGraphCols(ChangeGraph);
```

alexjg: T̵h̵a̵t̵'̵s̵ ̵f̵a̵i̵r̵.̵ ̵A̵ ̵c̵o̵u̵p̵l̵e̵ ̵o̵f̵ ̵g̵o̵o̵d̵ ̵c̵o̵m̵m̵e̵n̵t̵s̵ ̵w̵o̵u̵l̵d̵ ̵a̵l̵s̵o̵ ̵s̵u̵f̵f̵i̵c̵e̵ Ah nice!

PR #1228, opened 2025-12-09
sync request

orionz:
We found a problematic message in the wild that was using V1 while we know both sides support V2.  It would appear that the sync state didn't contain the other side's capabilities.  

This patch makes 4 changes:

1. Adds a `Capability::Request` so once side can request capabilities from the other should they be lost.  This should prevent sync connections from getting stuck in V1 when something goes wrong.
2. Stops excessive copying of hashes especially on the "send whole document" fast path
3. Will now send the whole doc anytime the number of changes to send is greater than 1/3 of the total.  The correct number varies by document makeup.  Our bestiary varies between 1/2 and 1/50th.
4. Significant code cleanup for better readability

### DIFF
--- a/rust/automerge/src/sync.rs
+++ b/rust/automerge/src/sync.rs
@@ -229,11 +229,6 @@ impl SyncDoc for Automerge {
             }
         }
 
-        // Only send the supported capabilities in the first message, the other end will store them
-        // in it's sync state and use them for subsequent messages
-
-        let supported_capabilities = sync_state.make_supported_capabilities();
-
         sync_state.have_responded = true;
         sync_state.last_sent_heads.clone_from(&our_heads);
         sync_state.sent_hashes.extend(message_builder.hashes());
@@ -242,7 +237,7 @@ impl SyncDoc for Automerge {
             .heads(our_heads)
             .have(our_have)
             .need(our_need)
-            .supported_capabilities(supported_capabilities)
+            .supported_capabilities(Some(vec![Capability::MessageV1, Capability::MessageV2]))
             .build();
 
         sync_state.in_flight = true;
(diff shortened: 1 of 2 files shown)


rust/automerge/src/sync/state.rs
```diff
@@ -120,6 +120,36 @@ impl State {
...
+
+    pub(crate) fn their(&self) -> Option<(&[Have], &[ChangeHash])> {
+        let have = self.their_have.as_ref()?;
+        let need = self.their_need.as_ref()?;
+        Some((have.as_slice(), need.as_slice()))
+    }
+
+    pub(crate) fn make_supported_capabilities(&mut self) -> Option<Vec<Capability>> {
+        if let Some(cap) = self.their_capabilities.as_mut() {
+            if let Some(pos) = cap.iter().position(|c| *c == Capability::Request) {
```
alexjg: This logic is kind of confusing to me. Is the idea that if the other end has some how lost their sync state then they will end up creating a new sync message which has a `Capability::Request` in it and so we will know to send the supported capabilities back to them? Then we remove the request capability in order to not keep sending the supported capabilities in every message?

orionz: yes exactly - adding a comment to confirm

alexjg: I'm wondering if it would be better to just always send the supported capabilities. They're very small.

rust/automerge/src/sync.rs
```diff
@@ -257,15 +231,12 @@ impl SyncDoc for Automerge {
 
         // Only send the supported capabilities in the first message, the other end will store them
         // in it's sync state and use them for subsequent messages
-        let supported_capabilities = if sync_state.have_responded {
-            None
-        } else {
-            Some(vec![Capability::MessageV1, Capability::MessageV2])
-        };
+
```
alexjg: We should remove the comments above this now. Also `make_supported_capabilities` doesn't need to return `Option` any more

PR #1265, opened 2026-01-23
author first pass

orionz:
Wanted feedback on the API before adding the wasm/js interfaces.  The new interface consists of `set_author()` `get_author()` `get_authors()` `get_actors_for_author()` and `get_author_for_actor()`

To keep things simple I'm leaving the author in extra bytes.  I'm having author only be assigned on changes where seq = 1.  I'm introdocing a new concept for extra bytes of a footer on changes.  The footer format would be leb128 encoded (id, length, bytes) ordered by id.  In theory any id's we dont know will just be more extra bytes where as ones we do know will come before and could be interpreted.  

The one big `todo` I havent addressed is re-using actorids.  When assigning an author we examine the document and find the actorid with the highest max_op and then reuse it.  I think this will make documents much smaller as a new actor wont be generated every time the document is re-opened.

Generally I've avoided adding special case code for author as much as possible and have left it all in the extra bytes.

Once this gets merged the next (harder) step will be adding the "revoke_author()" + diffs API


codeCraft-Ritik (commented):
Well structured code, especially the separation of concerns

rust/automerge/src/automerge.rs
```diff
@@ -284,6 +300,80 @@ impl Automerge {
...
+        self
+    }
+
+    /// Get the current author of this document.
+    pub fn get_author(&self) -> Option<&Author> {
+        self.author.as_ref()
+    }
+
+    /// Revoke all changes made by author after heads
+    /// Errors if given a heads not in the document yet
```
alexjg, line 350: "error" here means panic right? I think we've gone back on forth on this a few times but my current gut feeling is that we should just silently accept missing heads because it's very likely revocations will arrive out of band.

rust/automerge/src/op_set2/change/batch.rs
```diff
@@ -769,21 +775,25 @@ impl BatchApply {
 
     fn push(&mut self, c: Change) {
         assert!(!self.has_actor_seq(&c));
-        self.record_actor_seq(&c);
+        assert!(!self.has_actor_author(&c));
```
alexjg: I think this is too strict. If I understand correctly it will fire any time a batch has more than one changes with the same actor ID and where one has an author ID. E.g. this will fail:

     #[test]
     fn batch_multiple_changes_same_author() {
         let author = Author::try_from("aabbccdd").unwrap();
         let mut doc1 = AutoCommit::new().with_author(Some(author.clone()));
         doc1.put(&ROOT, "key1", "value1").unwrap();
         doc1.commit();
         doc1.put(&ROOT, "key2", "value2").unwrap();
         doc1.commit();
     
         let heads0 = [];
         let changes = doc1.get_changes(&heads0);
         assert_eq!(changes.len(), 2);
     
         let mut doc2 = AutoCommit::new();
         doc2.apply_changes_batch(changes).unwrap();
     }

What's the intent of the check?

rust/automerge/src/op_set2/columns.rs
```diff
@@ -384,6 +387,16 @@ impl Columns {
...
 
+        self.index
+            .counter
+            .splice(pos, 0, ops.clone().map(O::counter));
+
+        /*
+                for o in ops.clone() {
+                    log!("op={:?} vis={} top={}", o, O::visible(&o), O::top(&o));
+                }
+        */
```
alexjg: Nit: remove commented code

rust/automerge/src/op_set2/op_set.rs
```diff
@@ -523,11 +539,13 @@ impl OpSet {
...
-                let slow = self.seek_ops_by_index_slow(obj, index, seq_type, clock);
-                assert_eq!(found, slow, "fast != slow");
-            }
+            /*
+                        #[cfg(debug_assertions)]
+                        {
+                            let slow = self.seek_ops_by_index_slow(obj, index, seq_type, clock);
+                            assert_eq!(found, slow, "fast != slow");
+                        }
+            */
```
alexjg: nit: remove commented code

PR #1271, opened 2026-02-01
Fix duplicate splice results when using clone() + changeAt()

### LINKED ISSUE #1270: ChangeAt + clone causes duplicate splice results

```

    });

        // Works
    });

        // Works
    });
});
```

Thanks!

bxff:
Fixes #1270

## Problem

When using `clone()` with a new actor ID followed by `changeAt()`, duplicate content would appear in the document. For example:

```typescript
let doc = Automerge.change(Automerge.init(), doc => { doc.text = ""; });
// Expected: "abcdef"
// Actual: "abcdefdef" (duplicate "def")
```

This bug appeared flaky (~50% of the time with random actor IDs) but was actually deterministic based on actor ID ordering:
- When `cloneActorId < originalActorId` lexicographically: Bug occurs 100% of the time
- When `cloneActorId > originalActorId`: Bug never occurs

## Root Cause

The bug occurs due to a timing issue between diff computation and actor table insertion:

1. `fork()` creates a new document with the new actor stored as `Actor::Unused` (not yet in the actor table)
2. `isolate()` calls `patch_to()` which computes diffs using the current actor table
3. Events are logged to the `PatchLog` with OpIds referencing the current actor indices
4. Later, when a transaction opens, the new actor is inserted into the actor table
5. If the new actor sorts before existing actors, all actor indices shift
6. The `migrate_actors()` function is designed to fix this, but it has a gap: when `self.actors` is empty (which is the case for a freshly forked document), it just copies the new actor list and returns without migrating existing events
7. The events logged by `patch_to()` now have stale actor indices, causing patches to reference wrong objects

## Solution

Ensure the actor is inserted into the actor table **before** computing any diffs in `patch_to()`. This prevents the index mismatch from occurring in the first place.

The fix adds `ensure_actor_in_table()` at the start of `patch_to()`:

```rust
    // Ensure the actor is in the actor table before computing diffs.
    // If a new actor is inserted later, it would shift actor indices and
    // cause the diff computation to be incorrect. See issue #1270.
    self.doc.ensure_actor_in_table();
    
    // ... rest of the function
}
```

This is a preventative fix rather than a corrective one - by stabilizing the actor table before logging any events, we avoid the need for migration entirely.

## Changes

- `rust/automerge/src/automerge.rs`: Added `ensure_actor_in_table()` method
- `rust/automerge/src/autocommit.rs`: Call `ensure_actor_in_table()` at the start of `patch_to()`
- `rust/automerge/tests/test.rs`: Added regression test for issue #1270

## Testing

- All existing Rust tests pass (83 tests)
- All existing JavaScript tests pass (309 tests)
- New regression test verifies the fix works for the reported scenario

### DIFF
--- a/rust/automerge/src/autocommit.rs
+++ b/rust/automerge/src/autocommit.rs
@@ -715,6 +715,11 @@ impl AutoCommit {
     }
 
     fn patch_to(&mut self, after: &[ChangeHash]) {
+        // Ensure the actor is in the actor table before computing diffs.
+        // If a new actor is inserted later, it would shift actor indices and
+        // cause the diff computation to be incorrect. See issue #1270.
+        self.doc.ensure_actor_in_table();
+
         // we may be isolated so we dont use self.doc.get_heads()
         let before = self.get_heads();
         if before.as_slice() != after {
--- a/rust/automerge/src/automerge.rs
+++ b/rust/automerge/src/automerge.rs
@@ -332,6 +332,15 @@ impl Automerge {
         }
     }
 
+    /// Ensure the actor is in the actor table.
+    ///
+    /// This is useful when you need to ensure the actor table is stable before
+    /// computing diffs that depend on actor indices. If the actor is already
+    /// in the table, this is a no-op.
+    pub(crate) fn ensure_actor_in_table(&mut self) {
+        let _ = self.get_or_create_actor_index();
+    }
+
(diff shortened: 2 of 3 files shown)


bxff:
@alexjg Friendly ping - just want to make sure you saw my follow-up above with the pure Rust reproduction that confirms the bug is still present on main (the issue is in `diff_incremental()` patches, not the document state itself).

PR #1326, opened 2026-03-26
'`doc.fragments()`' experimental - rust only

orionz:
experimental fragments function where automerge tracks fragments incrementally as the document is loaded and manipulated.

Using SeqClock's internally to quickly calculate set membership for fragments.  Not 100% sure if this is to spec but its a frist pass.  Performance cost to load was less than 1% of load time. 

thanks to Cladue for the tests.

### DIFF
--- a/rust/automerge/src/change_graph.rs
+++ b/rust/automerge/src/change_graph.rs
@@ -1,6 +1,6 @@
 use std::borrow::Cow;
 use std::cmp::Ordering;
-use std::collections::{BTreeSet, HashMap};
+use std::collections::{BTreeSet, HashMap, HashSet, VecDeque};
 use std::num::NonZeroU32;
 use std::ops::Add;
 
@@ -46,6 +46,8 @@ pub(crate) struct ChangeGraph {
     nodes_by_hash: HashMap<ChangeHash, NodeIdx>,
     clock_cache: HashMap<NodeIdx, SeqClock>,
     seq_index: Vec<Vec<NodeIdx>>,
+    fragment_top: SeqClock,
+    fragments: Vec<FragmentNode>,
 }
 
 pub(crate) struct ChangeGraphCols(ChangeGraph);
@@ -102,6 +104,8 @@ impl ChangeGraph {
             heads: BTreeSet::new(),
             clock_cache: HashMap::new(),
             seq_index: vec![vec![]; num_actors],
+            fragments: vec![],
+            fragment_top: SeqClock::new(num_actors),
         }
     }
 
@@ -148,6 +152,10 @@ impl ChangeGraph {
         for clock in self.clock_cache.values_mut() {
             clock.rewrite_with_new_actor(idx)
         }
+        for f in &mut self.fragments {
+            f.clock.rewrite_with_new_actor(idx)
+        }
+        self.fragment_top.rewrite_with_new_actor(idx);
         self.seq_index.insert(idx, vec![]);
     }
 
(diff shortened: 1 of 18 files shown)


rust/automerge/src/types.rs
```diff
@@ -676,6 +676,10 @@ impl ChangeHash {
     pub(crate) fn checksum(&self) -> [u8; 4] {
         [self.0[0], self.0[1], self.0[2], self.0[3]]
     }
+
+    pub(crate) fn fragment_level(&self) -> usize {
+        self.0.iter().rev().take_while(|&&b| b == 0).count()
```
expede: We can do this, but also leading zeros may be simpler — we've switched to leading zeros (no `.rev`) in Subduction. (Or, well, in Subduction we take a function `Hash -> usize` so that folks can try optimising their own strategies but 1/256 makes sense for AM)

rust/automerge/src/change_graph.rs
```diff
@@ -558,10 +566,111 @@ impl ChangeGraph {
...
+                }
+            }
+        }
+        for i in supercede {
+            self.fragments.remove(i);
+        }
+        SeqClock::merge(&mut self.fragment_top, &clock);
+        self.fragments.push(FragmentNode { id, deps, clock });
+    }
+
```
expede: ✨ nice & succinct!

rust/automerge/src/change_graph.rs
```diff
@@ -558,10 +566,111 @@ impl ChangeGraph {
...
+            for p in self.parents(idx) {
+                if !seen.contains(&p) {
+                    let actor = self.actors[p.0 as usize].into();
+                    let seq = self.seq[p.0 as usize];
+                    if clock.get_for_actor(&actor) < NonZeroU32::new(seq) {
+                        seen.insert(p);
+                        to_visit.push_back(p);
+                    }
+                }
+            }
```
expede, line 663: I definitely don't understand the AM internals fully, but if `I` is large will this be in effect a full table scan? I guess with the columns you need to check withing a rough upper and lower bound?

rust/automerge/src/types.rs
orionz: Happy to follow whatever the spec is - but the rev() is free in this case

rust/automerge/src/change_graph.rs
orionz: So the magic happens in clock.get_for_actor() < seq because the clock is a vec<seq> that covers all the changes in the fragments deps.  So this is never a full table scan (unless theres one fragment and it covers all changes - in which case it is a full table scan but is still needed to generate the correct order.  Basically I save a clock per fragment that is the actor+seq of all changes in or before a fragment.  Then when walking through its deps I just need a clock which is the union of the clocks of the fragments deps.

expede (approved):
I have some Sedimentree benches for you, which I believe we can describe as "pretty good"

| Bench                       | BEFORE      | AFTER     | Speedup |
|-----------------------------|------------:|----------:|--------:|
| build_fragment_store / C1   |    24.9 s   | 27.6 ms   |   ~900× |
| build_fragment_store / C2   |    57.1 s   |  9.5 ms   | ~6,000× |

so yeah... PLEASE LET'S MERGE THIS!! :fire::fire::fire:

expede (changes requested):
ACTUALLY hold off on merging... we might have a semantic issue. I'm writing some tests to confirm

expede (approved):
I got a great walk through live on a call — thanks Orion, this looks great! 🚢

PR #1359, opened 2026-04-20
Stop panicking on ChangeCollector MissingOps

rgbkrk:
Superseded by #1366.

This PR was the first attempt to turn the #1327 `MissingOps` panic into a propagated error. After a smaller current-main reproduction, #1366 now fixes the root cause in `fork_at` dependency traversal instead.

Keeping this open only as history for the panic-propagation approach. Unless maintainers prefer that direction, #1366 is the PR to review.

## Original approach

- Make `from_build_meta_inner()` and `from_build_meta()` return `Result`.
- Propagate collector errors through `for_hashes()`.
- Add a regression test for the `MissingOps` panic.

## Related PRs

- #1366 fixes the root cause.
- #1360 keeps other fallible collector/import paths from panicking.
- #1367 documents the actor-stream invariant that came up while debugging the nteract usage.

### DIFF
--- a/rust/automerge/examples/test_missing_ops.rs
+++ b/rust/automerge/examples/test_missing_ops.rs
@@ -1,0 +1,474 @@
+// Attempt to reproduce the MissingOps panic in automerge's change collector.
+//
+// The panic occurs at collector.rs:761 when from_build_meta_inner calls
+// finish() on a VecEncoder that has gaps (None slots) in its op sequence.
+//
+// Key hypothesis: character-by-character splice_text from one actor,
+// interleaved with fork+merge from another actor, with
+// generate_sync_message calls in between, triggers the bug.
+
+use automerge::{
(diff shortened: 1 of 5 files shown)


alexjg:
Neither of the tests that are in this PR fail on `main` for me. Do they fail for you?

rgbkrk:
Oh, so maybe this was limited to when I was on 0.7.4. I'll go shake this a bit more.

rgbkrk:
Ok, I've got a test patch that fails on current main now @alexjg.

rgbkrk:
Ended up diving deeper, #1366 is the PR I'd want reviewed.

rgbkrk:
I'd love to see this work (or at least the intent) get folded into a PR. I recognize though that adding `Result` to many more functions and call sites will be a major version change difference (though panic-ing is far worse).

PR #1360, opened 2026-04-20
Propagate errors from change building, unbundle, and import_obj

rgbkrk:
This is defensive cleanup for fallible paths that already return `Result`. It is not the root-cause fix for #1327. I ended up handling that in #1366.

## Changes

- `OpEncoderStrategy::finish`: a missing dependency index now returns `Error::ChangesOutOfOrder` instead of panicking.
- `ChangeCollector::unbundle`: `builders[i].finish()` already returns `Result`, so the error is propagated with `?`.
- `Automerge::import_obj`: malformed actor hex now returns `AutomergeError::InvalidObjIdFormat`, matching the counter parse path.

## Tests

- Bad actor hex in `import_obj` returns an error.
- Missing dependency indexes in the collector return `ChangesOutOfOrder`.
- Missing ops in `unbundle` return `MissingOps`.

## Verification

```text
cargo fmt
cargo test -p automerge --manifest-path rust/Cargo.toml
```

### DIFF
--- a/rust/automerge/src/automerge.rs
+++ b/rust/automerge/src/automerge.rs
@@ -1145,7 +1145,10 @@ impl Automerge {
             let counter = s[0..n]
                 .parse()
                 .map_err(|_| AutomergeError::InvalidObjIdFormat(s.to_owned()))?;
-            let actor = ActorId::from(hex::decode(&s[(n + 1)..]).unwrap());
+            let actor = ActorId::from(
+                hex::decode(&s[(n + 1)..])
+                    .map_err(|_| AutomergeError::InvalidObjIdFormat(s.to_owned()))?,
+            );
             let actor = self
                 .ops
                 .lookup_actor(&actor)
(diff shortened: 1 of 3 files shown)


rust/automerge/src/automerge.rs
```diff
@@ -1145,7 +1145,10 @@ impl Automerge {
             let counter = s[0..n]
                 .parse()
                 .map_err(|_| AutomergeError::InvalidObjIdFormat(s.to_owned()))?;
-            let actor = ActorId::from(hex::decode(&s[(n + 1)..]).unwrap());
+            let actor = ActorId::from(
+                hex::decode(&s[(n + 1)..])
+                    .map_err(|_| AutomergeError::InvalidObjIdFormat(s.to_owned()))?,
```
FintanH, line 1150: There's a variant `InvalidActorId`, which I believe would be more suitable here

rgbkrk:
Honestly we can probably drop these PRs, they're super outdated now.

PR #1363, opened 2026-04-24
automerge-wasm: opt-in panic=unwind build flow

guybedford:
This builds the wasm artifact shipped to JS consumers with the nightly toolchain and `panic=unwind`, and registers a console-logging handler for hard wasm aborts. Together this means panics in exported Rust functions surface as `PanicError` exceptions at the JS boundary (async exports reject the returned promise) instead of silently aborting the WASM module, and the rarer case of a real wasm trap leaves a diagnostic in the console.

Both wasm-producing build paths switch to nightly:

* `rust/automerge-wasm/scripts/compile.mjs` (used by `npm run build`/`release` in the wasm package) now invokes `cargo +nightly build --target wasm32-unknown-unknown -Zbuild-std=std,panic_unwind` with `RUSTFLAGS="-C panic=unwind"` composed onto any user-provided `RUSTFLAGS`.
* `javascript/scripts/build.mjs` (used by the top-level `@automerge/automerge` package build) uses the same nightly + `-Zbuild-std` invocation.

The workspace pins stable in `rust/rust-toolchain.toml`; the nightly toolchain is selected per-invocation via rustup's `+nightly` argument so nothing else in the workspace is affected.

A `#[wasm_bindgen(start)]` function in `rust/automerge-wasm/src/lib.rs` registers an abort handler via `wasm_bindgen::__rt::set_on_abort`. When a wasm trap (memory OOB, OOM, double panic, etc.) sets `__instance_terminated`, the handler logs a `console.error` line so the abort is visible rather than only showing up as the generic "Module terminated" error on later calls.

```rust
#[wasm_bindgen(start)]
fn on_start() {
    fn log_abort() {
        web_sys::console::error_1(&"automerge-wasm: WASM instance aborted; …".into());
    }
    let _ = wasm_bindgen::__rt::set_on_abort(log_abort);
}
```

CI (`build_wasm`, `js_tests`, `node_18_packaging_test`) and the release workflow now install nightly + `rust-src` + the wasm32 target on the nightly toolchain. README and HACKING docs updated accordingly.

### Testing

* `npm run build && npm test` in `rust/automerge-wasm` — 155 passing
* `node scripts/build.mjs && npm test` in `javascript` — 312 passing

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -75,9 +75,11 @@ jobs:
         with:
           # Optional version of wasm-bindgen to install(eg. '0.2.83', 'latest')
           version: "0.2.118"
-      - name: Install wasm32 target
-        working-directory: rust
-        run: rustup target add wasm32-unknown-unknown
+      - name: Install nightly + wasm32 target + rust-src
+        run: |
+          rustup toolchain install nightly --profile minimal
+          rustup component add rust-src --toolchain nightly
+          rustup target add --toolchain nightly wasm32-unknown-unknown
       - name: run tests
         run: ./scripts/ci/wasm_tests
 
@@ -95,9 +97,11 @@ jobs:
         with:
           # Optional version of wasm-bindgen to install(eg. '0.2.83', 'latest')
           version: "0.2.118"
-      - name: Install wasm32 target
-        working-directory: rust
-        run: rustup target add wasm32-unknown-unknown
+      - name: Install nightly + wasm32 target + rust-src
+        run: |
+          rustup toolchain install nightly --profile minimal
+          rustup component add rust-src --toolchain nightly
+          rustup target add --toolchain nightly wasm32-unknown-unknown
       - name: run tests
         run: ./scripts/ci/js_tests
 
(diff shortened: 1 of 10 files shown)


alexjg:
Am I right in thinking that the main reason to make this optional is the use of the nightly compiler? If that's the case then I think we should just switch to using the nightly compiler for the JS package build and make the panic unwinding non optional. This seems like a useful enough improvement to be worth a little build hassle.

guybedford:
Yes, making this the default can be sensible I just didn't want to assume that - the nightly compiler is the issue exactly.

alexjg:
In that case let's go ahead and make it the default. I think we should pin to a specific version of nightly (whatever the current version is I guess) in version control so we have a reproducible build.

guybedford:
Added, I've also added an abort hook for reference.

alexjg:
@guybedford I've rebased on `main` and added a fix for the webpack build issue. That's on #1379 because I couldn't push to your branch. You're still down as an author on most of the commits there.

alexjg:
Closing in favor of #1379

PR #1366, opened 2026-05-04
Fix fork_at dependency traversal duplication

rgbkrk:
Fixes #1327.

`Automerge::fork_at` walks backwards from the requested heads and builds the set of changes to replay. The old traversal marked a hash as seen only after it was popped from the stack. In some DAG shapes the same dependency could be scheduled twice before either copy was processed.

That duplicate hash produced duplicate `BuildChangeMetadata` entries for the same `(actor, seq, op range)`. The collector would fill one builder and then try to finish the duplicate, surfacing as `MissingOps`.

## Changes

- Mark dependency hashes as seen when they are scheduled.
- Add a regression test for historical `fork_at` after sync/merge activity.

## Fails on main

With only the new test applied to `origin/main`:

```text
cargo test -p automerge test_fork_at_historical_heads_after_sync --test test
called `Result::unwrap()` on an `Err` value: MissingOps
```

## Related PRs

- #1360 is separate hardening for fallible paths that already return `Result`.
- #1367 documents the actor-stream invariant that came up while debugging nteract's fork usage.
- #1359 was the earlier panic-propagation approach and is now superseded by this root-cause fix.

## Verification

```text
cargo fmt --all
cargo test -p automerge test_fork_at_historical_heads_after_sync --test test
cargo test -p automerge --test test
```

### DIFF
--- a/rust/automerge/src/automerge.rs
+++ b/rust/automerge/src/automerge.rs
@@ -575,20 +575,24 @@ impl Automerge {
     ///
     /// This will create a new actor ID for the forked document
     pub fn fork_at(&self, heads: &[ChangeHash]) -> Result<Self, AutomergeError> {
-        let mut seen = heads.iter().cloned().collect::<HashSet<_>>();
-        let mut heads = heads.to_vec();
+        let mut seen = HashSet::new();
+        let mut stack = Vec::new();
+        for head in heads {
+            if seen.insert(*head) {
+                stack.push(*head);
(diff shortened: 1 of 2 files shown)


FintanH:
Thanks for the PR! I just validated that the original issue has been solved with the release of `0.11.0`. I'm going to close this PR for the time being, but please feel free to ping us if you still think this an issue.

PR #1367, opened 2026-05-04
Document fork actor concurrency invariant

rgbkrk:
This documents an invariant that came up while debugging nteract's `fork` and `fork_at` usage: an actor ID identifies one linear stream of changes. Concurrent branches need distinct actors.

This is separate from the `MissingOps` root-cause fix in #1366. It is guidance for using the API without accidentally creating same-actor concurrent histories.

## Changes

- Document the actor-stream invariant on `fork`, `fork_at`, `set_actor`, and `with_actor`.
- Point same-logical-actor historical work at `AutoCommit::isolate()` / isolated transactions.
- Add tests for same-actor fork collisions and `fork()` plus `isolate(...)`.

## Verification

```text
cargo fmt --all -- --check
cargo test -p automerge --test fork_actor_invariant
cargo test -p automerge --doc
RUSTDOCFLAGS="-D warnings" cargo doc -p automerge --no-deps
```

### DIFF
--- a/rust/automerge/src/autocommit.rs
+++ b/rust/automerge/src/autocommit.rs
@@ -31,8 +31,13 @@ use crate::{LoadOptions, VerificationMode};
 /// [`Self::merge()`].
 ///
 /// If you have a document you want to split into two concurrent threads of execution you can use
-/// [`Self::fork()`]. If you want to split a document from ealier in its history you can use
-/// [`Self::fork_at()`].
+/// [`Self::fork()`]. If you want to split a document from earlier in its history you can use
+/// [`Self::fork_at()`]. Both methods give the fork a new random actor ID because an actor ID
+/// identifies a single linear sequence of changes. If the original document and a fork can both
+/// write before being merged, do not reset them to the same actor ID; the merge may fail with
+/// [`AutomergeError::DuplicateSeqNumber`]. To make historical writes against a full document while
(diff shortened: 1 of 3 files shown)


rust/automerge/src/autocommit.rs
```diff
@@ -334,6 +339,12 @@ impl AutoCommit {
         patches
     }
 
+    /// Fork this document at the current heads for use by a different actor.
+    ///
+    /// The fork receives a new random actor ID. Keep that actor, or another actor that will not be
+    /// used concurrently elsewhere, if both documents may write before they are merged. Actor IDs
```
FintanH, line 345: I am not sure what you're trying to get across in the sentence about "Keep that actor". I might hazard a guess that you mean that the actor should not be manually set to another one? If so, you worded it better in the above documentation on `Autocommit` 😊

FintanH (commented):
I think these are great additions to the documentation! @alexjg could you get your eyes over these as well?

rgbkrk:
I'll just drop this PR as it is because it's out of date.

rust/automerge/src/autocommit.rs
rgbkrk: As in keep the randomly assigned actor ID rather than trying to reuse a previous actor ID.

PR #1371, opened 2026-05-05
rust: add experimental smallest wasm build

pvh:
## Summary
- Adds an opt-in `WASM_SMALLEST=1` path for `automerge-wasm` package builds.
- Provides `npm run release:smallest` as an explicit nightly/unstable build mode.
- Uses nightly `build-std`, immediate-abort panics, and no panic location details to minimize the generated WASM artifact.

## Strategy
The stable stack keeps the default release path conservative and mergeable. This PR captures the largest build-only size win we found as an explicit experimental mode instead of replacing the normal release build.

When `WASM_SMALLEST=1` is set, the build uses `cargo +nightly -Z build-std=std,panic_abort` with `RUSTFLAGS='-Zunstable-options -Cpanic=immediate-abort -Zlocation-detail=none'`. That rebuilds the standard library for a smaller panic/runtime footprint.

## Tradeoffs
- Requires nightly Rust and unstable Cargo/rustc flags.
- Produces warnings from existing no-op logging macro imports under nightly; the build still succeeds.
- Panic behavior is more abrupt and carries less diagnostic location information.
- This should be treated as an optional artifact mode until the team decides whether unstable build plumbing belongs in release automation.

## Size impact
Local measurement found this path reduced the optimized raw WASM artifact from roughly 1.46 MB on the stable stack to roughly 1.25 MB with nightly immediate-abort/location stripping.

## Test plan
- Verified `WASM_SMALLEST=1 PROFILE=wasm-release TARGET_DIR=wasm-release node ./target.mjs compile` succeeds locally.
- Earlier measured bindgen and `wasm-opt` outputs with the same nightly flags.

### DIFF
--- a/javascript/scripts/build.mjs
+++ b/javascript/scripts/build.mjs
@@ -113,13 +113,36 @@ function isStep(stepStr) {
 function buildWasm(outputDir, gitHead) {
   const automergeWasmPath = path.join(rustProjectDir, "automerge-wasm")
   const wasmProfile = process.env.WASM_PROFILE ?? "wasm-release"
+  const smallestWasm = process.env.WASM_SMALLEST === "1"
+  const cargoArgs = [
+    "build",
+    "--target",
+    "wasm32-unknown-unknown",
+    "--profile",
+    wasmProfile,
+  ]
+  const cargoEnv = { ...process.env, GIT_HEAD: gitHead }
+
+  if (smallestWasm) {
+    cargoArgs.unshift("+nightly")
+    cargoArgs.splice(2, 0, "-Z", "build-std=std,panic_abort")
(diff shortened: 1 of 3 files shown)


pvh:
Measured final artifact sizes from the top of this stack using the opt-in nightly-smallest path:

- `wasm-opt` artifact: **1,247,856 bytes raw**, **504,201 bytes gzip -9**
- bindgen-only artifact: **1,384,149 bytes raw**, **460,556 bytes gzip -9**

So `wasm-opt` still wins for smallest raw/runtime-optimized WASM, but the bindgen-only artifact currently wins for smallest gzip transfer size. That likely means we should keep tracking raw size and compressed delivery size separately when deciding what to publish or serve.

PR #1379, opened 2026-05-07
Wasm panic unwind

alexjg:
Rebasing #1363 on `main` to pick up the latest wasm-bindgen and switching to loading the webassembly as base64 in webpack. 

The latter point there might be a little surprising. The problem is that the panic unwinding stuff introduces a new import definition of type `0x4` in the webassembly for the exception types used by the webassembly exception handling spec. Unfortunately WebPacks wasm parser (@webassembly/wasm-parser) has not been updated for two years and chokes on this type, which means that webpack can't handle the bundler output. I've opted here to switch to bundling the base64 for webpack (and only for webpack, other bundlers are unaffected). We've already had this problem once before (https://github.com/automerge/automerge/pull/1025). I've [filed an issue](https://github.com/xtuc/webassemblyjs/issues/1210), if that gets fixed then we can revert to using the wasm directly. I think the advantages of panic handling outweight the disadvantage of a 30% file size increase for webpack users.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -75,9 +75,11 @@ jobs:
         with:
           # Optional version of wasm-bindgen to install(eg. '0.2.83', 'latest')
           version: "0.2.121"
-      - name: Install wasm32 target
-        working-directory: rust
-        run: rustup target add wasm32-unknown-unknown
+      - name: Install nightly + wasm32 target + rust-src
+        run: |
+          rustup toolchain install nightly --profile minimal
+          rustup component add rust-src --toolchain nightly
+          rustup target add --toolchain nightly wasm32-unknown-unknown
       - name: run tests
         run: ./scripts/ci/wasm_tests
 
@@ -95,9 +97,11 @@ jobs:
         with:
           # Optional version of wasm-bindgen to install(eg. '0.2.83', 'latest')
           version: "0.2.121"
-      - name: Install wasm32 target
-        working-directory: rust
-        run: rustup target add wasm32-unknown-unknown
+      - name: Install nightly + wasm32 target + rust-src
+        run: |
+          rustup toolchain install nightly --profile minimal
+          rustup component add rust-src --toolchain nightly
+          rustup target add --toolchain nightly wasm32-unknown-unknown
       - name: run tests
         run: ./scripts/ci/js_tests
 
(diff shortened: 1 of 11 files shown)


alexjg:
Alright, CI is failing because node 18 does not have `WebAssembly.JSTag`, as far as I can tell this doesn't appear until node 22. I'm kind of okay dropping support for node 18, it's been EOL for quite some time. Node 20 is still in maintenance though so I don't want to drop support for it if we can help it. There are a few options:

* Bundle two versions of the webassembly and choose which one we instantiate dynamically based on the presence of `WebAssembly.JSTag`
* Polyfill `WebAssembly.JSTag`. This would mean that exceptions thrown when calling JS functions from Rust would bubble up through the webassembly boundary and leave it in a very weird state. This might be okay for us though as we don't really call any fallible functions
* Drop `panic=unwind`

I think another thing to think about is that it seems to me like `panic=unwind` is really two features. One which propagates rust panics out of JS as nice exceptions and another which catches JS exceptions nicely in Rust. @guybedford does the propagating of rust panics as JS exceptions rely on `WebAssemblyl.JSTag`? If not would it be possible to split the feature into two options so that we can just enable the panic handling?

guybedford:
Unfortunately yes the exception handling work does depend on Node.js 22 only - this may be justification for keeping it as a feature flag for now until 20 support can be dropped (which is now EOL).

guybedford:
Specifically, the `WebAssembly.JSTag` is critical to how we are able to detect aborts, which while not strictly necessary for the pure panic=unwind story, does get bundled with it.

We could possibly investigate whether we can separate the unwind story from the well-defined aborts story to support something sooner here though.

guybedford:
I've posted an upstream tracking issue for this further in https://github.com/wasm-bindgen/wasm-bindgen/issues/5151.

guybedford:
@alexjg it would help to know if your primary limitation is the bundler or the Node.js version constraint here? Or both?

alexjg:
It's the node version. I _can_ get around it by creating two separate wasm blobs for node, but I would rather not add more bundles and build complexity if we can avoid it.

guybedford:
Ok, we'll continue to track at the upstream issue, and post any updates there.

joeblew999:
Adding myself to track this as I use automerge on Cloudflare . Pure rust in my case

guybedford:
I realise I didn't post the update on https://github.com/wasm-bindgen/wasm-bindgen/issues/5151 here - the Node.js 22 Exception.JSTag polyfill was landed in wasm-bindgen@0.2.122 resolving the blocker here which was tracked and resolved in specifically for this use case. Do just let me know if I can do more to assist!

PR #1380, opened 2026-05-07
Fix stale inactive patch log migration during sync

rgbkrk:
## Summary

- tolerate inactive, empty patch logs whose actor table is stale during sync receive
- propagate `PatchLogMismatch` from batch change application instead of panicking on `unwrap()`
- add a regression test that reproduces the stale inactive patch log sync path

## Details

During sync receive, `AutoCommit` can apply an incoming change while its internal patch log is inactive and empty. That patch log can still retain an older actor table. After the incoming change updates the document actor table, patch-log actor migration can see a mismatch.

For an inactive patch log with no events and no exposed IDs, there is nothing to migrate. Retargeting the log to the document actor table is safe and avoids a panic in the sync path. Non-empty mismatches still return `PatchLogMismatch`.

This is the upstream version of the downstream nteract/automerge patch used by nteract/desktop to stop `PatchLogMismatch` panics during runtime-state sync.

## Verification

- `cargo test --manifest-path rust/automerge/Cargo.toml sync_receive_with_stale_inactive_patch_log_does_not_fail -- --nocapture`
- `cargo fmt --all --check` from `rust/`
- `git diff --check origin/main..HEAD`

### DIFF
--- a/rust/automerge/src/autocommit.rs
+++ b/rust/automerge/src/autocommit.rs
@@ -1275,8 +1275,8 @@ impl OpRange {
 #[cfg(test)]
 mod tests {
     use crate::{
-        sync::SyncDoc, transaction::Transactable, ActorId, AutoCommit, ReadDoc, ScalarValue, Value,
-        ROOT,
+        patches::PatchLog, sync::SyncDoc, transaction::Transactable, ActorId, AutoCommit, ReadDoc,
+        ScalarValue, Value, ROOT,
     };
 
     fn is_send<S: Send>() {}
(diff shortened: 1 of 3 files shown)


alexjg:
I can't get the test that you've added here to fail when I remove the other changes. Any chance you could try and put together a test case which does exercise the failing code path?

rgbkrk:
Ok @alexjg `sync_receive_with_stale_inactive_patch_log_does_not_fail` fails on main now.

alexjg:
Great thanks. This test still fails on #1388 but the panic is where we would expect it - in the call to `receive_sync_message_log_patches` rather than an internal panic which can't be caught. Take a look at #1388 and let me know if you think it solves your problem.

rgbkrk:
Catching works great for me!

rgbkrk:
Have you had a chance to look at https://github.com/automerge/automerge/pull/1366

I'm much more confident of that one.

PR #1382, opened 2026-05-08
Hash DFS Sedimentree fragmentizer

expede:
Related #1326 

This PR was originally an experiment, but turns out that it actually may be worth merging. I was curious to see how much worse the performance would be to use hash walks for the DFS over the hand-tuned clock based ranges. The clock-based version was creating different fragments with different members — basically breaking the doc up differently. Sedimentree defines a particular fragmentation strategy with semantics that are important for doing distributed set reconciliation (which I won't go into here).

Why an experiment? If the hash version performed significantly worse, we may have had to reconsider how Sedimentree worked, but the non-compliant version was so much faster than our existing naive JS version by leveraging internals that I was hopeful that we could eat into some of that performance and still be reasonably fast.

Strategy:

* Leave existing implementation from #1326
* Bench that :point_up: 
* [Have Opus] write a spec-compliant implementation (called `Depth` below)
* Bench them head-to-head

The results were... surprising [machine generated output below]

# Fragmenter benchmark: depth-stratified vs. clock-based

**Bench:** `cargo bench -p automerge --bench fragments`
**Hardware:** AMD Ryzen AI 9 HX 370 (12c/24t @ 5.16 GHz boost) · 30 GiB · NixOS 25.11 / Linux 7.0.2 · rustc 1.89.0
**Bench Config**: single run, 100 samples/case, criterion defaults
**Topologies:**

- `linear` — one actor, one transaction per change.
- `concurrent2` — two actors making half the changes each on independent forks, then merged.
- `wide8` — eight actors making concurrent changes on independent forks, then merged.

## Query cost (head-to-head)

The only group where both strategies do work along the comparable critical path. Both start from the same prebuilt `Automerge` doc (the clock-based eager fragment cache already populated); we time `doc.fragments()` versus `doc.depth_fragments()`.

| Topology | Size | Clock-based mean (µs) | Clock-based 95% CI | Depth mean (µs) | Depth 95% CI | Speedup |
|---|---:|---:|:---|---:|:---|---:|
| linear | 1000 | 58.93 | [57.58, 60.54] | 31.28 | [31.06, 31.56] | 1.88× |
| linear | 10000 | 570.92 | [567.94, 574.34] | 239.78 | [238.86, 241.17] | 2.38× |
| concurrent2 | 1000 | 39.97 | [39.77, 40.19] | 25.32 | [25.26, 25.38] | 1.58× |
| concurrent2 | 10000 | 506.25 | [492.68, 524.80] | 252.35 | [243.28, 264.08] | 2.01× |
| wide8 | 1000 | 74.40 | [73.89, 75.22] | 38.91 | [38.66, 39.21] | 1.91× |
| wide8 | 10000 | 507.69 | [504.93, 510.76] | 287.80 | [285.18, 291.09] | 1.76× |

Depth is **1.58×–2.38× faster** on every topology and size. CIs do not overlap on any case.

## Fragment counts (semantic divergence)

Sampled during the bench run on prebuilt docs.

| Topology | Size | Heads | Clock-based fragments | Depth fragments |
|---|---:|---:|---:|---:|
| linear | 1000 | 1 | ~21 | ~4 |
| linear | 10000 | 1 | ~138 | ~46 |
| concurrent2 | 1000 | 2 | ~67 | ~5 |
| concurrent2 | 10000 | 2 | ~487 | ~45 |
| wide8 | 1000 | 8 | ~775 | ~6 |
| wide8 | 10000 | 8 | ~2649 | ~41 |

The clock-based design emits one `Fragment` per visited node above `fragment_top` (loose) and one per cached `FragmentNode`. Depth's spec-correct strategy emits one per commit with `fragment_level >= 1`, matching theory's ~1/256 rate. Numbers are approximate because `ChangeHash` content depends on per-run actor IDs.

## Why is this strategy faster?

It isn't because depth checks beat clock checks at the per-node level —
both are roughly O(1) reads (a leading-zero-byte count vs a per-actor
clock slot lookup). The win is structural.

The clock-based design emits one `Fragment` per cached `FragmentNode`
plus one per depth-0 commit above `fragment_top`. Each Fragment is
exported by running an independent `bfs_until_clock` (allocating a
fresh `HashSet`, `VecDeque`, and result `Vec`) and, for cached
fragments, a `calculate_clock(deps)` walk on top. On `wide8/10k` that
means ~2 600 small BFSes for ~10 000 nodes of work — per-BFS setup
dominates total cost.

The depth strategy emits one fragment per commit with `fragment_level >= 1`.
The same ~10 000 nodes are visited, but partitioned into ~50 BFSes
instead of ~2 600. The hot loop's `HashSet` stays warm through a
fragment-sized walk, parent-edge cache lines amortise across more
work, and we make ~100 result-`Vec` allocations instead of ~5 000.
Net effect: equal per-node cost × ≈50× fewer per-BFS setups + far
fewer small allocations = 1.5–2.5× speedup, even though the depth
path is lazy and the clock path is reading from a precomputed cache.

### DIFF
--- a/rust/automerge/Cargo.toml
+++ b/rust/automerge/Cargo.toml
@@ -76,3 +76,7 @@ harness = false
 [[bench]]
 name = "load_save"
 harness = false
+
+[[bench]]
+name = "fragments"
+harness = false
--- a/rust/automerge/benches/fragments.rs
+++ b/rust/automerge/benches/fragments.rs
@@ -1,0 +1,188 @@
+//! Benchmarks for the change-graph fragmenter.
+//!
+//! Two implementations live side by side on this branch:
+//!
+//! - `Automerge::fragments()` — the existing eager, clock-based prototype.
+//!   Runs `cache_fragment` per change inside `add_change`, so the work is
+//!   amortised across `apply_changes` / commit; query is cheap.
+//! - `Automerge::depth_fragments()` — spec-correct, depth-stratified BFS.
+//!   Lazy: no eager work; full BFS on every query.
+//!
+//! Three bench groups:
+//!
+//! - `fragments/build` — total cost of building a doc, one tx per change.
+//!   Measures the clock-based eager pipeline (depth has no eager cost).
+//!   Reference only.
+//! - `fragments/apply` — `apply_changes` of a pre-extracted Vec.
+//!   Same caveat: the clock-based eager path is always on when calling
+//!   `add_change`, so this also measures clock-based overhead.
+//! - `fragments/query` — head-to-head: time `fragments()` (clock-based)
+//!   vs `depth_fragments()` (spec) on the same prebuilt doc. This is the
+//!   apples-to-apples comparison of pure query cost.
+//!
+//! Topologies:
+//!
+//! - `linear` — one actor, one transaction per commit.
+//! - `concurrent2` — two actors making half the commits each on forks,
+//!   then merging at the end.
+//! - `wide8` — eight actors making concurrent commits on independent
+//!   forks, then merging.
+
+use automerge::{transaction::Transactable, Automerge, ChangeHash, ROOT};
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
+use std::hint::black_box;
+
+/// Build function for a parameterised topology.
+type TopologyBuilder = fn(u64) -> Automerge;
+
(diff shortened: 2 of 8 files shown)


expede:
Turns out that we were counting "what's a fragment" differently, and @orionz 's original PR actually handles an edge case more correctly! Closing in favour of his :D

Issue #1107, opened 2025-06-25
Support WebAssembly Memory64


georgewsu:
With 32-bit wasm, automerge rust is limited to 4GB memory.

This limits the total number/size of documents that automerge-repo can retain in its handleCache, which in turn limits how many/large documents can be concurrently used in a sync server.

Node 24 released support for WebAssembly Memory64 (V8 engine version 13.6)
https://nodejs.org/en/blog/release/v24.0.0

Node v24.3.0 uses v8 version 13.6.233.10 which supports 16GB
https://chromium.googlesource.com/v8/v8/+/refs/tags/13.6.233.10/src/wasm/wasm-limits.h

alexjg:
I think we'll need to wait for `wasm64-unknown-unknown` to be supported by `wasm-bindgen`. At that point it should be pretty straightforward to support `Memory64` technically.

PR #1393, opened 2026-05-21
Add opt-in WebAssembly Memory64 build target (#1107)

simleb:
> **Disclosure:** this PR was fully generated by an AI coding agent (Claude Code) on behalf of someone with effectively zero Rust/WASM background. It is intended as a conversation starter — to gauge appetite, surface objections, and build a little momentum on #1107 — rather than as a drop-in fix to merge as-is. The 4 GB ceiling is becoming pressing for our sync-server workload, so we are hoping to nudge things along.

## Problem

The default `automerge-wasm` build targets `wasm32-unknown-unknown`, which caps a single document/repo at 4 GB of linear memory. For sync servers and other `automerge-repo` consumers that hold many or large documents in `handleCache`, that limit is increasingly load-bearing.

Node 24 (V8 13.6+) and recent Chrome/Firefox now support the WebAssembly Memory64 proposal, raising the ceiling to ~16 GB.

Refs: #1107

## Solution

Add an **opt-in** Memory64 (`wasm64-unknown-unknown`) build alongside the existing 32-bit artifacts. The default build pipeline and existing consumers are untouched.

- `rust/automerge-wasm` gains `npm run release:memory64`, producing `nodejs-memory64/` and `web-memory64/` outputs next to today's `nodejs/ bundler/ deno/ web/`.
- A new `@automerge/automerge-wasm/memory64` sub-path export ships those artifacts.
- `@automerge/automerge` gains a matching `fullfat_node_memory64` entrypoint exposed at `@automerge/automerge/memory64`. The JS build script runs the wasm64 cargo build only when `BUILD_MEMORY64=1` is set, so the default build stays on stable Rust.
- A new `build_wasm_memory64` CI job exercises the path with `continue-on-error: true` (Tier 3 target, nightly toolchain — we don't want nightly regressions blocking PRs).

Consumers opt in with one import swap on top of `automerge-repo/slim` (no `automerge-repo` PR needed — it only depends on `@automerge/automerge/slim`):

```ts
import "@automerge/automerge/memory64"          // initializes wasm64
import { Repo } from "@automerge/automerge-repo/slim"
```

## Known caveats / open questions

These are the spots most likely to need a human eye:

1. **`wasm64-unknown-unknown` is Tier 3** — nightly Rust + `rust-src` only, no precompiled `std`, built via `-Z build-std`. Memory64 itself is still a Stage 4 proposal. The opt-in shape contains this: default users see nothing change.
2. **`getrandom` patch.** The crates.io release of `getrandom` (0.4.2) does not compile under wasm64 because its `wasm_js` backend is gated on `target_arch="wasm32"`. Upstream rust-random/getrandom#848 (merged 2026-05-18, unreleased) extends it to `target_family="wasm"`. To make the build work today the workspace `[patch.crates-io]`-pins `getrandom` to master. Before merging this almost certainly wants to be either (a) waited out until a release is cut, or (b) pinned to a specific commit.
3. **`hexane` (transitive via `automerge`'s `wasm` feature)** built fine for wasm64 on first try, but this hasn't been exercised broadly — worth keeping an eye on.
4. **No browser-side fullfat-memory64 entrypoint** was added — sync servers were the motivating use case and the `web-memory64/` artifact is published in `automerge-wasm` for anyone who wants to wire it up. Happy to add it if you want it in this PR.
5. **wasm-opt** is currently echoed-but-not-run for memory64 (same as the default pipeline). Recent `binaryen` does support memory64; could be enabled with `--enable-memory64` once turned on for the 32-bit path too.

## Test plan

What I ran locally (Node 24.15.0, macOS, nightly rustc 1.97.0):

- [x] `npm run release:memory64` in `rust/automerge-wasm` succeeds and produces `nodejs-memory64/` + `web-memory64/`.
- [x] `wasm-tools print` confirms the output is a true memory64 module: `(memory (;0;) i64 19)` (i64 index type, not i32).
- [x] The memory64 module loads & runs in Node 24 natively (no `--experimental-wasm-memory64` flag required on recent V8).
- [x] `require(\"@automerge/automerge-wasm/memory64\")` resolves and round-trips a small document.
- [x] `BUILD_MEMORY64=1 npm run build` in `javascript/` produces a working `@automerge/automerge/memory64` sub-path export. Both `require(...)` (CJS) and `import * as` (ESM) load and exercise the API end-to-end.
- [x] Default `npm run release` in `rust/automerge-wasm` and `npm run build` in `javascript/` still produce all 32-bit artifacts unchanged.
- [x] `rust/automerge-wasm` mocha suite: 155 passing.
- [x] `javascript` mocha suite: 312 passing.

What was **not** tested:

- [ ] Actually pushing past 4 GB in a real workload (didn't have a 16 GB-RAM box handy; items above prove the build target is wasm64 but not that the ceiling is gone in practice).
- [ ] Workerd / browser memory64 loading paths.
- [ ] CI Linux runner: the new `build_wasm_memory64` job is configured but has not yet executed.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -81,6 +81,38 @@ jobs:
       - name: run tests
         run: ./scripts/ci/wasm_tests
 
+  build_wasm_memory64:
+    runs-on: ubuntu-latest
+    # wasm64-unknown-unknown is Tier 3 (nightly-only, no precompiled std). Don't
+    # gate other jobs on it — nightly regressions shouldn't block PRs.
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-node@v6
+        with:
+          node-version: "24"
+      - uses: jetli/wasm-bindgen-action@v0.2.0
+        with:
+          version: "0.2.121"
+      - name: Install nightly toolchain with rust-src
+        run: rustup toolchain install nightly --component rust-src --profile minimal
+      - name: Build memory64 wasm
+        working-directory: rust/automerge-wasm
+        run: |
+          npm ci
+          npm run release:memory64
+      - name: Smoke test memory64 wasm in Node 24
+        working-directory: rust/automerge-wasm
+        run: |
+          node -e "
+            const wasm = require('./nodejs-memory64/automerge_wasm.cjs');
+            const d = wasm.create();
+            d.put('_root', 'k', 'v');
+            const got = d.materialize();
+            if (got.k !== 'v') { console.error('unexpected:', got); process.exit(1); }
+            console.log('memory64 smoke test passed:', got);
+          "
+
   js_tests:
     runs-on: ubuntu-latest
     needs:
--- a/javascript/package.json
+++ b/javascript/package.json
@@ -16,6 +16,7 @@
     "dist/cjs/**/*.js",
     "dist/cjs/**/*.cjs",
     "dist/cjs/**/*.wasm",
+    "dist/cjs/memory64/**/*",
     "dist/iife/**/*.js",
     "dist/mjs/**/*.js",
     "dist/mjs/**/*.cjs",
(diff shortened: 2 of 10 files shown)


rgbkrk:
Only thing I'll warn about here is that Memory64 is not supported on Safari yet, which also means all webviews on iOS and macOS can't support it (including Tauri apps).

Also you might want to tell your agent to format PR bodies without extra newlines. They're writing as if they're stuck in a limited commit window. Breathe little agent, breathe.

Issue #1287, opened 2026-02-16
Automerge-CLI : Developer Experience (DX) Features


codeCraft-Ritik:
### The CLI currently lacks powerful inspection and manipulation tools. The IDEAS.md file outlines several "standard" commands that are still needed.

**Document Forking :**

- Improvement: Implement a `fork `command that allows a user to create a new Automerge file starting from a specific change hash.
- Command Idea : `automerge fork <hash> <input_file> --out <output_file>`.

**Semantic Diffing :**

- Improvement: Create a `diff `command that shows the structural differences between two Automerge files in a human-readable format, showing what keys were added, removed, or changed.
- **File Path** : `automerge/rust/automerge-cli/src/main.rs.`

**Direct Key Manipulation :**

- Improvement: Implement `get `and `set `commands to read or modify specific paths within a document from the terminal (e.g., `automerge set foo.mpl "users[0].name" "Alice")`.

catancs:
I'd like to take a first pass at the `fork` command from this list (and from IDEAS.md): `automerge fork <hash> <input_file> --out <output_file>`. It maps cleanly onto the existing Automerge::fork_at API, so it's self-contained and doesn't touch the document internals. I'm scoping my PR to `fork` only for now — `diff`, `get`/`set` are larger and I'd rather not bundle them. Happy to follow up separately if this lands well. PR incoming shortly.

alexjg:
We want to keep the issue tracker for bugs that need fixing. This is more of a design discussion. Please open a design discussion about the idea in [discussions](https://github.com/automerge/automerge/discussions) section.

PR #1405, opened 2026-06-11
automerge-cli: add fork subcommand

catancs:
## What

Adds a `fork` subcommand to `automerge-cli`:

```
automerge fork <hash> [input_file] [--out <output_file>]
```

It loads a document, forks it at the given change hash via `Automerge::fork_at`, and writes the resulting document. Like the other subcommands it reads from stdin / writes to stdout when no file paths are given, and it reuses the existing `--skip-verifying-heads` flag. An unknown or malformed hash fails with a clear message instead of producing output.

Partially addresses #1287 (the `fork` item there and in `automerge-cli/IDEAS.md`). Scoped to `fork` only; `diff`/`get`/`set` are left for follow-ups.

## Also included

This PR removes `src/change.rs`, which referenced the long-removed `amf::` frontend API (and an undefined `frontend` binding) and could no longer compile. It was the only user of the `combine` and `maplit` dependencies, so those are dropped from `Cargo.toml` too. Split into its own commit.

## Tests

Adds integration tests covering forking at an earlier change (later state is dropped), forking at the head (state preserved), and an unknown hash (no document emitted). `cargo test`, `cargo fmt --check`, and `cargo clippy` all pass for the crate.


rust/automerge-cli/src/main.rs
```diff
@@ -145,6 +146,23 @@ enum Command {
         /// The file(s) to compact. If empty assumes stdin
         input: Vec<PathBuf>,
     },
+
+    /// Create a new document forked from a given change hash of an existing document
+    Fork {
+        /// The change hash to fork from
+        hash: String,
```
FintanH, line 153: I wonder if this could be made optional and if it's not set we use `fork` instead of `fork_at`

PR #1406, opened 2026-06-13
Prevent `AMresult` struct leakage in the ported WASM sync test suite

jkankiewicz:
I finally noticed that the `sync()` utility function within the `run_ported_wasm_sync_tests()` suite leaks two `AMresult` structs for every iteration of its internal loop. :flushed:

This PR finally fixes that bug.


jkankiewicz:
@alexjg, please review this when you get a chance; it shouldn't take long.

orionz:
Thanks man!

PR #1424, opened 2026-07-04
rust: Harden Cursor string parsing and add unit tests and docs

hayduker:
This PR adds unit tests and fills out the API docs for `Cursor`.

In writing the tests, I found an edge case where `Cursor::from_str` panics when the provided string is empty, has an empty actor ID (i.e. nothing after the `@`), or contains multi-byte UTF-8 characters and is indexed in the middle of one. I replaced the explicit array indexing in the original parsing code with `core::str::strip_prefix` and `core::str::split_once` so the function returns `None` in these situations instead of panicking, and hopefully is a little more readable too.

Based on [the note in the README](https://github.com/automerge/automerge#rust), I'm keeping an eye out for things to add docs and test coverage for as I go through the rust code and learn how it works. If this is useful, let me know and I can submit more changes like this.


FintanH (commented):
This PR looks good to me! I really appreciate the documentation improvements, and the fix is clear!

If you get a chance, could you resolve the CHANGELOG conflict? Otherwise, I think we could do it and if @alexjg is happy we can merge it.

hayduker:
Thanks @FintanH, just resolved the CHANGELOG conflict!

PR #1461, opened 2026-07-16
Hexane 1.0.0-alpha4

orionz:
Hexane changes

added streaming loaders - makes validating on load easy and very flexable 
delta column gets scope_to_value , seek_to_value, scan_to_value support 
moved shiftable trait into hexane 
added codex support and a compile time option for bijou


rust/hexane/benches/codec_compare.rs
alexjg, line 1: Do we still need these benches? It feels like we've established that bijou is faster and keeping the benches around doesn't help us that much?

rust/hexane/src/delta/indexed.rs
```diff
@@ -32,11 +33,61 @@ pub struct IndexedDeltaWeightFn;
...
+
+    // `compute` re-decodes the slab; let the streaming loader accumulate
+    // weights as the runs go past instead.
+    const ACCUMULATES: bool = true;
+
+    /// Incremental twin of [`compute_slab_agg`] — same endpoint math, with
+    /// `len == 0` standing in for the "no runs yet" min/max sentinels (an
+    /// empty weight stays `SlabAgg::default()`, matching `compute`).
+    ///
+    /// Checked arithmetic: this runs on untrusted bytes during load, and
```
alexjg: This doc comment feels more like an implementation note which should either be a normal non-documentation comment or we should drop it IMO.

rust/hexane/src/delta/mod.rs
```diff
@@ -803,12 +913,158 @@ impl<T: DeltaValue> DeltaColumn<T> {
...
+        self.splice_items_inner(
+            index,
+            del,
+            values.into_iter().map(|t| SpliceItem::Value(t.to_i64())),
+        )
+    }
+
+    /// Splice [`DeltaRun`]s in — the run-aware fast path for bulk data.
+    /// A run whose `prefix` matches the running value copies through as
+    /// one inner run; a mismatch costs one peeled fix-up delta.
```
alexjg: This doc comment and the one following on `copy_ranges` is very LLMish. The reference to  fix-up deltas is kind of opaque and in the following comment there are a bunch of strange jargons like "seam deltas re-anchor through the fix -up machinery". I think it's probably worth reviewing all the doc comments on public functions in this typ, I'm not sure they are very useful.

rust/hexane/src/delta/mod.rs
```diff
@@ -803,12 +913,158 @@ impl<T: DeltaValue> DeltaColumn<T> {
...
+                }
+            }
+        }
+    }
+
+    /// Single-point [`copy_ranges`](Self::copy_ranges) with a deletion
+    /// count and the strategy pinned — kept for the strategy-agreement
+    /// tests, which exercise `del`.
+    #[cfg(test)]
+    pub(crate) fn copy_ranges_with(
```
alexjg, line 1004: This is a fair amount of code to keep  for tests, is it worth it?

rust/hexane/src/column.rs
```diff
@@ -571,13 +628,125 @@ impl<'a, T: ColumnValueRef> Iter<'a, T> {
...
+// ── copy_ranges support ─────────────────────────────────────────────────────
+
+/// Execution strategy for `copy_ranges`.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum CopyStrategy {
+    Direct,
+    Invert,
+}
+
+/// Safety factor favouring Direct in [`Column::copy_strategy`].  The
```
alexjg: I'm struggling to understand what the `CopyStrategy` does exactly and what the  job of computing which strategy to use is. This comment is pretty confusing too. Any chance you could elaborate a bit?

rust/hexane/src/column.rs
```diff
@@ -1399,6 +1803,206 @@ where
...
+            );
+            self.weights.push(std::mem::take(&mut self.cur_weight));
+            self.cur_emitted = 0;
+        }
+        Ok(())
+    }
+
+    /// Drain and validate whatever was not pulled, apply the length check
+    /// from the load options, and return the finished column.
+    pub fn finalize(mut self) -> Result<Column<T, C, WF, Idx>, PackError> {
```
alexjg: This is the same logic as `finalize_with` but without a check function right? Can we just call `finalize_with` with a noop function here?

alexjg (commented):
A bunch of small comments here but nothing blocking IMO. Up to you whether to address or merge straight immediately.

rust/hexane/benches/codec_compare.rs
orionz: Ill get rid of them - but one thing to note - when I did a test with a LEB -> BiJou swap on automerge a actually got worse results and claude figured its because BiJou uses 3 bytes instead of two for ints in a range (something like 2048->64k) which is a key range used a lot in mid sized documents in automerge - so its not always a slam dunk - but yes will delete

rust/hexane/src/column.rs
orionz: Good catch

PR #1469, opened 2026-07-23
Use the updated `bijoux` lib with `i64` support

expede:
Martin asked if we could add `i64` support to bijou64, and I used it as an opportunity to consolidate the crates. `bijoux` is the bijou varint crate going forward. I also landed on zigzag encoding over alternatives (looks like that's what you were already doing). Worth noting that twos compliment was often quite a bit faster but seriously increased the space required by negative numbers.

### DIFF
--- a/rust/hexane/src/codec.rs
+++ b/rust/hexane/src/codec.rs
@@ -333,8 +333,8 @@ fn leb_bytes(bits: u64) -> u64 {
 // ── Bijou64 ─────────────────────────────────────────────────────────────────
 
 /// The [bijou64](https://github.com/inkandswitch/bijou) tag-byte codec
-/// (feature `bijou64`) — a thin [`Codec`] adapter over the `bijou64`
-/// crate, plus a zigzag mapping for signed values.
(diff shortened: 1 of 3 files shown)


rust/hexane/src/codec.rs
```diff
@@ -364,54 +364,49 @@ mod bijou64_impl {
...
                 Err(e) => Err(PackError::InvalidValue(format!("bijou64: {e}"))),
             }
         }
 
         fn try_read_signed(data: &[u8]) -> Result<(usize, i64), PackError> {
-            let (n, z) = Self::try_read_unsigned(data)?;
-            Ok((n, unzigzag(z)))
+            match bijoux::i64::decode(data) {
+                Ok((v, n)) => Ok((n, v)),
+                Err(e) => Err(PackError::InvalidValue(format!("bijou64s: {e}"))),
```
expede, line 408: Not sure if you'd actually prefer a semantic error here, but didn't want to add random variants to your other types without your input

PR #1489, opened 2026-08-03
hexane edit / copy_range

orionz:
a collection of new features for hexane - it now has a `copy_ranges` function to efficiently copy ranges of data from one column to anothers and an edit mode that allows for an interface `col.edit().seek(100).delete(10).insert(value)` to efficiently do non contigious edits of a column without having to re-decode the same slab multiple times


rust/hexane/tests/pending.rs
alexjg, line 1: Are we intending to fix these tests before a 1.0?

PR #1500, opened 2026-08-11
javascript/proxies: Use Map instead of object for cache 

### LINKED ISSUE #1415: Writing to Object.prototype-inherited property names inside change() produces inconsistent reads
Assigning to a property whose name shadows an Object.prototype member (e.g. toString, valueOf, hasOwnProperty) inside a change() block behaves inconsistently with reads.
Given a doc handle h, normal property writes work as expected:

```javascript
h.change(doc => {
  doc.foo = { x: 1 };
  console.log(doc.foo); // { x: 1 }
});
console.log(h.doc().foo); // { x: 1 }
```

But writing to an inherited property name does not read back correctly inside the proxy:

```javascript
h.change(doc => {
  doc.toString = { x: 1 };
  console.log(doc.toString); // ƒ toString() { [native code] }
});
console.log(h.doc().toString); // { x: 1 }
```

The read inside the change callback returns the inherited Object.prototype.toString function rather than the value just assigned, while a read from the materialized doc afterward returns the assigned value. The two disagree, which seems wrong — a read of doc.toString immediately after assigning it should reflect the assignment.
Likely the proxy's get trap is falling through to the prototype chain for these keys instead of consulting the underlying document state.

FintanH:
Reads in the map proxy's `get` trap were cached in a plain object and guarded by
`!cache[key]`.

For keys that shadow `Object.prototype` members such as `toString`, `valueOf`,
and `hasOwnProperty`, the lookup fell through the cache's prototype chain and
returned the inherited function instead of the document's value, so reads inside
`change()` disagreed with the materialized document (https://github.com/automerge/automerge/issues/1415).

The same truthiness check also treated cached falsy values as misses,
re-fetching them on every read.

A `Map` has no prototype chain to fall through, and `Map.has` checks key presence
rather than value truthiness, fixing both. It also forced the `get` trap's return
type to honestly include `undefined`.

Fixes https://github.com/automerge/automerge/issues/1415.

### DIFF
--- a/javascript/src/proxies.ts
+++ b/javascript/src/proxies.ts
@@ -67,7 +67,7 @@ type Target = {
   context: Automerge
   objectId: ObjID
   path: Array<Prop>
-  cache: object
+  cache: Map<Prop | symbol, AutomergeValue | undefined>
   trace?: any
 }
 
@@ -245,7 +245,7 @@ const MapHandler = {
   get<T extends Target>(
     target: T,
     key: any,
-  ): AutomergeValue | ObjID | boolean | { handle: Automerge } {
+  ): AutomergeValue | ObjID | boolean | { handle: Automerge } | undefined {
     const { context, objectId, cache } = target
     if (key === Symbol.toStringTag) {
       return target[Symbol.toStringTag]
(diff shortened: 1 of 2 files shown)


FintanH:
@alexjg I reworded the commit. I'll let you decide whether they should be squashed or rebased :blush:

PR #1513, opened 2026-08-19
ci: update to use nix

alexjg:
Problem: CI is currently heavily tied to Github Actions. This means it's very hard to reproduce CI failures locally.

Solution: extend the use of the existing nix tooling and make Github Actions mostly just call nix. The main downside of this approach is that nix is not available for windows, so for windows we use the existing approach.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,6 +3,8 @@
 
   inputs = {
     nixpkgs.url = "nixpkgs/nixos-26.05";
+    # Retain Node 18 for compatibility tests and Node 20 for release builds.
+    nixpkgs-node18.url = "nixpkgs/nixos-23.11";
     nixos-unstable.url = "nixpkgs/nixos-unstable-small";
 
     command-utils.url = "github:expede/nix-command-utils";
@@ -17,6 +19,7 @@
   outputs = {
     self,
     nixpkgs,
+    nixpkgs-node18,
     nixos-unstable,
     command-utils,
     flake-utils,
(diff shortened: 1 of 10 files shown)


flake.nix
```diff
@@ -262,6 +281,153 @@
         command_menu =
           command-utils.commands.${system}
           (release // build // bench // lint // watch // test // docs);
+
+        # Keep the CI entry points in the flake so local development and GitHub
+        # Actions invoke the same commands with the same pinned tools.
+        mk-ci-command = name: runtimeInputs: text:
```
FintanH, line 287: nit: `text` was confusing me as a name, after reading the usage, perhaps `command` might be better?

flake.nix
FintanH, line 1: I was expecting these checks to be a part of `nix flake check`, and so under the `checks` key. Is there are reason you didn't go for that approach?

FintanH (approved):
Mostly makes sense to me! Just a nit on one naming aspect, and one question 😊

flake.nix
alexjg: As I understand things a nix flake check runs in a sandbox with an immutable source tree (it can only write to $out) and no network access. The `ci-*` outputs here don't work like that:

* We don't version the `cargo.lock` and `package-lock.json` files. Instead the `cargo` and `npm` resolve dependencies at build time (i.e. they access the network)
* The JS packaging tests run `npm install` in test projects in a temporary directory (i.e. not in $out)
* `javascript/scripts/build.mjs` inspects git metadata which is not available in a nix check
* The `cargo deny` step intentionally fetches current advisory data

Putting these commands in `checks` would therefore fail. Now, there is an argument here that we are not getting the benefits of nix here because we will still have a drift from local environments to CI (e.g. if the lockfiles are different). We do have some advantages because the versions of various tools (e.g `wasm-bindgen-cli`) are always the same.

To do the full caboodle we would need to 

* Commit the `Cargo.lock` and `package-lock.json` files
* Use `crane` or similar to manage rust dependencies
* Use `buildNpmPackage` for all the JS dependencies
* Pass the git revision into the flake somehow so that we don't need to examine git metadata in `build.mjs`
* Somehow figure out how to run the packaging tests in a nix check

I think most of this is probably worth doing but can probably be done as a next step. Thoughts?

FintanH: Ya, I had a feeling that it was due to the sandboxing. That all makes sense to me :ok_hand: Happy to get this in as the first step!

PR #1531, opened 2026-09-01
Authors v2

alexjg:


rust/automerge/src/op_set2/types.rs
```diff
@@ -747,6 +747,7 @@ impl hexane::RleValue for Action {
 #[derive(PartialEq, Debug, Clone)]
 pub struct ChangeMetadata<'a> {
     pub actor: Cow<'a, ActorId>,
+    pub author: Option<Cow<'a, Author>>,
```
FintanH: @alexjg Is this always `Some` when a change is supposed to have an author, or is it the case that this should be `None` when `seq > 1`?

If it's the latter, I feel uneasy that this can be constructed incorrectly. 

Also the fact that I have to ask the question means that our types aren't speaking the language of our domain, and we have to constantly trace back what it was when we come back to this line of code again.

If it's the former, then I would suggest we use our own `enum`:
```rust
pub enum Authorship<'a> {
  Unauthored,
  Authored(Cow<'a, Author>),
}

impl<'a> Authorship<'a> {
  pub fn to_option(self) -> Option<Cow<'a, Author>> { /* conversion */ }
}
```

alexjg: It's the former but Fintan and I decided to just go with an `Option` for now.

PR #1540, opened 2026-09-06
Validate bundle counter encoding and lengths before expansion

domenkozar:
Malformed bundle counter columns passed to `AutoCommit::load` can panic or allocate hundreds of megabytes before the bundle is rejected. Fuzzing found a 174-byte input with a truncated signed LEB128 counter and two 189-byte inputs whose 65-byte counter column expands to 114,440,696 values (about 873 MiB). The latter exceeded a 10-second fuzz timeout.

Validate both the legacy counter column and the current inverse-counter column with `hexane::DeltaColumn::<i64>::load_with`. This checks encoding, cumulative delta arithmetic, and non-nullability before iteration. Load the operation-ID actor column without expanding its RLE runs, then require the counter column to have exactly the same decoded length before collecting its values. Missing or mismatched counters now return a parse error. This follows the length-validation pattern used by the document operation-column loader and adds no arbitrary document-size limit.

The validated column index adds temporary storage before counter collection. The length check prevents unmatched counter runs from being materialized; it is not a global resource limit for bundles with consistently large columns.

Regression coverage includes all three synthetic fuzz inputs, truncated/overflowing/null counters, too few or too many counters, missing counter columns, amplified runs through both encodings, and valid compressed and empty round trips. Both new fuzz inputs return errors through `AutoCommit::load` and `Bundle::try_from`; the 10 focused counter tests finish in under 0.01 seconds locally.

Validation (from `rust/`):

- `cargo test -p automerge`: 492 passed, 8 ignored, no failures, including integration tests and doctests.
- `cargo fmt -p automerge -- --check`
- `git diff --check`


domenkozar:
cc @FintanH @alexjg

alexjg:
Thanks for this, there's some good stuff here. A few comments:

Can you write the commit messages in your own words. Using LLMs is great but they write very bad commit messages. A good format we use frequently in this project is:

```
<title>

Problem: <present tense description of the problem in the codebase>

Solution: <imperative description of the change made and why it's a good one>

<Any other comments>
```

Likewise could you remove most of the LLM comments, they are needlessly verbose. Lot's of the justification in them belongs in the commit message.

Finally, the binary files used in the fixtures are hard to review. Can you either generate that input programmatically, or provide a script we can use to regenerate the fixtures.

domenkozar:
> Thanks for this, there's some good stuff here. A few comments:
> 
> Can you write the commit messages in your own words. Using LLMs is great but they write very bad commit messages. A good format we use frequently in this project is:
> 
> ```
> <title>
> 
> Problem: <present tense description of the problem in the codebase>
> 
> Solution: <imperative description of the change made and why it's a good one>
> 
> <Any other comments>
> ```
> 
> Likewise could you remove most of the LLM comments, they are needlessly verbose. Lot's of the justification in them belongs in the commit message.
> 
> Finally, the binary files used in the fixtures are hard to review. Can you either generate that input programmatically, or provide a script we can use to regenerate the fixtures.

Done

rust/automerge/src/op_set2/change/batch.rs
```diff
@@ -854,24 +854,55 @@ impl BatchApply {
...
 
-    fn import_ops(&mut self, doc: &mut Automerge) {
+    fn import_ops(&mut self, doc: &mut Automerge) -> Result<(), AutomergeError> {
+        for c in &self.changes {
+            doc.import_ops_to(c, &mut self.ops)?;
+        }
+        Ok(())
+    }
+
+    /// Record the changes in the change graph. Only called once their ops
```
alexjg: This particular requirement (that we call record_history only once ops are inserted) is going to cause problems with some ongoing work we are doing on revoking changes. When handling revocations we need to make sure we have added changes to the change graph before inserting the ops (so that we know if an op is revoked or not, the details aren't that important). Is there a way we can fix this without requiring that op insertion happen before updating history?

rust/automerge/src/storage/bundle/fixtures/truncated-counter-bundle.bin
FintanH, line 1: As @alexjg mentioned [yesterday](https://github.com/automerge/automerge/pull/1540#issuecomment-5599620049):
> Finally, the binary files used in the fixtures are hard to review. Can you either generate that input programmatically, or provide a script we can use to regenerate the fixtures.

Could we get one of the above for this binary file?

rust/automerge/src/storage/bundle/fixtures/slow-counter-bundle.bin
FintanH, line 1: Same goes for these binary files

rust/automerge/tests/fixtures/change_null_value_with_payload.automerge
FintanH, line 1: Same for this one

rust/automerge/src/op_set2/change/batch.rs
```diff
@@ -854,29 +854,97 @@ impl BatchApply {
...
+                    return Err(AutomergeError::MissingPred);
+                }
+            }
+        }
+        for op in &self.ops {
+            if let Some(target) = missing.remove(&op.id()) {
+                if target != (op.bld.obj, op.elemid_or_key()) {
+                    return Err(AutomergeError::MissingPred);
+                }
+            }
```
FintanH: I'm trying to untangle what this second loop is doing. We have iterated over the `ops` above and built `missing`. We then have to iterate over `ops` again to check the target here. Is there a way we can do it in one loop?

rust/automerge/src/op_set2/change/batch.rs
```diff
@@ -854,29 +854,97 @@ impl BatchApply {
...
+            for op in doc.ops().iter_range(&range) {
+                if let Some(target) = missing.remove(&op.id) {
+                    if target != (op.obj, op.elemid_or_key()) {
+                        return Err(AutomergeError::MissingPred);
+                    }
+                }
+            }
+        }
+        if !missing.is_empty() {
+            return Err(AutomergeError::MissingPred);
```
FintanH: This error is going to seem pretty opaque, I believe. If we are going to be reporting on missing predecessors, maybe we should provide more information for debugging?

rust/automerge/src/op_set2/change/batch.rs
```diff
@@ -854,29 +854,97 @@ impl BatchApply {
...
+    fn record_history(&self, doc: &mut Automerge) {
         for c in &self.changes {
-            doc.import_ops_to(c, &mut self.ops).unwrap();
             doc.update_history(c);
         }
         doc.remove_unused_actors(true);
     }
 
+    // Deletes are stored only as successor entries on their predecessors.
+    fn validate_preds(&self, doc: &Automerge) -> Result<(), AutomergeError> {
```
FintanH: I'm slightly concerned that this is going to be adding extra walks over the ops, where we might walk them already and keep track of this somewhere else.

domenkozar:
PTAL

PR #1543, opened 2026-09-07
Add Revocation to Automerge

FintanH:


rust/automerge/src/autocommit.rs
```diff
@@ -430,6 +430,20 @@ impl AutoCommit {
         self.doc.get_authors()
     }
 
+    pub fn revoke(&mut self, author: Author<'static>, heads: &[ChangeHash]) -> Vec<Patch> {
+        self.ensure_transaction_closed();
+        let mut patch_log = PatchLog::active();
+        self.doc.revoke(author, heads, &mut patch_log);
```
alexjg, line 436: I think here and in `unrevoke` we should actually be using `self.patch_log` rather than creating a new one. I'm working on a fix now.

PR #1545, opened 2026-09-09
javascript: Guard public type surface with API Extractor

FintanH:
Add [@microsoft/api-extractor] to catch 'forgotten exports' - types that are reachable through the public API but not re-exported from the entry point (e.g. Author). It analyses dist/index.d.ts and writes an API report to etc/automerge.api.md, which is committed.

scripts/ci/js_tests now runs 'npm run api-check', which fails when the public surface drifts from the committed report - so adding a method or re-exporting a type forces the report to be regenerated (npm run api-update) and reviewed. The workflow is documented in javascript/HACKING.md.

The 15 pre-existing forgotten exports (Author, Hash, the patch types, plus some genuine internals) are recorded as a baseline in the report rather than failing the build. The intent is to burn that list down and then promote ae-forgotten-export to a hard error.

[@microsoft/api-extractor]: https://www.npmjs.com/package/@microsoft/api-extractor


FintanH:
@alexjg Looking through the documentation, a [checked-in report] is the way the workflow is meant to be managed.

[checked-in report]: https://api-extractor.com/pages/overview/demo_api_report/

PR #1566, opened 2026-09-21
Refactor Actor API

FintanH:
While investigating https://github.com/automerge/automerge/issues/697, I was led down the path of exploring enforcing the invariants we maintain with actors through refactoring into new structs and a trait.

The changes should be reviewed by commit, as the refactoring builds upon itself step-by-step. The summary of each is:
1. (a) Introduce `ActorTable` which maintains the set of actors, in lexicographic order, providing an API for insertion, removal, and lookup.
(b) Introduce `ActorShift` and `ActorRemoval` as return types from `insert` and `remove`, so that areas where corresponding actor values are updated using these tokens.
2.  Introduce `ActorIndexed` to have a type to represent the `Vec`s that maintain values indexed by the `ActorTable`. These can only be constructed by providing an `ActorTable`, and their size can only be updated using `ActorShift` and `ActorRemoval`.
3. Introduce `HasActorIndices` for updating values and collections that reference actor indices inside of them, again using `ActorShift` and `ActorRemoval` for updating.


rust/automerge/src/actor/table.rs
```diff
@@ -0,0 +1,125 @@
...
+        Self::default()
+    }
+
+    /// Adopt a document's actor list in the order it was stored.
+    ///
+    /// A document's op columns index actors by position in the stored list,
+    /// so the indices are consistent with the columns. [`Self::lookup`] and
+    /// `OpId` ordering additionally rely on that order being sorted, which
+    /// documents saved by this implementation guarantee.
+    pub(crate) fn from_document_order(table: Vec<ActorId>) -> Self {
```
alexjg, line 53: Should we `assert!` that the table is unique and ordered here?

rust/automerge/src/actor.rs
```diff
@@ -0,0 +1,17 @@
...
+//! index*, and actor indices - not [`ActorId`](crate::ActorId)s - are what
+//! the rest of the crate stores: in `OpId`s, in clocks, in the change graph,
+//! in the op set's columns. Two things rely on the table being sorted:
+//!
+//! * `OpId` ordering tie-breaks on the actor index, which only agrees with
+//!   ordering by actor id when the table is sorted, and
+//! * looking up an actor binary-searches the table.
+//!
+//! [`ActorTable`] owns insertion and lookup so that those two facts are
+//! maintained in one place rather than at every call site.
```
alexjg, line 14: This doc comment feels like it really belongs on the `ActorTable` type?

rust/automerge/src/change_graph.rs
```diff
@@ -1136,8 +1136,14 @@ mod tests {
         fn actor(&mut self) -> ActorId {
             use rand::RngExt;
             let actor = ActorId::from(self.rng.random::<[u8; 16]>().to_vec());
-            self.graph.insert_actor(self.actors.len());
-            self.actors.push(actor.clone());
+            // Mirror `Automerge::put_actor`: keep the table sorted and shift
+            // the graph's actor indices to match.
+            let idx = self
```
alexjg: Should we replace TestGraphBuilder::actors with an `ActorTable` to avoid this duplication?

rust/automerge/src/op_set2/op_set.rs
```diff
@@ -1387,17 +1387,21 @@ impl OpSet {
...
+        if let ActorInsert::Inserted(shift) = &insert {
+            if shift.index() + 1 != self.actors.len() {
+                self.shift(shift)
             }
         }
         insert
     }
 
-    pub(crate) fn rewrite_with_new_actor(&mut self, idx: usize) {
+    fn shift(&mut self, shift: &ActorShift) {
```
alexjg, line 1403: I think we should call this `shift_actors`, otherwise we might confuse it with the "shift" concept in hexane iterators.

rust/automerge/src/clock.rs
```diff
@@ -19,20 +20,33 @@ impl SeqClock {
...
 
-    pub(crate) fn remove_actor(&mut self, idx: usize) {
-        self.0.remove(idx);
+    pub(crate) fn remove_actor(&mut self, removal: &ActorRemoval) {
+        self.0.remove(removal.index());
     }
 
-    pub(crate) fn rewrite_with_new_actor(&mut self, idx: usize) {
-        self.0.insert(idx, None)
+    pub(crate) fn shift(&mut self, shift: &ActorShift) {
```
alexjg, line 27: Rename this to `shift_actor`?

rust/automerge/src/patches/patch_log.rs
```diff
@@ -114,8 +114,10 @@ pub(crate) enum Event {
 }
 
 impl Event {
-    pub(crate) fn with_new_actor(self, idx: usize) -> Self {
-        match self {
+    /// Re-index the op id this event refers to, if it has one. Events which
+    /// carry no op id are returned unchanged.
```
alexjg, line 118: Possibly worth mentioning here that `map_id` will return `None` if `f` returns `None`. It's pretty obvious that this must be what is happening. But it is a _bit_ unusual for something called `map_`.

rust/automerge/src/patches/patch_log.rs
```diff
@@ -555,33 +513,22 @@ impl PatchLog {
...
-        self.events = dirty
-            .into_iter()
-            .map(|(obj, event)| (obj.with_new_actor(index), event.with_new_actor(index)))
-            .collect();
-        let dirty = std::mem::take(&mut self.expose);
-        self.expose = dirty
-            .into_iter()
-            .map(|id| id.with_new_actor(index))
-            .collect();
+    fn shift(&mut self, shift: &ActorShift) {
```
alexjg, line 516: I think we should call this `shift_actors`

alexjg (commented):
Left a few small comments but overall I really like this, it makes the actor table invariants much clearer.

rust/automerge/src/actor/table.rs
FintanH: Maybe `debug_assert`? Otherwise we would have to run the assertion in release execution every time, right?

